PR #1, opened 2019-10-28
MAT-5 - create data model & related

NoelLH:
Remaining to consider for models as of this morning
- [x] Look into locking on read, see e.g. https://stackoverflow.com/a/17721736/2803757 + https://dev.mysql.com/doc/refman/8.0/en/innodb-locking-reads.html - We probably want a repository helper for withdrawing funds safely that does this (will add helper in a later PR, adding notes to come back to this for now)
- [x] Consider safe currency maths. [Money for PHP](http://moneyphp.org/en/stable/) sounded promising but its use of strings seems a bit odd when Doctrine + MySQL have the Decimal type, and if I've read write it can't instantiate non-integer amounts which also seems problematic. Closer-to-native handling with [bcmath](https://www.php.net/manual/en/book.bc.php) might be less restrictive? [PHP Decimal](https://php-decimal.io/) could also be an option – it claims some benefits over bcmath but I'm not sure about its use of a less-PHP-common (to my knowledge) extension
- [x] All todos added on this PR, including handling timestamps in all models and using UUIDs at least for Donations. Lifecycle hooks for timestamps and even createdAt & updatedAt fields themselves might be better abstracted out to a base model or Trait?


NoelLH:
Reading more about UUIDs for primary keys and the complexities of making them as performant as numeric ones, I think for our case auto-inc primary IDs are the way to go. I'll look into the best way for Donations to get generated UUIDs in addition. These can identify donations to the outside world - meaning IDs aren't trivially guessable - while internally joins etc. can still use the numeric auto-incremented ID

PR #2, opened 2019-10-29
MAT-13 - transaction-safe matching allocation & related

NoelLH:


src/Domain/DonationRepository.php
```diff
@@ -25,12 +27,55 @@ public function doPull(SalesforceProxy $proxy): SalesforceProxy
...
+        $this->getEntityManager()->beginTransaction();
+        /** @var CampaignFunding[] $fundings */
+        $fundings = $this->getEntityManager()
+            ->getRepository(CampaignFunding::class)
+            ->getAvailableFundings($donation->getCampaign());
+
+        try {
+            // Loop as long as there are still campaign funds not allocated and we have allocated less than the donation
+            // amount
+            while ($currentFundingIndex < count($fundings) && bccomp($amountLeftToMatch, '0.00', 2) === 1) {
```
tbg-jasonfung: Nice I like the use of BCMath in the main matching element

PR #4, opened 2019-11-10
MAT-6, MAT-7, MAT-8, MAT-9, MAT-11, MAT-12, MAT-14 - Implement all core APIs & commands

NoelLH:


src/Domain/DonationRepository.php
```diff
@@ -4,27 +4,70 @@
...
 namespace MatchBot\Domain;
 
+use DateTime;
 use Doctrine\ORM\ORMException;
-use MatchBot\Client;
+use MatchBot\Application\HttpModels\DonationCreate;
+use MatchBot\Client\BadRequestException;
+use Ramsey\Uuid\Doctrine\UuidGenerator;
 
 class DonationRepository extends SalesforceWriteProxyRepository
```
NoelLH: @tbg-jasonfung I think the most critical things to review are in this file and the `CampaignFundingRepository` methods it calls to acquire locks.

Please shout if any questions on any of it.

tbg-jasonfung: Thanks @NoelLH I've primarily reviewed the suggested file and the 3 methods make sense to me and have also caught up on the use of Doctrines pessimistic lock - thanks

PR #5, opened 2019-11-11
MAT-16 - ECS deployment config + related

NoelLH:


NoelLH:
Need to get this merged to move infra work forward but please have a look @tbg-jasonfung and shout if any questions about how we're doing this.

PR #7, opened 2019-11-13
MAT-20 - add performance-enhancing indexes

NoelLH:


tbg-jasonfung:
@NoelLH The build seems to be stuck, PR approved, merge when this is unblocked

PR #19, opened 2019-11-21
MAT-30 - fund update fixes + inline docs

NoelLH:
* MAT-30 - fix unexpected CampaignFunding available amount resets
* MAT-30 - fix crashes if we get unexpected `null` fund amounts
* Explain what's going on with defined and arbitrary allocation orders

### DIFF
--- a/src/Domain/FundRepository.php
+++ b/src/Domain/FundRepository.php
@@ -40,7 +40,7 @@ class FundRepository extends SalesforceReadProxyRepository
             }
 
             // Then whether new or existing, set its key info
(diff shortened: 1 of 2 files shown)


src/Domain/FundRepository.php
```diff
@@ -58,10 +58,16 @@ public function pullForCampaign(Campaign $campaign): void
                 } else {
                     $campaignFunding->setAllocationOrder(200);
                 }
-            }
 
-            $campaignFunding->setAmountAvailable((string) $fundData['amountForCampaign']);
```
tbg-jasonfung: Yeah wow this would have been bad

PR #49, opened 2019-12-10
Document the fund sync check command's expected behaviour better

NoelLH:


tbg-jasonfung:
Good to merge after resolving small conflict @NoelLH

PR #55, opened 2019-12-18
MAT-59 - upgrade to PHP 7.4 & use typed properties where appropriate

NoelLH:


NoelLH:
> @NoelLH Is typed enforced in PHP 7.4? What happens if I don't include the property type?

@tbg-jasonfung the majority of places we set new values were already type safe because setter methods' arguments are typed.

Typed properties are probably most likely to cause changes where properties are set another way, e.g. by Doctrine loading a model and setting all its properties from the DB. They are an opt-in, backwards compatible feature. So we don't _have_ to do this but it seems like a good idea in the case of something like MatchBot, where the scope is reasonably small and we probably want to always know if data is not what we thought. e.g. in this app it's probably better to crash on an unexpectedly _undefined_ value in a model so we discover edge cases with missing data, rather than carrying on.

[This post](https://stitcher.io/blog/typed-properties-in-php-74) has a nice overview and highlights when it is and isn't OK for typed properties to be uninitialised. And [the RFC](https://wiki.php.net/rfc/typed_properties_v2) has a drier but more complete overview of the feature. One other thing to keep in mind is that `null` is not the same as `undefined`, which is why newly-nullable typed properties (`?int` etc.) must now be initialised with `null` explicitly.

The other cool change to types that's worth a look is [this](https://stitcher.io/blog/new-in-php-74#improved-type-variance-rfc) on co-/contra-variance. Basically it reduces a type 'strictness' in the language that never really made sense and opens the way for inheriting classes to give more specific, accurate return type hints on their overridden methods where these would previously have been incompatible with the parent class.

PR #62, opened 2020-01-29
MAT-59 - update `DonationPublicAuthMiddleware` route introspection for Slim v4.4+

NoelLH:
As detailed in https://github.com/slimphp/Slim/issues/2892 there is effectively a breaking
change in v4.4 which requires this middleware's approach and tests' route handling to be updated.


NoelLH:
I've also just combined in the contents of a branch `console-v5` which I started a few weeks ago, bringing all Symfony libs to v5.x branches.

PR #72, opened 2020-02-26
MAT-70 check whether the fund has been created yet before creating a new

tbg-jasonfung:
MAT-70 check whether the fund has been created yet before creating a new campaignFunding record

### DIFF
--- a/src/Domain/FundRepository.php
+++ b/src/Domain/FundRepository.php
@@ -52,15 +52,18 @@ class FundRepository extends SalesforceReadProxyRepository
             $campaignFunding = $this->campaignFundingRepository->getFunding($campaign, $fund);
             // Otherwise create one
             if (!$campaignFunding) {
-                $campaignFunding = new CampaignFunding();
-                $campaignFunding->setFund($fund);
-                $campaignFunding->addCampaign($campaign);
-                if ($fund instanceof Pledge) {
-                    $campaignFunding->setAllocationOrder(100);
-                } else {
-                    $campaignFunding->setAllocationOrder(200);
+                $availableFunds = getAvailableFundings($campaign);
+                // Only create and link the campaignFunding for the given campaign if it has not been created
+                if (!in_array($fund, $availableFunds)) {
+                    $campaignFunding = new CampaignFunding();
+                    $campaignFunding->setFund($fund);
+                    $campaignFunding->addCampaign($campaign);
+                    if ($fund instanceof Pledge) {
+                        $campaignFunding->setAllocationOrder(100);
+                    } else {
+                        $campaignFunding->setAllocationOrder(200);
+                    }
                 }
-
                 // It's crucial we don't try to 'update' the `amountAvailable` after MatchBot has created the entity,
                 // i.e. only call this when the entity is new. We also assume the amount for each fund is immutable
                 // and so only call `setAmount()` here too.


tbg-jasonfung:
@NoelLH I think this makes sense, otherwise please point me in the correct direction!

src/Domain/FundRepository.php
```diff
@@ -52,15 +52,18 @@ public function pullForCampaign(Campaign $campaign): void
...
             // Otherwise create one
             if (!$campaignFunding) {
-                $campaignFunding = new CampaignFunding();
-                $campaignFunding->setFund($fund);
-                $campaignFunding->addCampaign($campaign);
-                if ($fund instanceof Pledge) {
-                    $campaignFunding->setAllocationOrder(100);
-                } else {
-                    $campaignFunding->setAllocationOrder(200);
+                $availableFunds = getAvailableFundings($campaign);
```
NoelLH: @tbg-jasonfung you're calling this as if it was a global method and not on an object, so there is no way this assignment can work. If you use PhpStorm or similar this error is highlighted for you.

I presume the rest of the logic is therefore not tested manually, as well as not being covered by unit tests? We really need to be thorough and careful with MatchBot changes! As we discussed, a main benefit of picking this one up is getting comfortable with PHP unit tests, so we need to add some! This is also crucial for mentally clarifying the two distinct modes of how this will work and then verifying that the code does the right thing in both cases.

tbg-jasonfung: @NoelLH How do i test this locally since `pullForCampaign` is called when a donation is made?

tbg-jasonfung: > @NoelLH How do i test this locally since `pullForCampaign` is called when a donation is made? I've linked up my `donate-frontend` local repo with Sandbox8, made a donation through that and it directs me to `http://localhost:30030/v1/donations` but its saying `RESOURCE_NOT_FOUND`

Ignore the above comment, my `MatchBot` env salesforce endpoints were using `*-thebiggive`instead of `*biggive`

tbg-jasonfung: @NoelLH Is there a way for me to delete all local MatchBot data? I've run this command `docker-compose exec app composer doctrine:delete-and-recreate` however, when I make a donation, `CampaignFunding.amountAvailable` is created with the value before I purged the data and not the amount set in the Fund API

NoelLH: @tbg-jasonfung if you already pulled in the data and it's remembered when the DB is wiped successfully (does that command give any error output?) then presumably the value is being loaded from Redis, since that's where fund balances live as the first source of truth.

The quickest way to remove old values there is probably to `docker-compose rm` the redis container and then bring a new one up.

tbg-jasonfung: Thanks @NoelLH, think I got this figured out in the end, had to run this command:

`docker exec -it matchbot_redis_1 redis-cli FLUSHALL`

NoelLH: Good call @tbg-jasonfung, actually flushing the data is a bit less brute force and quicker than removing and re-creating the container completely!

It should also be possible to run the same with `docker-compose exec`, to use the container name as it is in `docker-compose.yml`. The advantage of this, apart from readability, is it will work for everyone whereas the number `_1` can vary in some cases. Might be worth adding that version of the command to the readme with an explanation, near where the delete command is introduced under _First run_ (_Data clearing on subsequent runs_ ?) to make this easier for anyone coming to the project?

NoelLH: I've now documented this separately here https://github.com/thebiggive/matchbot/commit/b6e527ceb1e8756cc824b53e19061caa6dd07fd0

NoelLH:
As this change can't work and is missing tests I'm going to close this and start a new branch, since as discussed today we now want to get this ready more urgently.

PR #73, opened 2020-03-17
MAT-70 - support multi-campaign Funds and Fund balance increases

NoelLH:


tbg-jasonfung:
@NoelLH See my last test regarding topping up champion fund after all funds have been exhausted, results are not what I expect

PR #88, opened 2020-07-21
MAT-93 - calculate & send Stripe transfer data

NoelLH:
* give Donation a method to calculate the amount for the charity
* send this amount and Connected Stripe Account ID during Payment Intent setup

### DIFF
--- a/src/Domain/Donation.php
+++ b/src/Domain/Donation.php
@@ -85,6 +85,8 @@ class Donation extends SalesforceWriteProxy
     protected ?string $transactionId = null;
 
     /**
+     * Core donation amount excluding any tip.
+     *
      * @ORM\Column(type="decimal", precision=18, scale=2)
      * @var string Always use bcmath methods as in repository helpers to avoid doing float maths with decimals!
      */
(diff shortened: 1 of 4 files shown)


src/Domain/Donation.php
```diff
@@ -587,11 +589,53 @@ public function hasPostCreateUpdates(): bool
...
+
+    /**
+     * Takes a bcmath string amount with 3 or more decimal places and rounds to
+     * 2 places, with 0.005 rounding up and below rounding down.
+     *
+     * @param string $amount    Simplified from https://stackoverflow.com/a/51390451/2803757 for
+     *                          fixed scale and only positive inputs.
+     * @return string
+     */
+    private function roundAmount(string $amount): string
```
tbg-jasonfung, line 635: @NoelLH Just for my own clarity, why are we passing an `amount` as a string and then rounding like this? Aren't there natural rounding php fns we can use?

NoelLH: @tbg-jasonfung in your example, the type is a *float*. Floats should not be used for currencies in almost any language. [Some reasons why](https://spin.atomicobject.com/2014/08/14/currency-rounding-errors/).

PHP does not have a Decimal type, so the options are using integer pence values everywhere – sometimes good but sometimes confusing – or using bcmath. Until we had fee calculations here there were no multiplications on our side so we could always use 2-place precision consistently and not worry about this. Now we have to round to get the fees, using pence internally is a slightly more attractive option but I still don't think it's worth changing. This should be as complicated as it gets to keep using bcmath and we maintain all its benefits and the testing of amount handling logic we've done so far.

PR #92, opened 2020-07-24
MAT-88 + MAT-92 start tagging stripe payment intent with useful metadata

tbg-jasonfung:
- MAT-88 start tagging stripe payment intent with useful metadata
- MAT-92 optionally set donation fields if available

### DIFF
--- a/src/Application/Actions/Donations/Create.php
+++ b/src/Application/Actions/Donations/Create.php
@@ -68,6 +68,27 @@ class Create extends Action
 
         try {
             $donation = $this->donationRepository->buildFromApiRequest($donationData);
+
+            // Optionally set these donation fields if available
+            if (isset($donationData->billingPostalAddress)) {
+                $donation->setDonorPostalAddress((string) $donationData->billingPostalAddress);
+            }
+
+            if (isset($donationData->emailAddress)) {
+                $donation->setDonorEmailAddress((string) $donationData->emailAddress);
+            }
+    
+            if (isset($donationData->firstName)) {
+                $donation->setDonorFirstName((string) $donationData->firstName);
+            }
+
+            if (isset($donationData->lastName)) {
+                $donation->setDonorLastName((string) $donationData->lastName);
+            }
+
+            if (isset($donationData->tipAmount)) {
+                $donation->setTipAmount((string) $donationData->tipAmount);
+            }
         } catch (\UnexpectedValueException $exception) {
             $message = 'Donation Create data initial model load';
             $this->logger->warning($message . ': ' . $exception->getMessage());
--- a/src/Application/Actions/Donations/Create.php
+++ b/src/Application/Actions/Donations/Create.php
@@ -120,6 +120,8 @@ class Create extends Action
                         'environment' => getenv('APP_ENV'),
                         'isGiftAid' => $donation->isGiftAid(),
                         'matchedAmount' => $donation->getFundingWithdrawalTotal(),
+                        'optInCharityEmail' => $donation->getCharityComms(),
+                        'optInTbgEmail' => $donation->getTbgComms(),
                         'tbgTipGiftAid' => $donation->isGiftAid(), // TODO use real value after MVP
                     ],
                     // See https://stripe.com/docs/connect/destination-charges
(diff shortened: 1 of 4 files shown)


src/Application/Actions/Donations/Create.php
```diff
@@ -91,6 +91,13 @@ protected function action(): Response
                     // See https://stripe.com/docs/api/payment_intents/object
                     'amount' => (100 * $donation->getAmount()),
                     'currency' => 'gbp',
+                    'metadata' => [
+                        'campaignId' => $donation->getCampaign()->getSalesforceId(),
```
NoelLH, line 95: @tbg-jasonfung I think campaign name as mentioned on the description would also be handy.

Thinking further it's probably good to include all 4 (charity + campaign) (names + IDs) and comms flags so we can see these all at a glance.

src/Application/Actions/Donations/Create.php
```diff
@@ -91,6 +91,13 @@ protected function action(): Response
...
+                        'charityName' => $donation->getCampaign()->getCharity()->getName(),
+                        'env' => getenv('APP_ENV'),
```
NoelLH, line 97: I would call this `environment` here for clarity unless we're short on space.

src/Application/Actions/Donations/Create.php
```diff
@@ -91,6 +91,13 @@ protected function action(): Response
...
+                        'isGiftAid' => $donation->isGiftAid(),
```
NoelLH, line 98: It's probably good to have `coreDonationGiftAid` and `tbgTipGiftAid` and set them both to the main flag's value initially, as it sounds like we plan to split the declaration soon.

src/Application/Actions/Donations/Create.php
tbg-jasonfung: @NoelLH I don't think we're sending campaign name through the API, should we start doing this?

NoelLH: @tbg-jasonfung we are storing the name, have a look at the model. You may need to add a getter if this is the first use. IDEs can generally generate this for you.

src/Application/Actions/Donations/Create.php
```diff
@@ -68,6 +68,27 @@ protected function action(): Response
 
         try {
             $donation = $this->donationRepository->buildFromApiRequest($donationData);
+
+            // Optionally set these donation fields if available
+            if (isset($donationData->billingPostalAddress)) {
```
NoelLH, line 73: We'll need to revisit whether we actually need to collect billing and postal address separately for Gift Aid + card processing purposes. This doesn't align with the WIP for the frontend but may still be the best option.

src/Application/Actions/Donations/Create.php
```diff
@@ -68,6 +68,27 @@ protected function action(): Response
 
         try {
             $donation = $this->donationRepository->buildFromApiRequest($donationData);
+
+            // Optionally set these donation fields if available
```
NoelLH, line 72: These setters are all doing what `buildFromApiRequest()` describes as its action, so should be in that method.

We're also missing country code.

src/Application/Actions/Donations/Create.php
```diff
@@ -68,6 +68,27 @@ protected function action(): Response
...
+            if (isset($donationData->billingPostalAddress)) {
+                $donation->setDonorPostalAddress((string) $donationData->billingPostalAddress);
```
NoelLH, line 74: The values which are strings should not need casting to string. What is the use case here? We might also want to consider whether blank string is helpful and test strings for `empty()` instead if it's better to leave the fields null.

src/Domain/DonationRepository.php
```diff
@@ -107,6 +107,31 @@ public function buildFromApiRequest(DonationCreate $donationData): Donation
         $donation->setCharityComms($donationData->optInCharityEmail);
         $donation->setTbgComms($donationData->optInTbgEmail);
 
+        // Optionally set these donation fields if available
+        if (!empty($donationData->billingPostalAddress)) {
```
tbg-jasonfung, line 111: @NoelLH I've just done an optional check for billingPostalAddress for now, we can revisit once we start sending the separate GA claim address

PR #96, opened 2020-08-04
MAT-102 – separate Create and Update / details setup steps

NoelLH:
Plus various Stripe Intent fixes and refactoring.


src/Domain/Donation.php
```diff
@@ -344,55 +348,44 @@ public function getDonorLastName(): ?string
...
-    public function setDonorLastName(string $donorLastName): void
+    public function setDonorLastName(?string $donorLastName): void
     {
         $this->donorLastName = $donorLastName;
     }
 
-    public function getDonorPostalAddress(): ?string
+    public function getDonorBillingAddress(): ?string
     {
         return $this->donorPostalAddress;
```
tbg-jasonfung, line 358: @NoelLH probably should look to update these so it matches the method

tests/Application/Actions/Donations/CreateTest.php
```diff
@@ -438,14 +441,18 @@ public function testSuccessWithMatchedCampaignUsingStripe(): void
             ->willReturn($paymentIntentMockResult)
             ->shouldBeCalledOnce();
 
+//        $stripePaymentIntentsProphecy->create(Argument::type('array'))
+//            ->willReturn($paymentIntentMockResult)
+//            ->shouldBeCalledOnce();
```
tbg-jasonfung, line 446: @NoelLH Not needed?

src/Domain/Donation.php
NoelLH: @tbg-jasonfung the property is left as-is intentionally because Doctrine ORM annotations use it to determine the database field name. Changing or customising that would mean more cognitive overhead. As the getter/setter are the external interface for the field I don't think it's a problem to clarify their names while leaving the field as-is.

PR #101, opened 2020-08-06
MAT-85 stripe update webhook

tbg-jasonfung:
MAT-85 stripe update webhook using Stripes own event method + update routes file

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -16,8 +16,7 @@ return function (App $app) {
     $app->get('/ping', Status::class);
 
     $app->group('/v1', function (RouteCollectorProxy $versionGroup) {
-        // Currently the only unauthenticated endpoint besides `/ping` app-wide,
-        // and the only unauthenticated endpoint in the `/v1` group.
+        // Current unauthenticated endpoint in the `/v1` group.
         $versionGroup->post('/donations', Donations\Create::class);
 
         $versionGroup->group('/donations/{donationId:[a-z0-9-]{36}}', function (RouteCollectorProxy $group) {
@@ -29,6 +28,9 @@ return function (App $app) {
 
     $app->put('/hooks/donation/{donationId:.{36}}', Hooks\DonationUpdate::class)
         ->add(DonationHookAuthMiddleware::class);
+
+    // Unauthenticated as Stripes SDK handles signature verification
+    $app->post('/hooks/stripe', Hooks\StripeUpdate::class);
 
     $app->options('/{routes:.+}', function ($request, $response, $args) {
         return $response;
--- a/app/routes.php
+++ b/app/routes.php
@@ -29,7 +29,7 @@ return function (App $app) {
     $app->put('/hooks/donation/{donationId:.{36}}', Hooks\DonationUpdate::class)
         ->add(DonationHookAuthMiddleware::class);
 
-    // Unauthenticated as Stripes SDK handles signature verification
+    // Authenticated through Stripes SDK signature verification
     $app->post('/hooks/stripe', Hooks\StripeUpdate::class);
 
     $app->options('/{routes:.+}', function ($request, $response, $args) {
--- a/src/Application/Actions/Hooks/StripeUpdate.php
+++ b/src/Application/Actions/Hooks/StripeUpdate.php
@@ -46,42 +46,32 @@ class StripeUpdate extends Action
         try {
             $event = \Stripe\Webhook::constructEvent($payload, $signature, $webhookSecret);
         } catch (\UnexpectedValueException $e) {
-            $logMessage = 'Invalid Payload';
-            $this->logger->warning($logMessage);
-            $error = new ActionError(ActionError::BAD_REQUEST, null ?? $logMessage);
-            return $this->respond(new ActionPayload(400, null, $error));
+            return $this->validationError("Invalid Payload", null);
         } catch (\Stripe\Exception\SignatureVerificationException $e) {
-            $logMessage = 'Invalid Signature';
-            $this->logger->warning($logMessage);
-            $error = new ActionError(ActionError::BAD_REQUEST, null ?? $logMessage);
-            return $this->respond(new ActionPayload(400, null, $error));
+            return $this->validationError("Invalid Signature", null);
         }
 
-        /** @var Donation $donation */
-        $donation = $this->donationRepository->findOneBy(['transactionId' => $event->data->object->id]);
+        if ($event instanceof StripeEvent) {
+            /** @var Donation $donation */
+            $donation = $this->donationRepository->findOneBy(['transactionId' => $event->data->object->id]);
 
-        if ($event instanceof StripeEvent && $donation) {
-            // Handle the event
-            switch ($event->type) {
-                case 'payment_intent.succeeded':
-                    $this->handlePaymentIntentSucceeded($event, $donation);
-                    break;
-                case 'payment_intent.created':
-                    $logMessage = 'Unsupported Action';
-                    $this->logger->warning($logMessage);
-                    $error = new ActionError(ActionError::BAD_REQUEST, null ?? $logMessage);
-                    return $this->respond(new ActionPayload(400, null, $error));
-                default:
-                    $logMessage = 'Unsupported Action';
-                    $this->logger->warning($logMessage);
-                    $error = new ActionError(ActionError::BAD_REQUEST, null ?? $logMessage);
-                    return $this->respond(new ActionPayload(400, null, $error));
+            if ($donation) {
+                // Handle the event
+                switch ($event->type) {
+                    case 'payment_intent.succeeded':
+                        $this->handlePaymentIntentSucceeded($event, $donation);
+                        break;
+                    default:
+                        return $this->validationError("Unsupported Action", null);
+                }
+            } else {
+                $logMessage = 'No Content';
+                $this->logger->warning($logMessage);
+                $error = new ActionError(ActionError::BAD_REQUEST, null ?? $logMessage);
+                return $this->respond(new ActionPayload(204, null, $error));
             }
         } else {
-            $logMessage = 'No Content';
-            $this->logger->warning($logMessage);
-            $error = new ActionError(ActionError::BAD_REQUEST, null ?? $logMessage);
-            return $this->respond(new ActionPayload(204, null, $error));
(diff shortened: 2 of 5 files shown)


src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -0,0 +1,70 @@
...
+            $logMessage = 'Invalid Signature';
+            $this->logger->warning($logMessage);
+            $error = new ActionError(ActionError::BAD_REQUEST, null ?? $logMessage);
+            return $this->respond(new ActionPayload(400, null, $error));
+        }
+
+        $payload = new ActionPayload(200, $event->data->object);
+
+        /** @var Donation $donation */
+        $donation = $this->donationRepository->findOneBy(['transactionId' => $event->data->object->id]);
```
NoelLH, line 62: @tbg-jasonfung I think it would be good to check for the object type. Non-Payment Intents (or whatever object we expect to subscribe to) should be handled gracefully in the case that those notifications are added in error, maybe with a 204 No Content response, so that this cannot stop Stripe sending the webhooks we _do_ want as a result of the webhook receiver appearing to be down.

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -0,0 +1,70 @@
...
+
+        if (!$donation) {
+            throw new DomainRecordNotFoundException('Donation not found');
```
NoelLH, line 65: @tbg-jasonfung similarly, this should probably return 204 No Content. I've written more about why on the Salesforce webhook counterpart's ticket – we want to ignore unknown items quietly so that multiple test environments (e.g. staging, regression) can have working webhook receivers at once.

app/routes.php
```diff
@@ -30,6 +29,9 @@
     $app->put('/hooks/donation/{donationId:.{36}}', Hooks\DonationUpdate::class)
         ->add(DonationHookAuthMiddleware::class);
 
+    // Unauthenticated as Stripes SDK handles signature verification
```
NoelLH, line 32: @tbg-jasonfung I would just tweak this comment to point out that the Action _is_ authenticated, just another way - middleware isn't the only way to do auth.

tbg-jasonfung:
@NoelLH I think this is ready for a first pass review, I've tried to map the metadata as accurately as I can using this test donation Barbora made yesterday: https://jsonblob.com/caced298-d895-11ea-bc69-436fb80d0dba - it doesn't appear to have a `Lastname` will probably need to do some processing on `Name` to help split these out

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -0,0 +1,133 @@
...
+    protected function action(): Response
+    {
+        $payload = $this->request->getBody();
+        $signature = $this->request->getHeaderLine('stripe-signature');
+        $webhookSecret = 'whsec_V4mcupoQsVuKmP8PuTMIAHf83E8LBRHt'; // TODO replace with one supplied in Infra
+
+        try {
+            $event = \Stripe\Webhook::constructEvent($payload, $signature, $webhookSecret);
+        } catch (\UnexpectedValueException $e) {
+            $logMessage = 'Invalid Payload';
```
NoelLH, line 49: @tbg-jasonfung what about using the recently added helper to slim down these 400s' setup?

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -0,0 +1,133 @@
...
+        $this->serializer = $serializer;
+
+        parent::__construct($logger);
+    }
+
+    protected function action(): Response
+    {
+        $payload = $this->request->getBody();
+        $signature = $this->request->getHeaderLine('stripe-signature');
+        $webhookSecret = 'whsec_V4mcupoQsVuKmP8PuTMIAHf83E8LBRHt'; // TODO replace with one supplied in Infra
```
NoelLH: Needs a var as discussed

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -0,0 +1,133 @@
...
+            return $this->respond(new ActionPayload(400, null, $error));
+        } catch (\Stripe\Exception\SignatureVerificationException $e) {
+            $logMessage = 'Invalid Signature';
+            $this->logger->warning($logMessage);
+            $error = new ActionError(ActionError::BAD_REQUEST, null ?? $logMessage);
+            return $this->respond(new ActionPayload(400, null, $error));
+        }
+
+        /** @var Donation $donation */
+        $donation = $this->donationRepository->findOneBy(['transactionId' => $event->data->object->id]);
```
NoelLH, line 61: @tbg-jasonfung we should check the event's object type before doing this lookup, rather than after

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -0,0 +1,133 @@
...
+
+        if ($event instanceof StripeEvent && $donation) {
+            // Handle the event
+            switch ($event->type) {
+                case 'payment_intent.succeeded':
+                    $this->handlePaymentIntentSucceeded($event, $donation);
+                    break;
+                case 'payment_intent.created':
```
NoelLH, line 69: Don't need 2 identical case branches

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -0,0 +1,133 @@
...
+                default:
+                    $logMessage = 'Unsupported Action';
+                    $this->logger->warning($logMessage);
+                    $error = new ActionError(ActionError::BAD_REQUEST, null ?? $logMessage);
+                    return $this->respond(new ActionPayload(400, null, $error));
+            }
+        } else {
+            $logMessage = 'No Content';
+            $this->logger->warning($logMessage);
+            $error = new ActionError(ActionError::BAD_REQUEST, null ?? $logMessage);
```
NoelLH, line 83: Not a bad request. The body should be an empty string in line with the 'No Content' code if we're going to follow this convention.

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -0,0 +1,133 @@
...
+            empty($event->data->object->id));
+        if ($missingRequiredField) {
+            $message = 'Hook missing required values';
+            $this->logger->warning("Donation ID {$event->data->object->id}: {$message}");
+            $error = new ActionError(ActionError::BAD_REQUEST, $message);
+
+            return $this->respond(new ActionPayload(400, null, $error));
+        }
+
+        $donation->setDonationStatus($event->status);
```
NoelLH, line 107: We probably need to do something to align this with the existing statuses - Collected and Paid in particular - for everything to continue working.

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -0,0 +1,133 @@
...
+        $donation->setDonorBillingAddress($event->data->object->billing_details->address->postal_code);
```
NoelLH, line 108: I'm not sure we need to set all these values that are immutable. They go to Stripe for reference so I think it's needless overhead to re-set them here. It's probably good to check they exist as expected still but we only really need to set status unless I'm missing something.

NoelLH:
> @NoelLH I think this is ready for a first pass review, I've tried to map the metadata as accurately as I can using this test donation Barbora made yesterday: https://jsonblob.com/caced298-d895-11ea-bc69-436fb80d0dba - it doesn't appear to have a `Lastname` will probably need to do some processing on `Name` to help split these out

@tbg-jasonfung  as noted inline, I don't think we need to set back values from Stripe which come from us and don't change. This includes stuff they save in a different format like Name.

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -0,0 +1,103 @@
...
+        $this->serializer = $serializer;
+
+        parent::__construct($logger);
+    }
+
+    protected function action(): Response
+    {
+        $payload = $this->request->getBody();
+        $signature = $this->request->getHeaderLine('stripe-signature');
+        $webhookSecret = getenv('STRIPE_WEBHOOK_SIGNING_SECRET');
```
NoelLH, line 44: @tbg-jasonfung as we already have a `'stripe'` key in the config object I think we should put this there too so it's obvious where to look for configs, but can do this in a later update.

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -0,0 +1,103 @@
...
+                    case 'payment_intent.succeeded':
+                        $this->handlePaymentIntentSucceeded($event, $donation);
+                        break;
+                    default:
+                        return $this->validationError("Unsupported Action", null);
+                }
+            } else {
+                $logMessage = 'No Content';
+                $this->logger->warning($logMessage);
+                $error = new ActionError(ActionError::BAD_REQUEST, null ?? $logMessage);
```
NoelLH, line 69: Still don't need this `$error`!

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -0,0 +1,103 @@
...
+                switch ($event->type) {
+                    case 'payment_intent.succeeded':
+                        $this->handlePaymentIntentSucceeded($event, $donation);
+                        break;
+                    default:
+                        return $this->validationError("Unsupported Action", null);
+                }
+            } else {
+                $logMessage = 'No Content';
+                $this->logger->warning($logMessage);
```
NoelLH, line 68: I think this can just be an `info()` call, as we expect it in sandboxes when receiving data from others, and should probably include the event type / ID for reference.

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -0,0 +1,103 @@
...
+
+    public function handlePaymentIntentSucceeded(StripeEvent $event, $donation): Response
+    {
+        // For now we support the happy success path,
+        // as this is the only event type we're handling right now,
+        // convert status to the one SF uses.
+        if ($event->status == 'succeeded') {
+            $donation->setDonationStatus('Collected');
+        } else {
+            return $this->validationError("Unsupported Status", null);
```
NoelLH, line 87: No need to set 2nd param

PR #104, opened 2020-08-11
MAT-108 charity comms webhook

tbg-jasonfung:
- MAT-108 start setting charity comm preferences when converting to hook model

### DIFF
--- a/src/Domain/Donation.php
+++ b/src/Domain/Donation.php
@@ -215,11 +215,9 @@ class Donation extends SalesforceWriteProxy
         $data['updatedTime'] = $this->getUpdatedDate()->format(DateTime::ATOM);
         $data['amountMatchedByChampionFunds'] = (float) $this->getConfirmedChampionWithdrawalTotal();
         $data['amountMatchedByPledges'] = (float) $this->getConfirmedPledgeWithdrawalTotal();
-        $data['optInCharityEmail'] = $this->getCharityComms();
 
-        // For backwards compatability with Enthuse
-        if ($this->getPsp() != 'stripe') {
-            unset($data['optInCharityEmail']);
+        if ($this->getCharityComms) {
+            $data['optInCharityEmail'] = $this->getCharityComms();
         }
 
         unset(
--- a/src/Domain/Donation.php
+++ b/src/Domain/Donation.php
@@ -215,7 +215,6 @@ class Donation extends SalesforceWriteProxy
         $data['updatedTime'] = $this->getUpdatedDate()->format(DateTime::ATOM);
         $data['amountMatchedByChampionFunds'] = (float) $this->getConfirmedChampionWithdrawalTotal();
         $data['amountMatchedByPledges'] = (float) $this->getConfirmedPledgeWithdrawalTotal();
-        $data['optInCharityEmail'] = $this->getCharityComms();
 
         unset(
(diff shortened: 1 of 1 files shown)


src/Domain/Donation.php
```diff
@@ -215,14 +215,19 @@ public function toHookModel(): array
         $data['updatedTime'] = $this->getUpdatedDate()->format(DateTime::ATOM);
         $data['amountMatchedByChampionFunds'] = (float) $this->getConfirmedChampionWithdrawalTotal();
         $data['amountMatchedByPledges'] = (float) $this->getConfirmedPledgeWithdrawalTotal();
+        $data['optInCharityEmail'] = $this->getCharityComms();
+
+        // For backwards compatability with Enthuse
+        if ($this->getPsp() != 'stripe') {
```
tbg-jasonfung, line 221: @NoelLH would you prefer strict `!==` here?

NoelLH: @tbg-jasonfung yes, comparisons should always be strict unless there is a reason not.

Moreover, why don't we set the value only when Stripe is involved rather than setting then un-setting? And is there a reason to base it specifically on PSP rather than whether a value is present?

tbg-jasonfung: @NoelLH Okay thanks - yeah think it makes sense to set the value when present rather than checking for psp, i've made a change to reflect this

src/Domain/Donation.php
```diff
@@ -215,11 +215,9 @@ public function toHookModel(): array
         $data['updatedTime'] = $this->getUpdatedDate()->format(DateTime::ATOM);
         $data['amountMatchedByChampionFunds'] = (float) $this->getConfirmedChampionWithdrawalTotal();
         $data['amountMatchedByPledges'] = (float) $this->getConfirmedPledgeWithdrawalTotal();
-        $data['optInCharityEmail'] = $this->getCharityComms();
 
-        // For backwards compatability with Enthuse
-        if ($this->getPsp() != 'stripe') {
-            unset($data['optInCharityEmail']);
+        if ($this->getCharityComms) {
```
NoelLH, line 219: @tbg-jasonfung this (a) must be a method call – I suggest using e.g. the free Jetbrains IDEs to highlight errors like this – and (b) must check for null as distinct from false, otherwise opt outs will be ignored.

tbg-jasonfung: Actually @NoelLH shouldn't we just set it regardless what the data is, SF side defaults `null` and uses what's currently saved on the record, otherwise for Stripe stuff it will set to T or F.

src/Domain/Donation.php
```diff
@@ -215,14 +215,14 @@ public function toHookModel(): array
         $data['updatedTime'] = $this->getUpdatedDate()->format(DateTime::ATOM);
         $data['amountMatchedByChampionFunds'] = (float) $this->getConfirmedChampionWithdrawalTotal();
         $data['amountMatchedByPledges'] = (float) $this->getConfirmedPledgeWithdrawalTotal();
+        $data['optInCharityEmail'] = $this->getCharityComms();
```
NoelLH, line 218: @tbg-jasonfung this is already set this way in `toApiModel()`, so if this is the solution then we don't need the double-set – deleting the `unset` line has the same effect.

PR #105, opened 2020-08-18
MAT-105 unit test stripe update hook

tbg-jasonfung:


tests/Application/Actions/Hooks/StripeUpdateTest.php
```diff
@@ -0,0 +1,59 @@
...
+        $container->set(StripeWebhook::class, $stripeRepoProphecy->reveal());
+        $container->set(EntityManagerInterface::class, $entityManagerProphecy->reveal());
+        $container->set(DonationRepository::class, $donationRepoProphecy->reveal());
+
+        $stripeSignature = 't=' . time() . 'v1=' . $this->getValidAuth($body) . 'v0=' . $this->getValidAuth($body);
+
+        $request = $this->createRequest('POST', '/hooks/stripe', $body)
+            ->withHeader('stripe-signature', $stripeSignature);
+        
+        $response = $app->handle($request);
```
tbg-jasonfung, line 44: @NoelLH any idea where I'm going wrong here, the test is throwing an error `Using $this when not in object context` - here's the stack trace:

> 1) MatchBot\Tests\Application\Actions\Hooks\StripeUpdateTest::testUnsupportedAction
Error: Using $this when not in object context

/var/www/html/src/Application/Actions/Hooks/StripeUpdate.php:53
/var/www/html/src/Application/Actions/Action.php:44
/var/www/html/vendor/slim/slim/Slim/Handlers/Strategies/RequestResponse.php:43
/var/www/html/vendor/slim/slim/Slim/Routing/Route.php:381
/var/www/html/vendor/slim/slim/Slim/MiddlewareDispatcher.php:81
/var/www/html/vendor/slim/slim/Slim/MiddlewareDispatcher.php:81
/var/www/html/vendor/slim/slim/Slim/Routing/Route.php:341
/var/www/html/vendor/slim/slim/Slim/Routing/RouteRunner.php:84
/var/www/html/app/routes.php:40
/var/www/html/vendor/slim/slim/Slim/MiddlewareDispatcher.php:283
/var/www/html/vendor/slim/slim/Slim/Middleware/RoutingMiddleware.php:60
/var/www/html/vendor/slim/slim/Slim/MiddlewareDispatcher.php:140
/var/www/html/vendor/slim/slim/Slim/MiddlewareDispatcher.php:81
/var/www/html/vendor/slim/slim/Slim/App.php:215
/var/www/html/tests/Application/Actions/Hooks/StripeUpdateTest.php:44

tbg-jasonfung: Ignore this, got it working in the end!

src/Domain/StripeWebhook.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace MatchBot\Domain;
```
NoelLH, line 3: This is not part of our domain – it doesn't have a persisted database object or support one. I think `Application\Actions\Hooks\Stripe` or possible `Application\HttpModels` would be more suitable homes.

tests/TestData/success.json
```diff
@@ -0,0 +1,155 @@
+{
```
NoelLH, line 1: Good call to make full mock files for these. However as we're making a global home for test data, I think this needs more specific naming so it's clear what it's for. The type of hook should be in a folder name or the file name. Same with the others.

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -36,18 +36,19 @@ public function __construct(
...
         // As `settings` is just an array for now, I think we have to inject Container to do this.
-        $this->webhookSecret = $container->get('settings')['stripe']['webhookSecret'];
+        $this->webhookSecret = (string) $container->get('settings')['stripe']['webhookSecret'];
 
         parent::__construct($logger);
     }
 
     protected function action(): Response
     {
+        $webhook = new StripeWebhook();
```
NoelLH, line 46: The idea behind extending Stripe's class was to allow it be injected as a service and therefore mocked. If we are instantiating it directly here then tests are not mocking it.

Using the Di approach, we would instead take `StripeWebhook` as a constructor argument and then use the DI setup (as in `dependencies.php`) to tell the container how to provide that class.

However, given how far you've got without doing this, maybe my premise was wrong and we can cover everything without mocking Stripe's client here, instead testing with fake data but real Stripe client code. More on my other comments.

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -36,18 +36,19 @@ public function __construct(
         $this->entityManager = $entityManager;
         $this->serializer = $serializer;
         // As `settings` is just an array for now, I think we have to inject Container to do this.
-        $this->webhookSecret = $container->get('settings')['stripe']['webhookSecret'];
+        $this->webhookSecret = (string) $container->get('settings')['stripe']['webhookSecret'];
```
NoelLH, line 39: Why do we need this cast? Tests should have a value set and it should be a string already. If an env is accidentally set up with no secret, the endpoint should probably crash!

tests/Application/Actions/Hooks/StripeUpdateTest.php
```diff
@@ -0,0 +1,179 @@
...
+
+        $donationRepoProphecy
+            ->push(Argument::type(Donation::class), false)
+            ->willReturn(true)
+            ->shouldBeCalledOnce();
+
+        $entityManagerProphecy = $this->prophesize(EntityManagerInterface::class);
+        $stripeRepoProphecy = $this->prophesize(StripeWebhook::class);
+
+        $container->set(StripeWebhook::class, $stripeRepoProphecy->reveal());
```
NoelLH, line 151: It's interesting that this works, because:

1. the container is not the source of `StripeWebhook` – you are instantiating it directly inside `action()` instead of getting it through dependency injection
2. the prophecy has no methods set up, so if it was being used it would never return the test data or verify a signature

If everything is being covered this suggests that maybe we didn't need to mock Stripe's SDK at all to cover this endpoint, in which case we don't need these container overrides and prophecies, and we quite possibly don't need the concrete class I suggested either (sorry!)

NoelLH (changes requested):
This is looking like a solid approach and thorough @tbg-jasonfung, but I've added a few suggestions. The key one is that you're not actually using DI or mocks for the SDK so can probably simplify accordingly.

src/Application/Actions/Hooks/StripeUpdate.php
tbg-jasonfung: @NoelLH CircleCI was throwing this yesterday `TypeError: Typed property MatchBot\Application\Actions\Hooks\StripeUpdate::$webhookSecret must be string, bool used` but on my local it was fine without it being casted to `String`

NoelLH: @tbg-jasonfung that suggests you are relying on a var definition which is in your local `.env` but not set up in the test for clean container runs. Have a look at how other vars are overridden in `phpunit.xml[.dist]`

tests/Application/Actions/Hooks/StripeUpdateTest.php
tbg-jasonfung: Thanks for the clarification @NoelLH think i'm getting this hang of this prophecy-wizardry business.

It looks like we still need to concrete class, doing the following, throws `Error: Cannot instantiate abstract class Stripe\\Webhook`

```
use Stripe\Webhook;

...

$webhook = new Webhook();
```

src/Application/Actions/Hooks/StripeUpdate.php
tbg-jasonfung: @NoelLH Cool thanks, this makes sense, i've removed the casting + updated the xml file

tests/Application/Actions/Hooks/StripeUpdateTest.php
NoelLH: @tbg-jasonfung you cannot instantiate the abstract class, but that's not what we were doing before. Does calling the method statically as before work?

PR #107, opened 2020-08-26
MAT-106 listen for charge success event + new chargeId DB field

tbg-jasonfung:
- MAT-106 Add new table field chargeId and related getter and setter methods
- MAT-106 refactor and listen for charge succeeded hook + start storing charge Id
- Update Stripe test data with charge object structure + rename files
- Update unit tests


src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -78,12 +77,20 @@ protected function action(): Response
...
+
+        if (!$donation) {
+            throw new DomainRecordNotFoundException('Donation not found');
+        }
+        
         // For now we support the happy success path,
         // as this is the only event type we're handling right now,
         // convert status to the one SF uses.
         if ($event->data->object->status === 'succeeded') {
+            $donation->setChargeId($event->data->object->id);
```
tbg-jasonfung, line 93: @NoelLH do you think it's worth sending the charge Id to SF and storing it there as well, or you think it'll be cleaner to just have payment intent Id to identify a donation on SF side?

NoelLH: I don't think it really needs to be in SF too @tbg-jasonfung as long as MatchBot can link them up. We can always check the object associations in Stripe admin.

src/Migrations/Version20200826135513.php
```diff
@@ -0,0 +1,35 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Add field for Stripe charge ID
+ */
+final class Version20200826135513 extends AbstractMigration
+{
+    public function getDescription() : string
+    {
+        return '';
```
NoelLH, line 17: @tbg-jasonfung it's good to set this too so console commands show the purpose.

tests/Application/Actions/Hooks/StripeUpdateTest.php
```diff
@@ -24,16 +25,10 @@ public function testUnsupportedAction(): void
         $container = $app->getContainer();
 
         $body = file_get_contents(dirname(__DIR__, 3) . '/TestData/StripeWebhook/pi_canceled.json');
-        $donation = $this->getTestDonation();
         $webhookSecret = $container->get('settings')['stripe']['webhookSecret'];
         $time = (string) time();
 
         $donationRepoProphecy = $this->prophesize(DonationRepository::class);
-        $donationRepoProphecy
```
NoelLH, line 32: We can probably remove this prophecy completely if it has no mocked methods left?

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -78,18 +77,26 @@ protected function action(): Response
...
     }
 
-    public function handlePaymentIntentSucceeded(Event $event, $donation): Response
+    public function handleChargeSucceeded(Event $event): Response
     {
+        /** @var Donation $donation */
+        $donation = $this->donationRepository->findOneBy(['transactionId' => $event->data->object->payment_intent]);
+
+        if (!$donation) {
+            throw new DomainRecordNotFoundException('Donation not found');
```
NoelLH, line 86: This should probably info log and return a 204 as well so as not to break multi-sandbox use

NoelLH (changes requested):
Key stuff looking good @tbg-jasonfung, just a few small suggestions

tests/Application/Actions/Hooks/StripeUpdateTest.php
tbg-jasonfung: @NoelLH When I remove this prophecy I get the following error:

`Error: Call to a member function setCampaignRepository() on null`

NoelLH: Ah, OK @tbg-jasonfung. If you look at the trace of that call, it's to do with the 'real' `DonationRepository` setup's reliance on `EntityManagerInterface` in `repositories.php`. It looks like we do need to mock the donation repo in this case but not the EM, so we can remove that and it's DI override instead.

NoelLH (approved):

PR #108, opened 2020-08-28
MAT-107 start listening to payout.paid events and update donation statuses accordingly

tbg-jasonfung:
- MAT-107 start listening to payout.paid events and update donation statuses accordingly

### DIFF
--- a/src/Application/Actions/Hooks/StripeUpdate.php
+++ b/src/Application/Actions/Hooks/StripeUpdate.php
@@ -37,6 +37,7 @@ class StripeUpdate extends Action
         $this->entityManager = $entityManager;
         $this->serializer = $serializer;
         // As `settings` is just an array for now, I think we have to inject Container to do this.
+        $this->apiKey = $container->get('settings')['stripe']['apiKey'];
         $this->webhookSecret = $container->get('settings')['stripe']['webhookSecret'];
 
         parent::__construct($logger);
@@ -66,6 +67,8 @@ class StripeUpdate extends Action
                 case 'charge.succeeded':
                     $this->handleChargeSucceeded($event);
                     break;
+                case 'payout.paid':
+                    $this->handlePayoutPaid($event);
                 default:
                     $this->logger->info('Unsupported Action');
                     return $this->respond(new ActionPayload(204));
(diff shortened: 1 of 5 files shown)


tbg-jasonfung:
@NoelLH difficult to simulate this on local, Stripes CLI doesn't support `payout.paid` events - but otherwise would be good to trial run this on staging

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -112,4 +116,72 @@ public function handleChargeSucceeded(Event $event): Response
 
         return $this->respondWithData($event->data->object);
     }
+
+    public function handlePayoutPaid(Event $event)
+    {
+        $stripeClient = new \Stripe\StripeClient($this->apiKey);
```
NoelLH, line 122: There is already a Stripe client available in the container, we should be making a new one.

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -112,4 +116,72 @@ public function handleChargeSucceeded(Event $event): Response
...
+            $balanceTransactions = $stripeClient->balanceTransactions->all($attributes);
+
+            foreach ($balanceTransactions->data as $balanceTransaction) {
+                $paidChargeIds[] = $balanceTransaction->source;
+                $lastBalanceTransactionId = $balanceTransaction->id;
+            }
+
+            $hasMore = $balanceTransactions->has_more;
+
+            // We get a Stripe exception if we start this with a null or empty value,
```
NoelLH, line 147: This doesn't make sense – if `$hasMore` is false then no more iterations of the `while` loop occur.

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -112,4 +116,72 @@ public function handleChargeSucceeded(Event $event): Response
...
+            // so we only include this if there's more items to iterate.
+            if ($hasMore) {
+                $attributes['start_after'] = $lastBalanceTransactionId;
+            }
+        }
+        $this->logger->info('Getting all paid charge Ids complete, found: ' . sizeof($paidChargeIds));
+
+        if (sizeof($paidChargeIds) > 0) {
```
NoelLH, line 155: Standard `count()`?

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -112,4 +116,72 @@ public function handleChargeSucceeded(Event $event): Response
...
+            $count = 0;
+
+            foreach ($paidChargeIds as $Id) {
```
NoelLH, line 158: Local var `$chargeId` would be clearer.

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -112,4 +116,72 @@ public function handleChargeSucceeded(Event $event): Response
...
+                        // We're confident to set donation status to paid because this
+                        // method is called only when Stripe event `payout.paid` is received.
+                        $donation->setDonationStatus('Paid');
+
+                        $this->entityManager->persist($donation);
+                        $this->donationRepository->push($donation, false);
+
+                        $count++;
+                    } else {
+                        $this->logger->error('Unexpected donation status found for charge Id: ' . $Id);
```
NoelLH, line 173: Surely donations already _Paid_ should not error? Remember that webhooks can be re-sent and delivery order is not guaranteed.

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -66,6 +67,9 @@ protected function action(): Response
                 case 'charge.succeeded':
                     $this->handleChargeSucceeded($event);
                     break;
+                case 'payout.paid':
+                    $this->handlePayoutPaid($event);
```
NoelLH, line 71: Unit tests?!

NoelLH (changes requested):

NoelLH:
> @NoelLH difficult to simulate this on local, Stripes CLI doesn't support `payout.paid` events - but otherwise would be good to trial run this on staging

@tbg-jasonfung I'm not sure what the behaviour is in sandboxes with simulated Connect bank accounts. I suspect we might not get fully realistic results there either.

In either case, this is all the more reason to have good unit tests covering this logic so that we can be confident the code is self-consistent and we are clear what we _expect_ Stripe to do, so that we verify our logic in response is valid and can more easily investigate if the Stripe behaviour diverges from what our test behaviour did.

src/Application/Actions/Hooks/StripeUpdate.php
tbg-jasonfung: @NoelLH I was going to ask on Friday, how do I mock a Stripe API response?

NoelLH: @tbg-jasonfung it should be very similar to all the other places we use Prophecy. You need to prophesise `StripeClient`. This is another reason to use the injected service instead of creating a new copy in the action.

src/Application/Actions/Hooks/StripeUpdate.php
tbg-jasonfung: @NoelLH the exception is encountered on the first call, where by default `$lastBalanceTransactionId` is set to null

NoelLH: @tbg-jasonfung in that case, we can just check for that ID being non-null? Then it's clear why we're doing this and it's less likely to get removed or refactored incorrectly.

tests/Application/Actions/Hooks/StripeUpdateTest.php
```diff
@@ -137,6 +138,152 @@ public function testSuccess(): void
...
+        $stripeClientProphecy->balanceTransactions = $stripeBalanceTransactionProphecy->reveal();
+
+        $container->set(DonationRepository::class, $donationRepoProphecy->reveal());
+        $container->set(EntityManagerInterface::class, $entityManagerProphecy->reveal());
+        $container->set(StripeClient::class, $stripeClientProphecy->reveal());
+
+        $request = $this->createRequest('POST', '/hooks/stripe', $body)
+            ->withHeader('Stripe-Signature', $this->generateSignature($time, $body, $webhookSecret));
+
+        $app->handle($request);
```
NoelLH, line 229: Check response code?

tests/Application/Actions/Hooks/StripeUpdateTest.php
```diff
@@ -137,6 +138,152 @@ public function testSuccess(): void
...
```
NoelLH, line 282: Check response code?

NoelLH (commented):
This is looking a lot healthier @tbg-jasonfung but there's a couple of very important elements I think we still need to test before this goes live

NoelLH (approved):

PR #111, opened 2020-09-25
MAT-110 update donation update endpoint to handle champion comms

tbg-jasonfung:
- MAT-110 add new champion comms field to Donation table
- MAT-110 update http models with new champion comms field
- MAT-110 start setting champion comms flag from request data
- MAT-110 update unit tests


NoelLH (approved):
Noice

PR #112, opened 2020-10-12
MAT-113 unconditionally return payout paid response

tbg-jasonfung:
- MAT-113 unconditionally return payout paid response

### DIFF
--- a/src/Application/Actions/Hooks/StripeUpdate.php
+++ b/src/Application/Actions/Hooks/StripeUpdate.php
@@ -111,7 +111,7 @@ class StripeUpdate extends Action
     }
 
     public function handlePayoutPaid(Event $event): Response
-    {   
+    {
         $count = 0;
         $payoutId = $event->data->object->id;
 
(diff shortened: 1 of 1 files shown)


src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -175,9 +175,9 @@ public function handlePayoutPaid(Event $event): Response
                     return $this->respond(new ActionPayload(400));
                 }
             }
-
-            $this->logger->info(sprintf('Acknowledging paid donations complete, persisted: %s', $count));
-            return $this->respondWithData($event->data->object);
         }
+        
+        $this->logger->info(sprintf('Acknowledging paid donations complete, persisted: %s', $count));
```
NoelLH, line 180: We need to initialise `$count` unconditionally for it to be safe to use here.

NoelLH (changes requested):

src/Application/Actions/Hooks/StripeUpdate.php
tbg-jasonfung: Good catch, thanks @NoelLH

NoelLH (approved):

PR #113, opened 2020-10-12
MAT-90 stripe refund webhook

tbg-jasonfung:
- MAT-90 add refunded event handling logic + update unit tests
- MAT-90 charge refunded event json test data

### DIFF
--- a/src/Application/Actions/Hooks/StripeUpdate.php
+++ b/src/Application/Actions/Hooks/StripeUpdate.php
@@ -177,7 +177,7 @@ class StripeUpdate extends Action
                 }
             }
         }
-        
+
         $this->logger->info(sprintf('Acknowledging paid donations complete, persisted: %s', $count));
         return $this->respondWithData($event->data->object);
     }
@@ -206,9 +206,12 @@ class StripeUpdate extends Action
             return $this->validationError(sprintf('Unsupported Status "%s"', $event->data->object->status));
         }
 
-        // Release match funds only if the donation was matched and the refunded amount is equal to the local txn amount.
+        // Release match funds only if the donation was matched and
+        // the refunded amount is equal to the local txn amount.
         // We multiply local donation amount by 100 to match Stripes calculations.
-        if ($donation->isReversed() && $donation->getCampaign()->isMatched() && $donation->getAmount() * 100 === $amountRefunded) {
+        if ($donation->isReversed() &&
+            $donation->getCampaign()->isMatched() &&
+            $donation->getAmount() * 100 === $amountRefunded) {
             $this->donationRepository->releaseMatchFunds($donation);
         }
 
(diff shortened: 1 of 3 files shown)


src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -180,4 +182,33 @@ public function handlePayoutPaid(Event $event): Response
...
+
+        if (!$donation) {
+            $this->logger->info(sprintf('Donation not found with Charge ID %s', $chargeId));
+            return $this->respond(new ActionPayload(204));
+        }
+
+        // For now we support the happy success path,
+        // as this is the only event type we're handling right now,
+        // convert status to the one SF uses.
+        if ($event->data->object->status === 'succeeded') {
```
NoelLH, line 201: Wrong status? If not then we need an update comment to explain how this relates to the Stripe statuses for the charge & refund.

NoelLH (changes requested):
One thing in line. Also, what about releasing match funds?

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -205,6 +206,10 @@ public function handleChargeRefunded(Event $event): Response
             return $this->validationError(sprintf('Unsupported Status "%s"', $event->data->object->status));
         }
 
+        if ($donation->isReversed() && $event->data->object->metadata->matchedAmount > 0) {
```
NoelLH, line 209: It feels safer, and more resilient to the fact hooks could come in in any order, to base this decision on our current local status rather than what's in the hook.

`if ($donation->isReversed() && $donation->getCampaign()->isMatched()) { ...` is used elsewhere and should be safe even if the actual matched amount is £0 I think. We could also do a check of the amount from the local txn info.

NoelLH (changes requested):
Looking good – one last thing I think

src/Application/Actions/Hooks/StripeUpdate.php
tbg-jasonfung: @NoelLH good shout, I've updated this with the suggested stricter checks before releasing match funds where the amount refunded must match the local amount. For partial refunds (tip refunds) I guess it'll belong to the same logic branch but we don't release match funds, probably one for another time once we have BDD written for this.

src/Application/Actions/Hooks/StripeUpdate.php
```diff
@@ -186,6 +186,7 @@ public function handlePayoutPaid(Event $event): Response
     public function handleChargeRefunded(Event $event): Response
     {
         $chargeId = $event->data->object->id;
+        $amountRefunded = (float) $event->data->object->amount_refunded;
```
NoelLH, line 188: @tbg-jasonfung wouldn't it make more sense to cast the other value to int, since both are whole numbers (pence)? Alternatively we could use `bccmp()` so we're not concerned about what type the numbers started as.

NoelLH (approved):

PR #131, opened 2020-11-17
Log all the donations gathered from the query

tbg-jasonfung:
- Log all the donations gathered from the query

### DIFF
--- a/src/Domain/SalesforceProxy.php
+++ b/src/Domain/SalesforceProxy.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace MatchBot\Domain;
 
+use DateTime;
 use Doctrine\ORM\Mapping as ORM;
 
 /**
@@ -32,5 +33,21 @@ abstract class SalesforceProxy extends Model
     public function setSalesforceId(string $salesforceId): void
     {
         $this->salesforceId = $salesforceId;
+    }
+
+    /**
+     * @return DateTime
+     */
+    public function getCreatedAt(): ?DateTime
+    {
+        return $this->createdAt;
+    }
+
+    /**
+     * @return string
+     */
+    public function getDonationStatus(): ?string
+    {
+        return $this->donationStatus;
     }
 }
--- a/src/Domain/SalesforceProxy.php
+++ b/src/Domain/SalesforceProxy.php
@@ -34,20 +34,4 @@ abstract class SalesforceProxy extends Model
     {
         $this->salesforceId = $salesforceId;
     }
-
-    /**
-     * @return DateTime
-     */
-    public function getCreatedAt(): ?DateTime
-    {
-        return $this->createdAt;
-    }
-
-    /**
-     * @return string
-     */
-    public function getDonationStatus(): ?string
-    {
-        return $this->donationStatus;
-    }
 }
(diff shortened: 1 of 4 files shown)


src/Domain/SalesforceWriteProxyRepository.php
```diff
@@ -65,6 +65,14 @@ public function pushAllPending(int $limit = 200): int
             ['id' => 'ASC'],
             $limit,
         );
+        
+        if (getenv('APP_ENV') === 'regression') {
+            $this->logger->info(sprintf(
+                'Donations to update: ',
+                $proxiesToUpdate,
```
NoelLH, line 72: I'm not sure this is going to tell you much @tbg-jasonfung, have you tried it on local? I would expect you'll get a summary of the objects that shows they exist but not any of their properties.

src/Domain/SalesforceWriteProxy.php
```diff
@@ -38,6 +38,22 @@ public function getSalesforceLastPush(): ?DateTime
...
+     */
+    public function getCreatedAt(): ?DateTime
+    {
+        return $this->createdAt;
+    }
+
+    /**
+     * @return string
+     */
+    public function getDonationStatus(): ?string
```
NoelLH, line 52: Doesn't belong in the base class

src/Domain/SalesforceWriteProxy.php
```diff
@@ -38,6 +38,22 @@ public function getSalesforceLastPush(): ?DateTime
         return $this->salesforceLastPush;
     }
 
+    /**
+     * @return DateTime
+     */
+    public function getCreatedAt(): ?DateTime
```
NoelLH, line 44: Shouldn't be in base class as that's not what uses the trait which defines the field.

src/Domain/SalesforceWriteProxyRepository.php
```diff
@@ -65,8 +65,17 @@ public function pushAllPending(int $limit = 200): int
             ['id' => 'ASC'],
             $limit,
         );
+
         foreach ($proxiesToUpdate as $proxy) {
             $this->push($proxy, false);
+            if (getenv('APP_ENV') === 'regression') {
+                $createdAt = $proxy->getCreatedAt();
+                $this->logger->info(
+                    'Donation Id to update: ' . $proxy->getSalesforceId() .
```
NoelLH, line 74: This won't tell us if there are newer ones beyond what the query reaches. Wouldn't it make more sense to query for the oldest and newest instead of following the existing query's limits?

NoelLH (changes requested):

src/Domain/SalesforceProxy.php
```diff
@@ -33,4 +34,20 @@ public function setSalesforceId(string $salesforceId): void
     {
         $this->salesforceId = $salesforceId;
     }
+
+    /**
+     * @return DateTime
+     */
+    public function getCreatedAt(): ?DateTime
```
NoelLH, line 41: These are both still wrong for the same reasons! The getters need to be defined at the same level as what they read from so that they are guaranteed to work.

NoelLH (changes requested):

src/Domain/SalesforceProxy.php
tbg-jasonfung: @NoelLH i'm not quite sure what you mean, where should they live?

NoelLH (approved):

PR #135, opened 2020-11-18
MAT-121 - fix and add int cast failsafes to Redis fund balance data

NoelLH:


tbg-jasonfung:
This PR should fix the following encountered error:

`[php7:notice] [pid 18] [client 172.19.0.1:40850] TypeError: bcdiv() expects parameter 1 to be string, int given - #0 /var/www/html/src/Application/Matching/OptimisticRedisAdapter.php(147): bcdiv(0, '100', 2)`

tbg-jasonfung (approved):

PR #139, opened 2020-11-19
MAT-130 - add `matchbot:reset` command

NoelLH:
Also remove unused Doctrine adapter and related logic


composer.json
```diff
@@ -86,6 +86,11 @@
             "Composer\\Config::disableProcessTimeout",
             "php matchbot-cli.php matchbot:push-donations"
         ],
+        "matchbot:reset": [
+            "Composer\\Config::disableProcessTimeout",
+            "php matchbot-cli.php matchbot:reset-matching",
+            "@doctrine:delete-and-recreate"
+        ],
```
tbg-jasonfung, line 93: @NoelLH do we need to also do this? `redis redis-cli FLUSHALL`

tbg-jasonfung (commented):
Looks good, just a small inline query

composer.json
NoelLH: @tbg-jasonfung I don't think so. The whole purpose of `matchbot:reset-matching` is to remove just the relevant fund balance / matching data from Redis in a clean way, without unnecessarily nuking everything like caches that share the instance.

tbg-jasonfung (approved):

PR #140, opened 2020-11-19
MAT-111 support variable fees

tbg-jasonfung:
- MAT-111 add new charity fee field on donation table
- MAT-111 calculate and set the charity depending on psp, card brand and card country
- MAT-111 update unit tests to include latest changes

### DIFF
--- a/src/Domain/Donation.php
+++ b/src/Domain/Donation.php
@@ -110,7 +110,7 @@ class Donation extends SalesforceWriteProxy
      * Fee the charity takes on,
      * For Enthuse: 1.2% of $amount + 0.20p
      * For Stripe (EU / UK): 1.5% of $amount + 0.20p
-     * For Stripe (Non EU / Amex): 3.2% of $amount + 0.20
+     * For Stripe (Non EU / Amex): 3.2% of $amount + 0.20p
      *
      * @ORM\Column(type="decimal", precision=18, scale=2)
      * @var string Always use bcmath methods as in repository helpers to avoid doing float maths with decimals!
--- a/src/Domain/Donation.php
+++ b/src/Domain/Donation.php
@@ -108,7 +108,7 @@ class Donation extends SalesforceWriteProxy
 
     /**
      * Fee the charity takes on,
-     * For Enthuse: 1.2% of $amount + 0.20p
+     * For Enthuse: 1.9% of $amount + 0.20p
      * For Stripe (EU / UK): 1.5% of $amount + 0.20p
      * For Stripe (Non EU / Amex): 3.2% of $amount + 0.20p
      *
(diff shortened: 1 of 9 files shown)


src/Domain/Donation.php
```diff
@@ -269,6 +291,8 @@ public function toApiModel(): array
             'tipAmount' => (float) $this->getTipAmount(),
             'tipGiftAid' => $this->hasTipGiftAid(),
             'transactionId' => $this->getTransactionId(),
+            'cardBrand' => '',
```
NoelLH, line 294: These 2 should be based on the current values?

src/Domain/Donation.php
```diff
@@ -442,6 +474,36 @@ public function setAmount(string $amount): void
...
+     * @param string $cardBrand
+     * @param string $cardCountry - default to 'GB' on create, an update command will override this where necessary
+     * @param string $charityFee
+     */
+    public function setCharityFee(string $psp, string $cardBrand = '', string $cardCountry = 'GB'): void
+    {
+        $feeAmountFixed = '0.20';   // 20p fixed per-donation
+
+        if ($psp === 'enthuse') {
+            $feeRatio = '0.012';
```
NoelLH, line 488: Should be 0.019 – we used 1.2% in the old method because Stripe data was its only use case

src/Domain/Donation.php
```diff
@@ -98,6 +106,17 @@ class Donation extends SalesforceWriteProxy
      */
     protected string $amount;
 
+    /**
+     * Fee the charity takes on,
+     * For Enthuse: 1.2% of $amount + 0.20p
```
NoelLH, line 111: Should be 1.9%

src/Domain/Donation.php
```diff
@@ -442,6 +474,36 @@ public function setAmount(string $amount): void
         $this->amount = $amount;
     }
 
+    /**
+     * @param string $psp
+     * @param string $cardBrand
+     * @param string $cardCountry - default to 'GB' on create, an update command will override this where necessary
+     * @param string $charityFee
+     */
+    public function setCharityFee(string $psp, string $cardBrand = '', string $cardCountry = 'GB'): void
```
NoelLH, line 483: Probably safer for defaults to be null and types nullable

NoelLH (changes requested):
Looks good so far – few things inline

src/Domain/Donation.php
tbg-jasonfung: @NoelLH updating this to '0.019' seems to be breaking some irrelevent tests, i'm not 100% sure why?! Any ideas from this output log:

```
MatchBot\Tests\Application\Actions\Donations\CreateTest::testSuccessWithStripeAccountIDMissingInitiallyButFoundOnRefetch
TypeError: Argument 1 passed to MatchBot\Domain\Donation::setClientSecret() must be of the type string, null given, called in /var/www/html/src/Application/Actions/Donations/Create.php on line 151
```
and 

```
MatchBot\Tests\Application\Actions\Donations\CreateTest::testSuccessWithMatchedCampaignUsingStripe
TypeError: Argument 1 passed to MatchBot\Domain\Donation::setClientSecret() must be of the type string, null given, called in /var/www/html/src/Application/Actions/Donations/Create.php on line 151
```

PR #142, opened 2020-11-20
MAT-133 support enthuse gift aid fee calculation

tbg-jasonfung:
- MAT-133 support enthuse gift aid fee calculation


NoelLH (approved):
It would probably be good to add a unit test covering the Gift Aid case too, but for now I think the most important thing is to get this manually tested asap

PR #143, opened 2020-11-20
MAT-133 Add enthuse ga fee calculation unit test

tbg-jasonfung:
- Add enthuse ga fee calculation postive and negative unit tests


NoelLH (approved):
Nice

PR #145, opened 2020-11-24
Include charity and champion comms data

tbg-jasonfung:
- Include charity and champion comms data


NoelLH:
I don't think we get this data in the hook, so this isn't the right fix. See https://github.com/thebiggive/matchbot/pull/146

PR #149, opened 2020-11-25
MAT-136 get connect account charge status

tbg-jasonfung:
- MAT-136 update stripe paid status handling for connected accounts
- MAT-136 Update unit tests + add more mock data files


NoelLH (approved):
Nice!

PR #160, opened 2020-12-01
Develop -> master

NoelLH:


NoelLH:
As the changes here only impact logging (except one earlier exit point for blank JWTs), or have been effectively QA'd as [detailed here](https://thebiggive.atlassian.net/browse/MAT-139?focusedCommentId=21069), I think this is pretty safe to push through.

PR #181, opened 2021-01-07
MAT-149 upgrade to 8.0

tbg-jasonfung:
- Update docker image to 8.0 and packages
- Fix tests using non numeric values
- Stop using method that will be deprecated in next phpunit release, import and start using 'prophecy-phpunit' package
- Start using constructor property promotion and union types

### DIFF
--- a/src/Application/Actions/Donations/Create.php
+++ b/src/Application/Actions/Donations/Create.php
@@ -30,9 +30,9 @@ class Create extends Action
     public function __construct(
         private DonationRepository $donationRepository,
         private EntityManagerInterface $entityManager,
-        protected LoggerInterface $logger,
         private SerializerInterface $serializer,
-        private StripeClient $stripeClient
+        private StripeClient $stripeClient,
+        LoggerInterface $logger
     ) {
         parent::__construct($logger);
     }
--- a/src/Application/Actions/Donations/Create.php
+++ b/src/Application/Actions/Donations/Create.php
@@ -6,6 +6,7 @@ namespace MatchBot\Application\Actions\Donations;
 
 use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 use Doctrine\ORM\EntityManagerInterface;
+use JetBrains\PhpStorm\Pure;
 use MatchBot\Application\Actions\Action;
 use MatchBot\Application\Actions\ActionError;
 use MatchBot\Application\Actions\ActionPayload;
@@ -25,23 +26,14 @@ use Symfony\Component\Serializer\SerializerInterface;
 
 class Create extends Action
 {
-    private DonationRepository $donationRepository;
-    private EntityManagerInterface $entityManager;
-    private SerializerInterface $serializer;
-    private StripeClient $stripeClient;
-
+    #[Pure]
     public function __construct(
-        DonationRepository $donationRepository,
-        EntityManagerInterface $entityManager,
-        LoggerInterface $logger,
-        SerializerInterface $serializer,
-        StripeClient $stripeClient
+        private DonationRepository $donationRepository,
+        private EntityManagerInterface $entityManager,
+        protected LoggerInterface $logger,
+        private SerializerInterface $serializer,
+        private StripeClient $stripeClient
     ) {
-        $this->donationRepository = $donationRepository;
-        $this->entityManager = $entityManager;
-        $this->serializer = $serializer;
-        $this->stripeClient = $stripeClient;
-
         parent::__construct($logger);
     }
 
--- a/src/Application/Actions/Donations/Get.php
+++ b/src/Application/Actions/Donations/Get.php
@@ -17,7 +17,7 @@ class Get extends Action
     #[Pure]
     public function __construct(
         private DonationRepository $donationRepository,
-        protected LoggerInterface $logger
+        LoggerInterface $logger
     ) {
         parent::__construct($logger);
     }
--- a/src/Application/Actions/Donations/Get.php
+++ b/src/Application/Actions/Donations/Get.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace MatchBot\Application\Actions\Donations;
 
+use JetBrains\PhpStorm\Pure;
 use MatchBot\Application\Actions\Action;
 use MatchBot\Domain\DomainException\DomainRecordNotFoundException;
 use MatchBot\Domain\Donation;
@@ -13,14 +14,11 @@ use Psr\Log\LoggerInterface;
 
 class Get extends Action
 {
-    private DonationRepository $donationRepository;
-
+    #[Pure]
     public function __construct(
-        DonationRepository $donationRepository,
-        LoggerInterface $logger
+        private DonationRepository $donationRepository,
+        protected LoggerInterface $logger
     ) {
-        $this->donationRepository = $donationRepository;
-
         parent::__construct($logger);
     }
 
--- a/src/Application/Actions/Donations/Update.php
+++ b/src/Application/Actions/Donations/Update.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace MatchBot\Application\Actions\Donations;
 
 use Doctrine\ORM\EntityManagerInterface;
+use JetBrains\PhpStorm\NoReturn;
+use JetBrains\PhpStorm\Pure;
 use MatchBot\Application\Actions\Action;
 use MatchBot\Application\Actions\ActionError;
 use MatchBot\Application\Actions\ActionPayload;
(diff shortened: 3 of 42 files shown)


src/Application/Actions/Donations/Create.php
```diff
@@ -25,23 +26,14 @@
...
+    #[Pure]
     public function __construct(
-        DonationRepository $donationRepository,
-        EntityManagerInterface $entityManager,
-        LoggerInterface $logger,
-        SerializerInterface $serializer,
-        StripeClient $stripeClient
+        private DonationRepository $donationRepository,
+        private EntityManagerInterface $entityManager,
+        protected LoggerInterface $logger,
```
NoelLH, line 33: I'm not sure it makes sense to re-declare `$logger`'s visibility – I don't think we should be promoting it because that's handled by the parent class through the parent constructor call.

src/Application/Actions/Donations/Get.php
```diff
@@ -13,14 +14,11 @@
...
 class Get extends Action
 {
-    private DonationRepository $donationRepository;
-
+    #[Pure]
     public function __construct(
-        DonationRepository $donationRepository,
-        LoggerInterface $logger
+        private DonationRepository $donationRepository,
+        protected LoggerInterface $logger
```
NoelLH, line 20: As above

src/Application/Actions/Donations/Update.php
```diff
@@ -25,23 +27,14 @@
...
+    #[Pure]
     public function __construct(
-        DonationRepository $donationRepository,
-        EntityManagerInterface $entityManager,
-        LoggerInterface $logger,
-        SerializerInterface $serializer,
-        StripeClient $stripeClient
+        private DonationRepository $donationRepository,
+        private EntityManagerInterface $entityManager,
+        protected LoggerInterface $logger,
```
NoelLH, line 34: As above

src/Application/Actions/Hooks/DonationUpdate.php
```diff
@@ -19,20 +20,13 @@
...
-
+    #[Pure]
     public function __construct(
-        DonationRepository $donationRepository,
-        EntityManagerInterface $entityManager,
-        LoggerInterface $logger,
-        SerializerInterface $serializer
+        private DonationRepository $donationRepository,
+        private EntityManagerInterface $entityManager,
+        protected LoggerInterface $logger,
```
NoelLH, line 27: As above

src/Application/Actions/Status.php
```diff
@@ -6,23 +6,19 @@
...
-    private EntityManagerInterface $entityManager;
-    private ?Redis $redis;
-
+    #[Pure]
     public function __construct(
-        EntityManagerInterface $entityManager,
-        LoggerInterface $logger,
-        ?Redis $redis
+        private EntityManagerInterface $entityManager,
+        protected LoggerInterface $logger,
```
NoelLH, line 19: As above

src/Application/Auth/DonationHookAuthMiddleware.php
```diff
@@ -14,11 +15,10 @@ class DonationHookAuthMiddleware implements MiddlewareInterface
...
     use ErrorTrait;
 
-    private LoggerInterface $logger;
-
-    public function __construct(LoggerInterface $logger)
-    {
-        $this->logger = $logger;
+    #[Pure]
+    public function __construct(
+        protected LoggerInterface $logger
```
NoelLH, line 20: Was `private`

src/Application/Auth/DonationPublicAuthMiddleware.php
```diff
@@ -15,11 +16,10 @@ class DonationPublicAuthMiddleware implements MiddlewareInterface
...
     use ErrorTrait;
 
-    private LoggerInterface $logger;
-
-    public function __construct(LoggerInterface $logger)
-    {
-        $this->logger = $logger;
+    #[Pure]
+    public function __construct(
+        protected LoggerInterface $logger
```
NoelLH, line 21: Was `private`

src/Application/Matching/OptimisticRedisAdapter.php
```diff
@@ -19,26 +20,30 @@
...
-    private Redis $redis;
     /**
-     * @var int How many seconds the authoritative source for real-time match funds should keep data, as a minimum.
-     *          Because Redis sets an updated value on each change to the balance, the case where using the database
-     *          value could be problematic (race conditions with high volume access) should not overlap with the case
-     *          where Redis copies of available fund balances are expired and have to be re-fetched.
+     * @param Redis $redis
+     * @param EntityManagerInterface $entityManager
+     * @param array $fundingsToPersist
+     * @param int $maxPartialAllocateTries -
```
NoelLH, line 27: Doc descriptions are backwards

src/Application/Matching/OptimisticRedisAdapter.php
```diff
@@ -19,26 +20,30 @@
...
-
-    public function __construct(Redis $redis, EntityManagerInterface $entityManager)
-    {
-        $this->redis = $redis;
-        $this->entityManager = $entityManager;
+    #[Pure]
+    public function __construct(
+        private Redis $redis,
+        private EntityManagerInterface $entityManager,
+        private array $fundingsToPersist = [],
```
NoelLH, line 40: I'm not sure we make anything clearer by moving these 3 to constructor args, given no code instantiates the class passing them in.

src/Application/Persistence/RetrySafeEntityManager.php
```diff
@@ -15,27 +16,19 @@
...
-    {
-        $this->connectionSettings = $connectionSettings;
-        $this->logger = $logger;
-        $this->ormConfig = $ormConfig;
 
+    public function __construct(
+        private ORM\Configuration $ormConfig,
+        private array $connectionSettings,
+        private LoggerInterface $logger,
+        private int $maxLockRetries = 3
```
NoelLH, line 25: Lost doc for this, and it doesn't need to be / isn't used as a constructor arg?

src/Client/Common.php
```diff
@@ -5,18 +5,20 @@
...
 {
     private array $clientSettings;
-    private Client $httpClient;
-    protected LoggerInterface $logger;
 
-    public function __construct(array $settings, LoggerInterface $logger)
-    {
+    #[Pure]
+    public function __construct(
+        private array $settings,
```
NoelLH, line 17: Not sure this makes sense as we don't use the resulting `$this->settings`

src/Client/Common.php
```diff
@@ -5,18 +5,20 @@
...
+        protected LoggerInterface $logger,
+        private Client $httpClient
```
NoelLH, line 19: This was not a constructor arg before and I think this actually breaks the client instantiation. Have a look at `Client\Campaign`, Client\Donation` etc. in the dependencies – my IDE highlights them as broken because of the new required constructor arg here.

tests/Application/Commands/RetrospectivelyMatchTest.php
```diff
@@ -58,7 +61,7 @@ public function testNonNumericDaysBackRefusesToRun(): void
     public function testNonWholeDaysBackIsRounded(): void
     {
         $commandTester = new CommandTester($this->command);
-        $commandTester->execute(['days-back' => '7.5']);
+        $commandTester->execute(['days-back' => 7.5]);
```
NoelLH, line 64: Inputs from the CLI are always strings unless cast – did this not work with the string value?

tests/Application/Commands/RetrospectivelyMatchTest.php
```diff
@@ -73,7 +76,7 @@ public function testNonWholeDaysBackIsRounded(): void
     public function testWholeDaysBackProceeds(): void
     {
         $commandTester = new CommandTester($this->command);
-        $commandTester->execute(['days-back' => '8']);
+        $commandTester->execute(['days-back' => 8]);
```
NoelLH, line 79: As above

NoelLH (changes requested):
Mostly looks good – suggestions inline

tests/Application/Commands/RetrospectivelyMatchTest.php
tbg-jasonfung: @NoelLH yeah, running tests gives me this error:

```
TypeError: round(): Argument #1 ($num) must be of type int|float, string given
```

NoelLH: @tbg-jasonfung doesn't this suggest the code should be casting to avoid the problem, rather than the test no longer testing for this case which I think can happen? Have you tried running the command for real?

NoelLH (approved):

PR #197, opened 2021-04-07
MAT-162 - improve Stripe Payment Intent create/update error log detail

NoelLH:


NoelLH:
@alihejazi this is a pretty simple MatchBot tweak PR, to get an idea of how this looks and the review interface.

I'll merge this one straight in as I'm keen to get this logging to Production so we can potentially catch the internal code for the errors we're seeing this morning – before the underlying Stripe account issue is fixed – but let me know if any questions or comments on the code changes.

PR #219, opened 2021-04-28
MAT-170 - hopefully fix create-then-update logic for good!

NoelLH:
* for post-error records over 30 seconds old, create-then-update in one go
* for newer ones, log an error but leave them ready to pick up next time


alihejazi (approved):
Looks good.

PR #237, opened 2021-09-02
MAT-179 - fix fee cover transfer amounts

NoelLH:
Also BG2-1871 – fix unit test so we are actually checking the VAT scenarios
diverge as expected

### DIFF
--- a/tests/Application/Fees/CalculatorTest.php
+++ b/tests/Application/Fees/CalculatorTest.php
@@ -5,14 +5,21 @@ declare(strict_types=1);
 namespace MatchBot\Tests\Application\Fees;
 
 use MatchBot\Application\Fees\Calculator;
+use MatchBot\Tests\Application\VatTrait;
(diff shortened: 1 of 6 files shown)


tests/Application/Fees/CalculatorTest.php
```diff
@@ -73,10 +81,17 @@ public function testStripeUSCardSEKDonation(): void
         $this->assertEquals('5.74', $calculator->getCoreFee());
     }
 
-    public function testZeroVatChargedForUsaCharity()
```
NoelLH, line 76: @alihejazi this was not testing what it claimed to before because the default VAT config for unit tests is all zeroed – see `phpunit.xml`.

To test the differing behaviour with realistic VAT configs now that we are live for VAT, you need to give the unit test different settings and have at least one test cover the UK vs. non-UK cases. I have now done that in `CalculatorTest` alongside my other changes.

alihejazi (approved):

PR #249, opened 2021-09-21
MAT-188 Treat non ready campaigns like 404

alihejazi:


.env.example
```diff
@@ -1,39 +0,0 @@
-APP_ENV=local
```
NoelLH, line 1: @alihejazi accidental delete?

alihejazi: @NoelLH oops yes accidental

src/Client/Campaign.php
```diff
@@ -17,6 +17,11 @@ public function getById(string $id): array
     {
         try {
             $response = $this->getHttpClient()->get("{$this->getSetting('campaign', 'baseUri')}/$id");
+
+            if (!json_decode($response->getBody()->getContents(), true)['ready']) {
```
NoelLH, line 21: This part cannot lead to a `RequestException` so it would sit more naturally and leave the code more readable outside the `try/catch`.

Additionally, `json_decode()` of the whole response doesn't need to happen twice. We can assign the existing decode result to a local var and then check that.

I think it would also be prudent to check if the property is set so we don't need to worry about the order changes are released. If the property is absent we should assume we have an older version of the API contract and not exclude the campaign.

src/Client/Campaign.php
```diff
@@ -17,6 +17,11 @@ public function getById(string $id): array
...
+                // uncaught exception
+                throw new CampaignNotReadyException('Campaign not ready');
```
NoelLH, line 23: I'm not sure we gain much from having a distinct exception class here – we want behaviour to be as close to the existing 'not found' case as possible I think, including the sandbox-specific cases we already covered.

If we must have a new Exception class it should extend the existing one so that inheritance leads all the existing catches to continue working as they do now. But given there is a distinct exception message, I'm not convinced we need a separate class at all.

NoelLH (changes requested):

src/Client/Campaign.php
```diff
@@ -26,6 +26,18 @@ public function getById(string $id): array
             throw $exception;
         }
 
-        return json_decode($response->getBody()->getContents(), true);
+        $campaign_as_json = json_decode($response->getBody()->getContents(), true);
```
NoelLH, line 29: @alihejazi I would ideally make this camelCase to match all the other vars' style in the app. Otherwise I think this is good to go

PR #269, opened 2021-11-30
MAT-199 - make data Salesforce-friendly for LastName (required) as well as FirstName (optional)

NoelLH:


src/Domain/Donation.php
```diff
@@ -462,19 +460,31 @@ public function setChampionComms(?bool $championComms): void
...
     }
 
-    public function getDonorFirstName(): ?string
+    public function getDonorFirstName(bool $salesforceSafe = false): ?string
     {
-        return $this->donorFirstName;
+        $firstName = $this->donorFirstName;
+
+        if ($salesforceSafe) {
+            $firstName = $this->makeSalesforceSafe($firstName, false);
```
alihejazi, line 468: @NoelLH I'm concerned that passing `false` as a parameter variable here (meaning not required)  will still return a blank name ' ' or null name instead of "N/A" for the first name when it's blank or null. If this is the intended behavior then I'm happy to merge.

alihejazi: @NoelLH Though perhaps for reconciliation, it may be easier to populate the first names with "N/A" too so we can search for these affected donations more easily in the future, and subsequently fix them?
I may be overthinking this. Would love to hear your thoughts.

NoelLH: @alihejazi given we've gone several campaigns with blank first names before I think minimising the scale of the change should be the priority at such a critical time. We can also query for ' ' or null just as easily as 'N/A'?

alihejazi (approved):

PR #338, opened 2022-03-01
MAT-200 - try to recover Doctrine EntityManager connection

NoelLH:
When we hit campaign creation race condition in Create action


NoelLH:
@alihejazi please see Jira comments too: I have now fleshed them out with context and my thinking around the options at each stage of the dev process.

alihejazi (approved):

PR #339, opened 2022-03-02
MAT-233 PRD noise tweaks

alihejazi:


src/Application/Auth/DonationRecaptchaMiddleware.php
```diff
@@ -29,34 +29,45 @@ public function __construct(
...
 
-        $result = $this->captcha->verify(
-            $captchaCode,
-            $request->getAttribute('client-ip') // Set to original IP by previous middleware
-        );
+            if ($result->isSuccess()) {
+                $response = $handler->handle($request);
+                break;
+            } else {
+                $this->logger->info('Security: captcha failed, attempt: ' . ($counter + 1));
```
NoelLH, line 63: I think this should be at least warning level when it happens a second time. But I think there is also a more fundamental change needed for the callouts – more on the overall review comment + Jira

NoelLH (changes requested):
@alihejazi I took a more detailed look at the error behaviour yesterday. We had 36 timeouts / connection drops just from ~1pm to midnight. This is extremely high and suggests the approach is not fit for purpose even with retries.

I next took a look at https://stackoverflow.com/questions/4189973/file-get-contents-connection-timed-out and https://stackoverflow.com/a/68830933/2803757 which points out there is another arg we can pass to ReCaptcha (not dependent on using the Buzz library mentioned, we already share the underlying one) which can tell it to use cURL. This should be more reliable and timeout can probably be controlled / aligned with where we use Guzzle. I believe we already have `ext-curl` as a Guzzle dependency so it should be easy to set this up in our `dependencies.php`.

NoelLH (approved):

PR #358, opened 2022-04-02
MAT-239 – expand Gift Aid claim result handling

NoelLH:


src/Application/Messenger/Handler/GiftAidResultHandler.php
```diff
@@ -20,14 +20,28 @@ public function __construct(
...
-            'Donation ID %s Gift Aid claim failure reported by ClaimBot',
+            'Donation ID %s Gift Aid claim result returned by ClaimBot',
             $donationMessage->id,
         ));
 
         /** @var Donation $donation */
         $donation = $this->donationRepository->findOneBy(['uuid' => $donationMessage->id]);
 
-        $donation->setTbgGiftAidRequestFailedAt(new \DateTime());
+        if ($donationMessage->responseSuccess === false) {
```
alihejazi, line 30: `if (!$donationMessage->responseSuccess)`

src/Application/Messenger/Handler/GiftAidResultHandler.php
```diff
@@ -20,14 +20,28 @@ public function __construct(
...
+            $donation->setTbgGiftAidRequestFailedAt(new \DateTime());
+        }
+
+        if ($donationMessage->responseSuccess === true) {
```
alihejazi, line 34: `=== true` is redundant

alihejazi (approved):

src/Application/Messenger/Handler/GiftAidResultHandler.php
NoelLH: @alihejazi this is not the same. We need to explicitly ensure that it is `false`, not `null`.

src/Application/Messenger/Handler/GiftAidResultHandler.php
NoelLH: I think it is still clearer when we're dealing with a nullable type and one that's so fundamental.

PR #419, opened 2022-09-12
MAT-109 create Stripe Customer and link PaymentIntent on donation-create

alihejazi:


NoelLH:
Closing as we make customers in ID now

PR #420, opened 2022-09-13
MAT-109 – supporting setting existing Stripe Customer ID

NoelLH:
~~Todo:~~

- [x] DB migrations reflecting model update
- [x] Addition of JWT requirement for setting a Customer ID (inc. new unit test for that mode)
- [x] Conditional removal of captcha verify code requirement when JWT's set

### DIFF
--- a/src/Application/Actions/Donations/Create.php
+++ b/src/Application/Actions/Donations/Create.php
@@ -10,7 +10,8 @@ use JetBrains\PhpStorm\Pure;
 use MatchBot\Application\Actions\Action;
 use MatchBot\Application\Actions\ActionError;
 use MatchBot\Application\Actions\ActionPayload;
-use MatchBot\Application\Auth\Token;
+use MatchBot\Application\Auth\DonationToken;
+use MatchBot\Application\Auth\PersonManagementAuthMiddleware;
 use MatchBot\Application\HttpModels\DonationCreate;
 use MatchBot\Application\HttpModels\DonationCreatedResponse;
 use MatchBot\Domain\Campaign;
@@ -185,7 +186,7 @@ class Create extends Action
 
         $response = new DonationCreatedResponse();
         $response->donation = $donation->toApiModel();
-        $response->jwt = Token::create($donation->getUuid());
+        $response->jwt = DonationToken::create($donation->getUuid());
 
         // Attempt immediate sync. Buffered for a future batch sync if the SF call fails.
         $this->donationRepository->push($donation, true);
--- a/src/Application/Actions/Donations/Create.php
+++ b/src/Application/Actions/Donations/Create.php
@@ -93,7 +93,11 @@ class Create extends Action
         }
 
         if ($customerId !== $donation->getPspCustomerId()) {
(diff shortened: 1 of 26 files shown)


src/Application/Actions/Donations/Create.php
```diff
@@ -84,6 +92,14 @@ protected function action(): Response
             $donation = $this->donationRepository->buildFromApiRequest($donationData);
```
alihejazi, line 92: @NoelLH This line seems redundant. Isn't it already in the `try` block above?

src/Application/Actions/Donations/Create.php
```diff
@@ -84,6 +92,14 @@ protected function action(): Response
...
         }
 
+        if ($customerId !== $donation->getPspCustomerId()) {
```
alihejazi, line 95: @NoelLH It seems redundant to me that we're passing the `customerId` both in the route parameters and in the request's body. Why not just in one place? In that case, I'm not sure we'd need this `if` block and instead I think we could rely on the middleware verifying that the incoming request to create a donation does indeed belong to the `customerId `specified when creating that donation.

src/Application/Actions/Donations/Create.php
```diff
@@ -40,9 +41,16 @@ public function __construct(
...
      * @throws \MatchBot\Application\Matching\TerminalLockException if the matching adapter can't allocate funds
+     * @see PersonManagementAuthMiddleware
      */
     protected function action(): Response
     {
+        // The route at `/people/{personId}/donations` validates that the donor has permission to act
+        // as the person, and sets this attribute to the Stripe Customer ID based on JWS claims, all
+        // in `PersonManagementAuthMiddleware`. If the legacy route was used or if no such ID was in the
+        // JWS, this is null.
+        $customerId = $this->request->getAttribute('pspId');
```
alihejazi, line 52: @NoelLH I might be mistaken but shouldn't this be `$this->request->getAttribute('personId');`?

src/Application/Actions/Donations/Create.php
NoelLH: @alihejazi It's a re-try, as noted in the comment. If it failed the first time then `$donation` is undefined. This is also not changing in this PR.

src/Application/Actions/Donations/Create.php
NoelLH: @alihejazi I think this redundancy is a sensible trade-off because the serialisation of the data object relies on `pspCustomerId` being available and persisted just like any other property. Removing it from the object definition means way more variation and complexity around this compared to other properties.

Contra-wise, removing it from the route means much more opaque (and therefore risky), messy and slow authentication middleware behaviour.

src/Application/Actions/Donations/Create.php
NoelLH: @alihejazi Nope, don't think so. Check out the `$request->withAttribute(...)` call and logic leading up to it in the middleware.

alihejazi (approved):

PR #422, opened 2022-09-23
MAT-264 – add List Payment Methods

NoelLH:


src/Application/Auth/PersonManagementAuthMiddleware.php
```diff
@@ -37,18 +43,26 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
-        return $handler->handle($request->withAttribute('pspId', IdentityToken::getPspId($jws)));
+        $this->checkCompleteness($request);
+
+        return $handler->handle($request->withAttribute('pspId', IdentityToken::getPspId($this->jws)));
+    }
+
+    protected function checkCompleteness(ServerRequestInterface $request): void
+    {
+        // No-op: we allow both values for this base middleware. This should be extended
+        // by middlewares that have tighter requirements.
```
alihejazi, line 66: @NoelLH If that's the case, then perhaps we should make this middleware class an abstract class?

alihejazi (approved):

src/Application/Auth/PersonManagementAuthMiddleware.php
NoelLH: @alihejazi This cannot be abstract because it is invoked directly too, for `Donations\Create`.

The idea is that this general version of the middleware is less constrained, but its subclasses can be more so.

If we made it abstract and extended it for `Donations\Create`, we would still have to make the subclass implement the method as a no-op, so as I see it that would be more complexity and an extra class for no real benefit.

PR #428, opened 2022-10-11
Log spacing tweak

NoelLH:
Fix naive ".ERROR" CloudWatch Logs Insights / filters matching "...error"


NoelLH:
Included in https://github.com/thebiggive/matchbot/pull/429

PR #449, opened 2022-11-28
MAT-274 – support retrying Payment Intent update on Stripe object lock errors

NoelLH:


tests/Application/Actions/Donations/UpdateTest.php
```diff
@@ -1218,6 +1219,210 @@ public function testAddDataHitsAlreadyCapturedStripeExceptionWithFeeChange(): vo
...
+        $donation->setTbgComms(true);
+        $donation->setCharityComms(false);
+        $donation->setChampionComms(false);
+        $donation->setDonorBillingAddress('Y1 1YX');
+
+        $donationRepoProphecy = $this->prophesize(DonationRepository::class);
+        $donationRepoProphecy
+            ->findOneBy(['uuid' => '12345678-1234-1234-1234-1234567890ab'])
+            ->willReturn($this->getTestDonation()) // Get a new mock object so DB has old values.
+            ->shouldBeCalledOnce();
```
bdsl, line 1251: shouldBeCalledOnce is probably unnecessary - generally I expect we want [Mocks for Commands, Stubs for Queries](https://blog.ploeh.dk/2013/10/23/mocks-for-commands-stubs-for-queries/l)
```suggestion
            ->willReturn($this->getTestDonation()); // Get a new mock object so DB has old values.
```

tests/Application/Actions/Donations/UpdateTest.php
```diff
@@ -1218,6 +1219,210 @@ public function testAddDataHitsAlreadyCapturedStripeExceptionWithFeeChange(): vo
...
+            ->deriveFees(Argument::type(Donation::class), null, null)
+            ->willReturn($donation) // Actual fee calculation is tested elsewhere.
+            ->shouldBeCalledOnce();
+
+        // Persist as normal.
+        $entityManagerProphecy = $this->prophesize(EntityManagerInterface::class);
+        $entityManagerProphecy->beginTransaction()->shouldBeCalledOnce();
+        $entityManagerProphecy->persist(Argument::type(Donation::class))->shouldBeCalledOnce();
+        $entityManagerProphecy->flush()->shouldBeCalledOnce();
+        $entityManagerProphecy->commit()->shouldBeCalledOnce();
```
bdsl, line 1265: Why are we mocking out both the repositories and the entity manager? It seems a bit confusing to be mocking out two layers at once. I guess there must be some parts of the code under test using the repositories and other parts going straight through to the EM.

tests/Application/Actions/Donations/UpdateTest.php
```diff
@@ -1218,6 +1219,210 @@ public function testAddDataHitsAlreadyCapturedStripeExceptionWithFeeChange(): vo
...
+        $this->assertEquals(500, $response->getStatusCode());
+
+        $payloadArray = json_decode($payload, true);
+        $this->assertEquals([
+            'error' => [
+                'type' => 'SERVER_ERROR',
+                'description' => 'Could not update Stripe Payment Intent [C]',
+            ],
+        ], $payloadArray);
+    }
```
bdsl, line 1424: at 100 lines this is IMHO quite long and hard to follow as a test. Partly that's because I'm new to the codebase, partly I think it is awkwardly long. Could do some extract function refactorings as an easy way to get it a bit more compact. Same applies to previous test.

Also looks like there's probably quite a bit of code in common between the two tests, so factoring that out might make it easier to see what the differences are.

src/Application/Actions/Donations/Update.php
```diff
@@ -194,35 +195,32 @@ private function addData(Donation $donation, HttpModels\Donation $donationData):
...
+            } catch (RateLimitException $exception) {
+                if ($exception->getStripeCode() !== 'lock_timeout') {
+                    throw $exception; // Only re-try when object level lock failed.
+                }
+
+                $this->logger->info(sprintf(
+                    'Stripe lock "rate limit" hit while updating payment intent for donation %s – retrying in 1s...',
+                    $donation->getId(),
+                ));
+                sleep(1);
```
bdsl, line 208: this will work for now, but we may want to change it soon, e.g. to make the sleep duration injected so we don't have to wait a second when running the test for this, or to use a queue system to do the updates.

src/Application/Actions/Donations/Update.php
```diff
@@ -377,4 +379,37 @@ private function save(Donation $donation): void
         // sending the updated status to Salesforce in a future batch sync.
         $this->donationRepository->push($donation, false);
     }
+
+    /**
+     * @throws RateLimitException if e.g. Stripe had locked the PI while processing.
+     */
+    private function updatePaymentIntent(Donation $donation): void
```
bdsl, line 386: I see that this is a straightforward extract function. :+1:

bdsl (approved):
Added some comments inline, no blockers. LGTM.

alihejazi (approved):
Looks good to me.

tests/Application/Actions/Donations/UpdateTest.php
NoelLH: @bdsl Yes, that's why. You can see in the corresponding Action's constructor where this happens, and I think its (`Update`'s) code shows why it's quite a helpful choice.

tests/Application/Actions/Donations/UpdateTest.php
bdsl: Agree a bit of duplication is generally much better than a bit of unclear abstraction. I'm not really bothered by duplication per-se, just by a test method being too long and complex to easily understand as a whole.

tests/Application/Actions/Donations/UpdateTest.php
bdsl: Maybe we can discuss it if it comes up again on another PR when we're less pushed for time. I think the mocks for commands, stubs for queries principle is very helpful when writing or editing tests, but it's unlikely to be worthwhile to edit old tests purely for the sake of following the principle.

PR #451, opened 2022-11-29
MAT-276 – fix log error noise in new edge case of an edge case

NoelLH:
When Stripe object lock acquisition takes two goes, and then the second Payment Intent update fails due to capture being done, and we have no fee change to send, this can just be an `INFO` log.

### DIFF
--- a/src/Application/Actions/Donations/Update.php
+++ b/src/Application/Actions/Donations/Update.php
@@ -209,70 +209,30 @@ class Update extends Action
 
                 try {
                     $this->updatePaymentIntent($donation);
-                } catch (RateLimitException $exception) {
+                } catch (RateLimitException $retryException) {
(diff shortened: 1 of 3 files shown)


src/Application/Actions/Donations/Update.php
```diff
@@ -412,4 +372,64 @@ private function updatePaymentIntent(Donation $donation): void
             // Note that `on_behalf_of` is set up on create and is *not allowed* on update.
         ]);
     }
+
+    /**
+     * @return Response? Response to send client, if appropriate. HTTP 500.
```
bdsl, line 377:
```suggestion
     * @return ?Response Response to send client, if appropriate. HTTP 500.
```

bdsl (approved):

PR #453, opened 2022-11-30
MAT-260: Lock donations to prevent concurrent status updates

bdsl:
I've tested donating on my local, works as expected. Will need some more testing in staging of course.

### DIFF
--- a/src/Domain/DonationRepository.php
+++ b/src/Domain/DonationRepository.php
@@ -670,4 +670,17 @@ class DonationRepository extends SalesforceWriteProxyRepository
             $this->getEntityManager()->persist($donation);
         }
     }
+
+    /**
+     * Locks row in DB to prevent concurrent updates
+     */
+    public function findOneBy(array $criteria, ?array $orderBy = null): ?Donation
+    {
+        $donation = parent::findOneBy($criteria, $orderBy);
+
+        // lock to prevent concurrent status changes. See jira MAT-260
+        $this->find($donation->getId(), LockMode::PESSIMISTIC_WRITE);
+
+        return $donation;
+    }
 }
--- a/src/Domain/DonationRepository.php
+++ b/src/Domain/DonationRepository.php
@@ -7,6 +7,7 @@ namespace MatchBot\Domain;
 use DateTime;
 use Doctrine\Common\Cache\CacheProvider;
 use Doctrine\DBAL\Exception as DBALException;
+use Doctrine\DBAL\LockMode;
 use GuzzleHttp\Exception\ClientException;
 use MatchBot\Application\Fees\Calculator;
 use MatchBot\Application\HttpModels\DonationCreate;
@@ -674,13 +675,15 @@ class DonationRepository extends SalesforceWriteProxyRepository
     /**
      * Locks row in DB to prevent concurrent updates
      */
-//    public function findOneBy(array $criteria, ?array $orderBy = null): ?Donation
-//    {
-//        $donation = parent::findOneBy($criteria, $orderBy);
-//
-//        // lock to prevent concurrent status changes. See jira MAT-260
-//        $this->find($donation->getId(), LockMode::PESSIMISTIC_WRITE);
-//
-//        return $donation;
-//    }
+    public function findAndLockOneBy(array $criteria, ?array $orderBy = null): ?Donation
+    {
+        $donation = $this->findOneBy($criteria, $orderBy);
+
+        // lock to prevent concurrent status changes. See jira MAT-260
+        // Donation is already in Doctrine identity map so this won't actually reload it from the DB
+        // and we don't need the return value.
+        $this->find($donation->getId(), LockMode::PESSIMISTIC_WRITE);
+
+        return $donation;
+    }
 }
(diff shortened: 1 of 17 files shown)


bdsl:
I've been testing this using my local frontend, identity and matchbot, with help from @alihejazi Currently getting an error ` "Entity 'MatchBot\\Domain\\Campaign' has no field ''. You can therefore not call 'findOneBy' on the entities' repository."`. The confusing thing for me is that it has a stack trace that includes DonationRepository line 135, ( https://github.com/thebiggive/matchbot/blob/d8b4cd182dc84fe044a5e66d17d18b2e0d16985f/src/Domain/DonationRepository.php#L135 ) which is not changed in this PR.

bdsl:
I've checked out the `develop` branch on my local, but I'm still getting the same error. Not sure if I'm not testing right.

bdsl:
codecov is complaining about lack of coverage in new code, but since it's about persistence I'm not sure there's anything to meaningfully unit test.

src/Domain/DonationRepository.php
```diff
@@ -670,4 +671,19 @@ private function persistQueuedDonations(): void
...
+     */
+    public function findAndLockOneBy(array $criteria, ?array $orderBy = null): ?Donation
+    {
+        $donation = $this->findOneBy($criteria, $orderBy);
+
+        // lock to prevent concurrent status changes. See jira MAT-260
+        // Donation is already in Doctrine identity map so this won't actually reload it from the DB
+        // and we don't need the return value.
+        $this->find($donation->getId(), LockMode::PESSIMISTIC_WRITE);
+
```
bdsl, line 686: Hmm I think this isn't quite enough. If two requests come in at the same time, the lock will make them one of them pause at line 685. But the data has already been loaded from the DB at line 680, so when it acquires the lock and continues it will still have that stale data.

Probably we just need to add this:
```suggestion
        $this->_em->refresh($donation);
        
```

bdsl: (added)

src/Domain/DonationRepository.php
```diff
@@ -670,4 +671,20 @@ private function persistQueuedDonations(): void
             $this->getEntityManager()->persist($donation);
         }
     }
+
+    /**
+     * Locks row in DB to prevent concurrent updates
+     */
+    public function findAndLockOneBy(array $criteria, ?array $orderBy = null): ?Donation
+    {
+        $donation = $this->findOneBy($criteria, $orderBy);
```
NoelLH, line 680: @bdsl I think it would be good here just to add a code comment pointing out the reason to first select without a lock, for quick reference.

tests/HttpRequests/create-donation.http
```diff
@@ -0,0 +1,4 @@
+POST http://localhost:30030/v1/donations
```
NoelLH, line 1: What's this?

NoelLH (commented):
@bdsl I think broadly this approach is looking promising. My main concern is that it seems risky to lock in _all_ (or many) supported scenarios when we could benefit from a fix being out soon and don't really have time spare to load test the impact.

In particular, I'm thinking that things like locking for a `GET` donor request could have implications like slowing down or breaking Stripe hooks unnecessarily, when really the impact of slightly stale info being returned to the donor here is minimal. I would rather start with the fewest number of locks that we know are needed to avoid problems, so that we are being as conservative as we can be from a performance and lock contention perspective.

(I can't comment on a 0 line file, but I'm also not clear what the empty `public/test.php` is doing?)

tests/HttpRequests/create-donation.http
bdsl: > @bdsl I think broadly this approach is looking promising. My main concern is that it seems risky to lock in _all_ (or many) supported scenarios when we could benefit from a fix being out soon and don't really have time spare to load test the impact.

That's reasonable - do you think actually we may just need to get the lock from the stripe callback handler? It's a lock enforced by the DB not just an advisory lock we I suppose we only need to acquire it from one side of the possibly conflicting processes.

> (I can't comment on a 0 line file, but I'm also not clear what the empty `public/test.php` is doing?)

Sorry I must have committed that by mistake, I was running phpInfo from that file to work out how to use xdebug.

bdsl: This http file is something we can run in e.g. PhpStorm to manually test the create donation handler. Very minimal at the moment but may be useful. Or I can delete it if you think it will cause a problem.

NoelLH: @bdsl I'm not totally sure to be honest. If going for just one side, my first instinct would be to lock in the donation API's `PUT` because we know that that's where the "bad" update can occur in terms of unexpectedly reverting the donation status – but it could be you're right that either is sufficient so long as both endpoints continue to use transactions that include the `find*(...)`.

It doesn't seem too likely to be that locking in both these places would increase the chance of failure for donors vs. locking in one, so perhaps we could try with the two for now and make a ticket to come back to assessing the minimum that works more rigorously after the campaign?

src/Domain/DonationRepository.php
bdsl: Done.

tests/HttpRequests/create-donation.http
bdsl: Makes sense. I've reduced the amount we're using locking to just the update and stripe handler in https://github.com/thebiggive/matchbot/pull/453/commits/40f17efccd7bcc9632984e6d16238fba1cdb4917

NoelLH: Nice, thanks @bdsl! LGTM

PR #454, opened 2022-12-03
MAT-278 DB Migration

alihejazi:


NoelLH (approved):
Technically we don't need the empty args, but that shouldn't break anything. Looks good.

PR #455, opened 2022-12-03
MAT-260 – fix null handling in new repo helper

NoelLH:
Missing donations are expected e.g. when a sandbox is handling Stripe hooks, as Stripe have only 1 test mode. And in all cases, a missing item in the first find call should return `null` instead of crashing.

### DIFF
--- a/src/Domain/DonationRepository.php
+++ b/src/Domain/DonationRepository.php
@@ -679,14 +679,18 @@ class DonationRepository extends SalesforceWriteProxyRepository
     {
         // We can't actually lock the row until we know the ID of the donation, so we fetch it first
         // using the criteria, and then call find once we know the ID to lock.
-
+        /** @var Donation|null $donation */
         $donation = $this->findOneBy($criteria, $orderBy);
+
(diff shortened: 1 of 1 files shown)


alihejazi (approved):

bdsl:
LGTM, apologies for the mistake here. 

This is the sort of issue that Psalm will prevent if we set it to catch [PossiblyNullReference/](https://psalm.dev/docs/running_psalm/issues/PossiblyNullReference)s.

NoelLH:
All good @bdsl. In the mean time, in my IntelliJ IDE I get orange highlights when this happens, but I'm not sure if it's a built-in inspection config thing or a plugin doing that one. I definitely appreciate that it's better to guarantee the check happens regardless of individual devs' editors.

bdsl:
Thanks for the reminder, I've now installed the [Php Inspections ​(EA Extended)​](https://plugins.jetbrains.com/plugin/7622-php-inspections-ea-extended-) plugin in my PhpStorm. Highly reccomended.

image

PR #457, opened 2022-12-03
MAT-260 – fix status-reverted donation's data

NoelLH:


bdsl (approved):
Code looks fine, I haven't double checked the actual salesforceId or the reason for doing this.

bdsl (approved):

PR #462, opened 2022-12-13
Mat 282

bdsl:


src/Migrations/Version20221213144526.php
```diff
@@ -0,0 +1,28 @@
...
+declare(strict_types=1);
+
+namespace MatchBot\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
```
bdsl, line 12:
```suggestion
```

src/Migrations/Version20221213144526.php
```diff
@@ -0,0 +1,25 @@
...
+final class Version20221213144526 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Remove gift aid added to donations by mistake';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('UPDATE Donation set giftAid = 0 where Donation.salesforceId IN ("a066900001vWmAaAAK", "a066900001vWlRHAA0", "a066900001vVkwZAAS") LIMIT 3');
```
alihejazi, line 19: @bdsl I think you might be missing some more fields to update. See the similar PR made for MAT-245:
https://github.com/thebiggive/matchbot/pull/382/files

bdsl: @alihejazi Thanks. https://github.com/thebiggive/matchbot/pull/462/commits/84f43057f5cf913d148e0427d73488108feeb55a should fix it.

src/Migrations/Version20221213144526.php
```diff
@@ -0,0 +1,30 @@
...
+    {
+        return 'Remove gift aid added to donations by mistake';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql(<<<'SQL'
+                UPDATE Donation 
+                set giftAid = 0, tipGiftAid = 0, salesforcePushStatus = 'pending-update'
+                WHERE Donation.salesforceId IN ('a066900001vWmAaAAK', 'a066900001vWlRHAA0', 'a066900001vVkwZAAS') LIMIT 3
```
alihejazi, line 22:
```suggestion
                WHERE salesforceId IN ('a066900001vWmAaAAK', 'a066900001vWlRHAA0', 'a066900001vVkwZAAS') LIMIT 3
```

src/Migrations/Version20221213144526.php
```diff
@@ -0,0 +1,30 @@
...
+            SQL
+        );
+    }
+    public function down(Schema $schema): void
+    {
+        throw new \Exception('no going back');
```
alihejazi, line 28: I don't think there's a need for an exception
```suggestion
```

src/Migrations/Version20221213144526.php
bdsl: We don't need it, it works either way doesn't it? This is just slightly more explicit.

alihejazi: Ok

src/Migrations/Version20221213144526.php
alihejazi: Ok

alihejazi (approved):

PR #465, opened 2022-12-14
MAT-275: Optimize query in \MatchBot\Domain\DonationRepository::findW…

bdsl:
…ithExpiredMatching

Grouping by the donation ID, instead of by all fields of donation allows MySQL to use the existing index on ID for grouping purposes.

Id is a primary key, so we know that when the ID matches all fields of donation match.

### DIFF
--- a/src/Domain/DonationRepository.php
+++ b/src/Domain/DonationRepository.php
@@ -406,7 +406,7 @@ class DonationRepository extends SalesforceWriteProxyRepository
             ->leftJoin('d.fundingWithdrawals', 'fw')
             ->where('d.donationStatus = :expireWithStatus')
             ->andWhere('d.createdAt < :expireBefore')
-            ->groupBy('d')
+            ->groupBy('d.id')
             ->having('COUNT(fw) > 0')
             ->setParameter('expireWithStatus', 'Pending')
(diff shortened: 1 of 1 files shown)


bdsl:
Not sure what to do about the failing Codecov check - I don't think this can be meaningfully unit tested, but I don't like leaving things on red.

src/Domain/DonationRepository.php
```diff
@@ -406,7 +406,7 @@ public function findWithExpiredMatching(): array
             ->leftJoin('d.fundingWithdrawals', 'fw')
             ->where('d.donationStatus = :expireWithStatus')
             ->andWhere('d.createdAt < :expireBefore')
-            ->groupBy('d')
+            ->groupBy('d.id')
```
alihejazi, line 409: This looks fine but is there any chance you can test it works fine locally?

alihejazi (approved):

src/Domain/DonationRepository.php
bdsl: I have run the command locally using 
```
# SLACK_API_TOKEN=xoxb-token SLACK_CHANNEL=channel php matchbot-cli.php matchbot:expire-match-funds
```

and it seems to run as normal.

PR #467, opened 2022-12-20
Mat-280 Retry non-cancellation updates to donations in case of DB lock conflict

bdsl:


composer.json
```diff
@@ -24,7 +24,7 @@
         "doctrine/annotations": "^1.13",
         "doctrine/cache": "^1.11.3",
         "doctrine/migrations": "^3.3",
-        "doctrine/orm": "^2.8",
+        "doctrine/orm": "^2.14.0",
```
bdsl, line 27: see https://github.com/doctrine/orm/releases/tag/2.14.0

src/Application/Actions/Donations/Update.php
```diff
@@ -75,6 +78,9 @@ protected function action(): Response
                 'json'
             );
         } catch (UnexpectedValueException | TypeError $exception) {
+
+        //    echo $exception; die();
+
```
bdsl, line 83:
```suggestion
```

tests/Application/Actions/Donations/UpdateHandlesLockExceptionTest.php
```diff
@@ -136,4 +175,31 @@ public function request($method, $path, $params, $opts)
...
+        $donation->createdNow();
+        $donation->setDonationStatus('Pending');
+        $donation->setCampaign($campaign);
+        $donation->setPsp('stripe');
+        $donation->setCurrencyCode('GBP');
+        $donation->setAmount('1');
+        $donation->setUuid(Uuid::uuid4());
+        $donation->setDonationStatus('Pending');
+        $donation->setDonorFirstName('Donor first name');
+        $donation->setDonorLastName('Donor last name');
```
bdsl, line 202:
```suggestion
        $donation->setDonorLastName('Donor last name');

```

tests/Application/Actions/Donations/UpdateHandlesLockExceptionTest.php
```diff
@@ -136,4 +170,63 @@ public function request($method, $path, $params, $opts)
 
         return $stripeClient;
     }
+
+    /**
+     * @return Donation
+     */
```
bdsl, line 176:
```suggestion
```

src/Application/Actions/Donations/Update.php
```diff
@@ -124,9 +126,23 @@ protected function action(): Response
...
+                    $this->entityManager->refresh($donation, LockMode::PESSIMISTIC_WRITE);
+                }
+                return $this->addData($donation, $donationData);
+            } catch (LockWaitTimeoutException $lockWaitTimeoutException) {
+                \usleep(100_000 * (2 ** $retryCount)); // pause for 0.1, 0.2, 0.4 and then 0.8s before giving up.
+                $retryCount++;
+            }
+        }
+        throw new \Exception(
+            'Retry count exceeded trying to update donation, retried ' . self::MAX_UPDATE_RETRY_COUNT . " times",
```
alihejazi, line 142: @bdsl I think it'd be nice to mention the donation id in this exception message, so we as devs don't have to dig deeper into logs to find the affected donation id.

Also, I think you should be logging the `$retryCount` instead of `self::MAX_UPDATE_RETRY_COUNT`. It's much more truthful. Imagine we introduce a bug without knowing that caused the retries to happen 10 times where `self::MAX_UPDATE_RETRY_COUNT` was still 4. In such a case, the exception would incorrectly log we retried 4 times, when in fact the truth would be 10 times. The incorrect log in such an instance would add much more confusion when trying to debug what happened wrong, let alone noticing the issue in the first place.

tests/Application/Actions/Donations/UpdateHandlesLockExceptionTest.php
```diff
@@ -136,4 +170,60 @@ public function request($method, $path, $params, $opts)
 
         return $stripeClient;
     }
+
+    public function getDonation(): Donation
```
alihejazi, line 174: @bdsl Nice move externalising this into its own method to setup the test donation.

src/Application/Actions/Donations/Update.php
bdsl: Two good ideas.

```suggestion
            "Retry count exceeded trying to update donation #{$donation->getId()} , retried $retryCount times",
```

alihejazi: Nice one @bdsl!

alihejazi (approved):

PR #468, opened 2022-12-23
Use new v2.x of `ramsey/uuid-doctrine` instead of our fork

NoelLH:
Changes from https://github.com/ramsey/uuid-doctrine/pull/180 are now released as per https://github.com/ramsey/uuid-doctrine/issues/179

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "middlewares/client-ip": "^2.0",
         "monolog/monolog": "^2.2",
         "php-di/php-di": "^6.3",
-        "ramsey/uuid-doctrine": "dev-support-other-ems",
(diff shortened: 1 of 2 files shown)


bdsl:
LGTM. I suppose when this is in production https://github.com/thebiggive/uuid-doctrine could be deleted or archived.

bdsl (approved):

NoelLH:
Fork of the lib's now archived @bdsl.

Could you please delete feature branches after merge if they're not likely being reused?

NoelLH:
@bdsl This is what I was thinking. It would be annoying for projects to end up auto-deleting branches we use frequently, even if it is reversible.

PR #473, opened 2023-01-10
MAT-275 – OPTIMIZE Donation and FundingWithdrawal

NoelLH:


NoelLH:
> Looks worth a try. Approving now but just in case, do we know for sure this won't lock the tables for more than a few seconds? I see the docs say "OPTIMIZE TABLE is performed online for regular and partitioned InnoDB tables. Otherwise, MySQL locks the table during the time OPTIMIZE TABLE is running."

@bdsl I'm not sure to be honest and don't know how we would establish this with confidence without spending a lot of time on making a copy of Production. I think it's probably worth chancing it so long as we run when not many big campaigns are open.

PR #479, opened 2023-01-20
MAT-289 – Stripe payout.paid message consumer fix and logging config improvement

NoelLH:
* Stripe payout.paid message consumer: use database transactions as required for object locks
* Symfony Messenger general: Provide a non-null logger to message middlewares

### DIFF
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -129,7 +129,7 @@ return function (ContainerBuilder $containerBuilder) {
             return $factory;
         },
 
-        LoggerInterface::class => function (ContainerInterface $c) {
+        LoggerInterface::class => function (ContainerInterface $c): Logger {
             $settings = $c->get('settings');
 
             $loggerSettings = $settings['logger'];
@@ -155,6 +155,9 @@ return function (ContainerBuilder $containerBuilder) {
         },
 
         MessageBusInterface::class => static function (ContainerInterface $c): MessageBusInterface {
+            /** @var LoggerInterface $logger */
+            $logger = $c->get(LoggerInterface::class);
+
             $sendMiddleware = new SendMessageMiddleware(new SendersLocator(
                 [
                     Messages\Donation::class => [ClaimBotTransport::class],
@@ -162,7 +165,7 @@ return function (ContainerBuilder $containerBuilder) {
                 ],
                 $c,
             ));
-            $sendMiddleware->setLogger($c->get(LoggerInterface::class));
+            $sendMiddleware->setLogger($logger);
 
             $handleMiddleware = new HandleMessageMiddleware(new HandlersLocator(
                 [
@@ -170,7 +173,7 @@ return function (ContainerBuilder $containerBuilder) {
                     StripePayout::class => [$c->get(StripePayoutHandler::class)],
                 ],
             ));
-            $handleMiddleware->setLogger($c->get(LoggerInterface::class));
+            $handleMiddleware->setLogger($logger);
 
             return new MessageBus([
                 $sendMiddleware,
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -155,20 +155,26 @@ return function (ContainerBuilder $containerBuilder) {
         },
 
         MessageBusInterface::class => static function (ContainerInterface $c): MessageBusInterface {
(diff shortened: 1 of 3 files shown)


src/Application/Messenger/Handler/StripePayoutHandler.php
```diff
@@ -153,8 +159,11 @@ public function __invoke(StripePayout $payout): void
             }
         }
 
-        if ($count > 0) {
+        // Ensure we always close a transaction if we opened one, even if no charge IDs
+        // were applicable to update.
+        if (count($chargeIds) > 0) {
             $this->entityManager->flush();
+            $this->entityManager->commit();
         }
```
bdsl, line 167: @NoelLH what do you think about using `EntityManagerInterface::wrapInTransaction` instead of `beginTransaction`? Would mean we wouldn't have to remember to put the `commit` in. Or we could use transactional, which we have our implementation of in `\MatchBot\Application\Persistence\RetrySafeEntityManager::transactional`, but it is for some reason deprecated by Doctrine.

image

(no need for flush, wrapInTransaction does that too)

src/Application/Messenger/Handler/StripePayoutHandler.php
```diff
@@ -113,7 +113,13 @@ public function __invoke(StripePayout $payout): void
             return;
         }
 
-        foreach ($this->getOriginalDonationChargeIds($paidChargeIds, $connectAccountId, $payoutCreated) as $chargeId) {
+        $chargeIds = $this->getOriginalDonationChargeIds($paidChargeIds, $connectAccountId, $payoutCreated);
+
+        if (count($chargeIds) > 0) {
```
bdsl, line 118: Might be a bit clearer with the if reversed to make a guard condition like we have above, instead of wrapping the beginTransaction:

```php
        if ($chargeIds === []) {
            $this->logger->info(sprintf(
                'Payout: Exited with no original donation charge IDs for Payout ID %s',
                $payoutId,
            ));
            
            return;
        }
```

src/Application/Messenger/Handler/StripePayoutHandler.php
```diff
@@ -153,8 +159,11 @@ public function __invoke(StripePayout $payout): void
             }
         }
 
-        if ($count > 0) {
+        // Ensure we always close a transaction if we opened one, even if no charge IDs
+        // were applicable to update.
+        if (count($chargeIds) > 0) {
             $this->entityManager->flush();
```
bdsl, line 165: Are you sure we shouldn't use a separate transaction for each donation? Otherwise for any one donation will rollback all of them. Possibly we should flush each donation into the local db before we call `\MatchBot\Domain\SalesforceWriteProxyRepository::push` ?

app/dependencies.php
```diff
@@ -155,20 +155,29 @@ function (ContainerInterface $c): Auth\DonationPublicAuthMiddleware {
...
+            ));
+            $sendMiddleware->setLogger($logger);
+
+            $handleMiddleware = new HandleMessageMiddleware(new HandlersLocator(
+                [
+                    Messages\Donation::class => [$c->get(GiftAidResultHandler::class)],
+                    StripePayout::class => [$c->get(StripePayoutHandler::class)],
+                ],
+            ));
+            $handleMiddleware->setLogger($logger);
```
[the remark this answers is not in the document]
NoelLH: Agree!

src/Application/Messenger/Handler/StripePayoutHandler.php
NoelLH: @bdsl I think the last time I looked at it in the context of MatchBot, it was going to make certain use cases more complex because of the nesting and number of rollback points, make unit testing harder, or both. I can't remember exactly why it wasn't tenable though – maybe here it will become an option if we move to smaller txns

src/Application/Messenger/Handler/StripePayoutHandler.php
NoelLH: I'm kind of in 2 minds here @bdsl. I think there is an argument for persisting nothing if a payout contains mystery donations, but it's true that we shouldn't both push to SF and rollback. I'll try to do it with single donation txns for now.

bdsl (approved):
LGTM

PR #486, opened 2023-01-30
MAT-287: Report failed payouts to Stripe

bdsl:


src/Application/StripeChatterFactory.php
```diff
@@ -0,0 +1,27 @@
...
+
+use Symfony\Component\Notifier\Bridge\Slack\SlackTransport;
+use Symfony\Component\Notifier\Chatter;
+use Symfony\Component\Notifier\ChatterInterface;
+
+class StripeChatterFactory
+{
+    private readonly string $apiToken;
+
+    public function __construct(#[\SensitiveParameter] string $apiToken)
```
bdsl, line 13: SensitiveParameter doesn't work until PHP 8.2, but we may as well write it in advance.

src/Application/SlackChannelChatterFactory.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace MatchBot\Application;
+
+use Symfony\Component\Notifier\Bridge\Slack\SlackTransport;
+use Symfony\Component\Notifier\Chatter;
+use Symfony\Component\Notifier\ChatterInterface;
+
+class SlackChannelChatterFactory
```
bdsl, line 9: This class was partly motived by not knowing how to get PHP-DI to let us have two different services of the same type, but I don't think it's a bad class to have.

src/Domain/DonationRepository.php
```diff
@@ -151,6 +151,7 @@ public function buildFromApiRequest(DonationCreate $donationData): Donation
             $this->fundRepository->pullForCampaign($campaign);
 
             $this->getEntityManager()->flush();
+            $this->getEntityManager()->persist($campaign);
```
NoelLH, line 154: @bdsl I think this is
1. not related to the stated issue?
2. probably redundant because `SalesforceReadProxyRepository::pull()` has:
`$this->getEntityManager()->persist($proxy);`

NoelLH (commented):
One comment inline re a change I'm not sure is meant to be on this PR.

Everything for the new feature / error handling looks great!

src/Domain/DonationRepository.php
bdsl: Thanks, I must have left this change in by mistake when working on the other issue. Feel free to delete this line or I will when I get to my desk.

NoelLH (approved):

PR #489, opened 2023-02-14
BG2-2289: Set donations to re-push to salesforce to reclaim gift aid

bdsl:


bdsl:
Not needed - concluded that it is not possible to reclaim gift aid for these donations.

PR #490, opened 2023-02-17
ID-39: Add controller for deleting card from Stripe

bdsl:


app/routes.php
```diff
@@ -44,6 +45,11 @@
             ->add($ipMiddleware)
             ->add(RateLimitMiddleware::class);
 
+        $versionGroup->delete('/people/{personId:[a-z0-9-]{36}}/payment_methods/{payment_method_id:[a-zA-Z0-9_]{10,50}}', DeletePaymentMethod::class)
```
NoelLH, line 48: @bdsl Once > 1 endpoint has a matching path prefix and also an identical middleware stack, I think it's a good time to consider moving all the actions which align into a `group()`.

bdsl: Thanks @NoelLH , will remember that for future.

PR #496, opened 2023-02-23
Refactor: Specify allowed values of Donation Status property in type …

bdsl:
…system

### DIFF
--- a/src/Domain/Donation.php
+++ b/src/Domain/Donation.php
@@ -24,6 +24,17 @@ use Ramsey\Uuid\UuidInterface;
  */
 class Donation extends SalesforceWriteProxy
 {
+    public const STATUSES = [
+        'Pending',
+        'Collected',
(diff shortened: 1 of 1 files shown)


src/Domain/Donation.php
```diff
@@ -24,6 +24,17 @@
...
+        'Pending',
+        'Collected',
+        'Paid',
+        'Cancelled',
+        'Refunded',
+        'Failed',
+        'Chargedback',
+        'RefundingPending',
+        'PendingCancellation',
+    ];
```
bdsl, line 37: I've checked the live database to make sure this isn't an incomplete list:

```SELECT DISTINCT donationStatus from Donation; ```

returns

```
Collected	
Paid	
Cancelled	
Pending	
Refunded	
```

dee-jo (approved):

PR #498, opened 2023-03-07
MAT-292 – delay pushing donations to SF after a payout, for safer SQS interaction

NoelLH:


tests/Application/Messenger/Handler/StripePayoutHandlerTest.php
```diff
@@ -339,8 +343,7 @@ public function testSuccessfulUpdateWithBalanceTransactionsPagination(): void
             ->shouldBeCalledOnce();
         $donationRepoProphecy
             ->push(Argument::type(Donation::class), false)
-            ->willReturn(true)
-            ->shouldBeCalledTimes(2);
+            ->shouldNotBeCalled();
```
bdsl, line 346: I can't do a GH code suggestion here, but I think lines 344 to 346 are probably superfluous now. Prophecy automatically checks that no unexpected methods are called on the test double, so we don't need to tell that `push` should not be called - just like we don't need to tell it that `pushAllPending` and `abandonOldCancelled` should not be called.

But I admit I'm now a bit confused about what the intended use of `shouldNotBeCalled` in the Prophecy library is.

If we do have it maybe `->push(Argument::cetera())` is better than `->push(Argument::type(Donation::class), false)`. Basically applying De Morgans law, saying that push shouldn't be called *with any arguments*.

Not a blocker.

src/Application/Messenger/Handler/StripePayoutHandler.php
```diff
@@ -145,10 +147,10 @@ public function __invoke(StripePayout $payout): void
                 // method is called only when Stripe event `payout.paid` is received.
                 $donation->setDonationStatus('Paid');
 
+                $donation->setSalesforcePushStatus(SalesforceWriteProxy::PUSH_STATUS_PENDING_UPDATE);
```
bdsl, line 150: As we're adding a call to it it might be worth narrowing down the param type on `setSalesforcePushStatus` like so:

```php
    /**
     * @psalm-param self::PUSH_STATUS_* $salesforcePushStatus // <--------- new line
     * @param string $salesforcePushStatus
     */
    public function setSalesforcePushStatus(string $salesforcePushStatus): void
    ````

bdsl (approved):
LGTM, added some comments but none of them are blockers.

tests/Application/Messenger/Handler/StripePayoutHandlerTest.php
bdsl: I found https://stackoverflow.com/questions/35099612/only-this-and-nothing-more-with-prophecy/35254573#35254573 - seems like `shouldNotBeCalled` is useful either when there are no other expectations set on the test double, or when the method that should not be called is one that the test reader might expect *would* be called, and we want to keep it for extra clarity.

NoelLH: @bdsl 
> when the method that should not be called is one that the test reader might expect _would_ be called

This is why I kept it essentially. I think that especially when going back to this issue / diff it could help understand the change at a glance, as well as giving a more helpful message if a future change breaks the test.

PR #499, opened 2023-03-10
Write proxy method rename

NoelLH:
Tiny refinement following trying to explain this method today! The old name is confusing both in looking too much like it might relate to donation status when in a Donation-related context, and in saying "all" when it has a limit by default


bdsl (approved):
LGTM

PR #501, opened 2023-03-13
ID-32: Run tests integrated with MySQL in CircleCI

bdsl:
Sorry this is in matchbot and the Jira was `id` - it felt like it was most urgent to get this done for `matchbot`. If this is approved it should be pretty simple to copy the setup into `identity` and run similar checks there.

The other idea suggest in the ticket that isn't done yet but could be is checking that `migrations:diff` doesn't attempt to make any changes if run in Circle - i.e. that whatever changes it would make have already been committed into a migrations file.

### DIFF
--- a/integrationTests/CharityRepositoryTest.php
+++ b/integrationTests/CharityRepositoryTest.php
@@ -19,7 +19,7 @@ use PHPUnit\Framework\TestCase;
 
         $found = $sut->find("non-existant-id");
 
-        $this->assertSame(null, $found);
+        $this->assertNull($found);
     }
 
     public function testItFindsACharity(): void
--- a/integrationTests/CharityRepositoryTest.php
+++ b/integrationTests/CharityRepositoryTest.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace MatchBot\IntegrationTests;
+
+use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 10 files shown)


phpunit-integration.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0"?>
```
bdsl, line 1: most of this file is a direct copy of phpunit.xml . The biggest difference is that in this file we do not have anything about `MYSQL_HOST`, so the code can use the real host as configured in the environment.

At some point we might want to set up two separate hosts, one for manual testing in local and a different one for these automated tests.

integrationTests/CharityRepositoryTest.php
```diff
@@ -0,0 +1,50 @@
...
+ * correctly and migrations have been run.
+ */class CharityRepositoryTest extends IntegrationTest
+{
+    public function testItDoesNotFindCharityThatDoesNotExist(): void
+    {
+        $sut = $this->getService(CharityRepository::class);
+
+        $found = $sut->find("non-existant-id");
+
+        $this->assertSame(null, $found);
```
bdsl, line 22:
```suggestion
        $this->assertNull($found);
```

dee-jo (approved):

PR #502, opened 2023-03-14
MAT-291 – also set `salesforcePushStatus` on previously-patched refunded donations

NoelLH:


bdsl (approved):
LGTM - I hadn't realised in https://github.com/thebiggive/matchbot/pull/497 that this data needed to be updated in SF as well as in the matchbot db

PR #505, opened 2023-03-22
BG2-2296: Refactor Donation Status from string to Enum

bdsl:
AFAIK this will be our first time using PHP Enums ( https://www.php.net/manual/en/language.types.enumerations.php )

By making the status an Enum instead of a string in PHP code we can be absolutely sure that the system can never deal with any status other than those we've explicitly listed as the cases of the enums, and that those are the only cases we need to document. If we ever decide to invent a new status in future it will make it very obvious that needs to be documented in the same way, as it won't be usable unless its listed next to the new statuses.

It will also makes the documentation very easy to find when reading any line of code that deals with donation statuses.

### DIFF
--- a/src/Domain/Donation.php
+++ b/src/Domain/Donation.php
@@ -153,8 +153,6 @@ class Donation extends SalesforceWriteProxy
      */
     protected ?bool $charityComms = null;
 
-    //     #[Column(type: Types::STRING, enumType: DonationStatus::class)]
-
     /**
      * @ORM\Column(type="boolean", nullable=true)
      * @var bool


src/Domain/Donation.php
```diff
@@ -161,19 +143,18 @@ class Donation extends SalesforceWriteProxy
...
+    protected DonationStatus $donationStatus = DonationStatus::NotSet;
 
     /**
      * @ORM\Column(type="boolean", nullable=true)
      * @var bool    Whether the donor opted to receive email from the charity running the campaign
      */
     protected ?bool $charityComms = null;
 
+    //     #[Column(type: Types::STRING, enumType: DonationStatus::class)]
+
```
bdsl, line 157:
```suggestion
```

src/Domain/DonationStatus.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    public const SUCCESS_STATUSES = [self::Collected, self::Paid];
+
+    public const NEW_STATUSES = [self::NotSet, self::Pending];
+
+    /**
+     * @link https://thebiggive.slack.com/archives/GGQRV08BZ/p1576070168066200?thread_ts=1575655432.161800&cid=GGQRV08BZ
+     */
+    public const REVERSED_STATUSES = [self::Refunded, self::Failed, self::Chargedback];
+
```
bdsl, line 21: Comparing the below list of statues to the list of possible statuses that were on the Donation class (copied here for ref):

```

        'Pending',
        'Collected',
        'Paid',
        'Cancelled',
        'Refunded',
        'Failed',
        'Chargedback',
        'RefundingPending',
        'PendingCancellation',
```

The following ones were not in the Donation class list: *NotSet*, *Reserved* but they are below because they are mentioned in code so they have to be here. We can delete them if/when we delete the code that references them. SA checks don't allow the references to exist without the cases existing below.

bdsl: Did an SQL query to check all the donation statuses we have in the database - anything that doesn't exist in the database and we don't have code to create in future could be deleted from the Enum. DB results:

`SELECT DISTINCT donationStatus FROM `Donation`; `
```
"Cancelled"
"Pending"
"Collected"
"Refunded"
"Paid"
"Failed"
"Chargedback"
```

src/Application/Actions/Donations/Update.php
```diff
@@ -268,7 +269,7 @@ private function addData(Donation $donation, HttpModels\Donation $donationData,
 
     private function cancel(Donation $donation, Response $response, array $args): Response
     {
-        if ($donation->getDonationStatus() === 'Cancelled') {
+        if ($donation->getDonationStatus() === DonationStatus::Cancelled) {
```
dee-jo, line 272: We might need to do a global search the entire code base to check all string statuses are replaced

dee-jo (approved):
LGTM

src/Application/Actions/Donations/Update.php
bdsl: Thanks, used regex search for `['"]Reserved['"]`, `['"]Paid['"]` etc. None appear where they shouldn't, they've all been replaced with references to the enum case.

PR #506, opened 2023-03-23
Bg2-2296 document donation statuses ii

bdsl:
Putting the docs here means (at least in PHPStorm) that we can hover over any line of code that mentions a status to see the documentation:

image

### DIFF
--- a/src/Domain/DonationStatus.php
+++ b/src/Domain/DonationStatus.php
@@ -20,6 +20,12 @@ enum DonationStatus: string
     public const REVERSED_STATUSES = [self::Refunded, self::Failed, self::Chargedback];
 
     case Cancelled = 'Cancelled';
+
+    /**
+     * Exists in database entries from 2020 only. There is now no code that can set a Chargedback status.
+     * We may want to see if eventually these can be archived and moved out of the live DB, and then this case can be
+     * removed.
+     */
     case Chargedback = 'Chargedback';
     case Collected = 'Collected';
     case Failed = 'Failed';
--- a/src/Domain/DonationStatus.php
+++ b/src/Domain/DonationStatus.php
@@ -26,6 +26,5 @@ enum DonationStatus: string
     case NotSet = 'NotSet';
     case Paid = 'Paid';
     case Pending = 'Pending';
-    case PendingCancellation = 'PendingCancellation';
     case Refunded = 'Refunded';
 }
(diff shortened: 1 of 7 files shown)


src/Domain/DonationStatus.php
```diff
@@ -5,29 +5,102 @@
 /**
  * @link https://docs.google.com/document/d/11ukX2jOxConiVT3BhzbUKzLfSybG8eie7MX0b0kG89U/edit?usp=sharing
  * @todo Consider vs. Stripe options
```
bdsl, line 7: I can't actually access the link - I just moved the comment over from Donation. Should we delete it here?

Not sure what the todo consider means - it's another comment that I moved from the Donation class.

src/Domain/DonationStatus.php
```diff
@@ -5,29 +5,102 @@
...
- *
- * @todo before merging BG2-2296 - check that the prod DB doesn't have any statuses recorded not listed below.
```
bdsl, line 9: I did do this before merging the previous PR, forgot to delete the todo earlier.

src/Domain/DonationStatus.php
NoelLH: @bdsl Yeah, I think we can delete the link as nobody in the current team has access – whatever it was didn't get put in a shared Drive.

I think the note was to think about the pro's and con's of migrating new donations to use Stripe's own status strings verbatim. Given we have use cases where we simplify multiple Stripe statuses to one internal one (e.g. disputes lost are _Refunded_), my view now is that this is unlikely to be worth it. It would buy us slightly more clarity when comparing things to Stripe but with several downsides I can think of off-hand:
* more complexity to manage
* likelihood of regressions while we switch over
* less comparability to historic data from other PSPs
* greater long-term lock-in to using Stripe exclusively

tests/Domain/DonationTest.php
```diff
@@ -27,6 +28,22 @@ public function testBasicsAsExpectedOnInstantion(): void
...
 
+    public function testPendingDonationDoesNotHavePostCreateUpdates(): void
+    {
+        $donation = new Donation();
+        $donation->setDonationStatus(DonationStatus::Pending);
+
+        $this->assertFalse($donation->hasPostCreateUpdates());
+    }
+
+    public function testPendingDonationHasPostCreateUpdates(): void
```
NoelLH, line 39:
```suggestion
    public function testPaidDonationHasPostCreateUpdates(): void
```

NoelLH (commented):
Nice! Suggested a test rename, otherwise good to go.

PR #508, opened 2023-03-24
Tidy up captcha dependencies; remove old donation create endpoint

NoelLH:
And update tests so that all scenarios covered use the endpoint and middlewares still in live use, not the old one.


bdsl (dismissed):
LGTM, nice simplifications.

NoelLH:
I think we need to clear the backlog of everything awaiting fixes or QA, and discuss the move to `/v2`, before we actually merge.

composer.json
```diff
@@ -26,7 +26,6 @@
         "doctrine/migrations": "^3.7",
         "doctrine/orm": "^2.14.0",
         "firebase/php-jwt": "^6.3",
-        "google/recaptcha": "^1.2",
```
bdsl, line 29: probably worth removing from `composer.lock` as well. Could just run `composer remove google/recaptcha` or we could just do `composer update` since we don't have a schedule to do updates and it's worth doing them all from time to time anyway.

integrationTests/CreateDonationTest.php
```diff
@@ -49,23 +50,27 @@ public function testCannotCreateDonationWithNegativeTip(): void
...
+    public function testItReturns400OnRequestToDonateNullAmount(): void
     {
         $campaignId = $this->setupNewCampaign();
 
         $response = $this->getApp()->handle(
             new ServerRequest(
-                'POST',
-                '/v1/donations',
+                method: 'POST',
+                uri: '/v1/people/12345678-1234-1234-1234-1234567890ab/donations',
```
bdsl, line 60: this string is repeated *a lot* of times, and it has to be the same every time. Can we replace it with a constant instead of adding new copies of it?

tests/TestData/Identity.php
```diff
@@ -0,0 +1,48 @@
...
+        //      "complete":false,
+        //      "psp_id":"cus_aaaaaaaaaaaa11"
+        //  }
+        // }
+        $dummyPersonAuthTokenValidUntil2050 = 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJodHRwczovL3VuaXQtdGVz' .
+            'dC1mYWtlLWlkLXN1Yi50aGViaWdnaXZldGVzdC5vcmcudWsiLCJpYXQiOjE2NjM0MzYxNTQsImV4cCI6MjUyNDYwODAwMCwic3ViIjp7' .
+            'InBlcnNvbl9pZCI6IjEyMzQ1Njc4LTEyMzQtMTIzNC0xMjM0LTEyMzQ1Njc4OTBhYiIsImNvbXBsZXRlIjpmYWxzZSwicHNwX2lkIjoi' .
+            'Y3VzX2FhYWFhYWFhYWFhYTExIn19.KdeGTDkkWCjI4-Kayay0LKn9TXziPXCUxxTPIZgGxxE';
+
+        return $dummyPersonAuthTokenValidUntil2050;
```
bdsl, line 46: could we do the encoding at the time we run the tests, so that it will be possible to see what's going on here and edit without having to read the comments? Or at least add a comment about how to generate the encoded data.

I tried decoding it but it didn't work, not sure why.  There was a chunk of unreadable stuff at the end of the decoded string, and this makes test error with `Slim\Exception\HttpUnauthorizedException: Unauthorised`

```
        return base64_encode('{"typ":"JWT","alg":"HS256"}{"iss":"https://unit-test-fake-id-sub.thebiggivetest.org.uk","iat":1663436154,"exp":2524608000,"sub":{"person_id":"12345678-1234-1234-1234-1234567890ab","complete":false,"psp_id":"cus_aaaaaaaaaaaa11"}})׆L9$X(�লk-
```

bdsl: And if we can't make this do the encoding at runtime then for clarity I'd suggest having two `public const string`s instead of public static functions.

NoelLH: @bdsl If you're trying to replace a whole token then you need a header, body and signature. https://developer.atlassian.com/cloud/bitbucket/understanding-jwt-for-apps/#structure-of-a-jwt-token

I don't think there is a particularly easy way to set up tests to make Identity tokens without giving this app more 'real' runtime secrets than it needs, which is why I went for a long lived one (some time ago, so I also don't really remember)

composer.json
NoelLH: Yep, thought I did the former before – think it got lost in a later merge resolve

integrationTests/CreateDonationTest.php
bdsl: Thanks - is there a reason for preferring a function rather than a constant though? IMHO constants are generally preferable because they're simpler, there's no need to look to see how it's implemented when you find a usage.

tests/TestData/Identity.php
bdsl: I don't follow - base64 is a non-cryptographic and reversible encoding that doesn't require any secret to run. So if we need a base64 encoded value why wouldn't we be able to generate it at runtime. Can chat tomorrow if you have time.

integrationTests/CreateDonationTest.php
NoelLH: No no strong preference either way here really, I just went with what was closest to the most similar helper

bdsl: OK not a blocker.

tests/TestData/Identity.php
NoelLH: @bdsl Per the link above, a JWT has 3 separately base64'd parts, the last of which is a signature which requires knowledge of a secret.

bdsl: I still don't get it I'm afraid. Surely if we can base64 encode something then anyone who can read it can decode it, since there's no secret involved in base64. And if they can decode it why not keep it decoded in the first place? A secret is needed to compute the HMAC but that's a separate step to the base64 encoding - the key needs to be secret, the HMAC can't be secret.

There's also a separate question about why we would need to use a real secret at all in tests - we should be able to use an alternate fake secret that is only for test purposes and doesn't actually need to be kept secret.

But I don't think the question about DX of these tests files is important enough to hold up merging this further now.

bdsl: ah I'm a bit less confused now - I was trying to decode it as a whole which is why it doesn't work, it has to be decoded one part at a time in base64, or as a whole by pasting into a tool like https://jwt.io/ 

But it looks like the 'secret' we're using here is not a real secret at all and there's no reason for it to be- it's already just "unitTestJWTSecret" and it's already in the repo at https://github.com/thebiggive/matchbot/blob/dea3e98984e3d12a43560525e2a13b08655e431e/phpunit.xml#L41

tests/TestData/Identity.php
```diff
@@ -0,0 +1,53 @@
...
+    }
```
bdsl, line 52: So this works:

```suggestion
        return JWT::encode(
            [
                "iss" => "https://unit-test-fake-id-sub.thebiggivetest.org.uk",
                "iat" => 1663436154,
                "exp" => 2524608000,
                "sub" => [
                    "person_id" => "12345678-1234-1234-1234-1234567890ab",
                    "complete" => false,
                    "psp_id" => "cus_aaaaaaaaaaaa11",
                ]
            ], getenv('JWT_ID_SECRET'), 'HS256'
        );
    }
    }
```

bdsl: Feel free to merge as-is, I'll probably make a separate PR after to do the change above and similar for the getTestIdentityTokenComplete function.

tests/TestData/Identity.php
NoelLH: Ah, fair enough! Yep, agree it would be better DX if this was done in code but will go ahead and merge what we have so far for today.

PR #512, opened 2023-03-27
MAT-287 – clearer failed payout notifications

NoelLH:
* Make it explicit which environment a notification is coming from
* Move implicit Slack configuration requirement nearer to other settings

### DIFF
--- a/src/Application/Actions/Hooks/StripePayoutUpdate.php
+++ b/src/Application/Actions/Hooks/StripePayoutUpdate.php
@@ -26,8 +26,8 @@ use Symfony\Component\Notifier\Message\ChatMessage;
  */
 class StripePayoutUpdate extends Stripe
 {
-
     private SlackChannelChatterFactory $chatterFactory;
+    protected string $slackStripeChannel;
 
     public function __construct(
         ContainerInterface $container,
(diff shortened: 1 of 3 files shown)


src/Application/Actions/Hooks/StripePayoutUpdate.php
```diff
@@ -26,8 +26,8 @@
  */
 class StripePayoutUpdate extends Stripe
 {
-
     private SlackChannelChatterFactory $chatterFactory;
+    protected string $slackStripeChannel;
```
bdsl, line 30: This class isn't designed to be inherited, is it?
```suggestion
    private string $slackStripeChannel;
```

src/Application/Actions/Hooks/StripePayoutUpdate.php
```diff
@@ -37,6 +37,18 @@ public function __construct(
...
+         * @var array{
+         *    notifier: array{
+         *      slack: array{
+         *        stripe_channel: string
+         *      }
+         *    }
+         *  } $settings
+         */
+        $settings = $container->get('settings');
+        $this->slackStripeChannel = $settings['notifier']['slack']['stripe_channel'];
```
bdsl, line 51: would it be worth injecting the slackStripeChannel into the constructor instead of pulling it from the container here?

bdsl (commented):
Added comments inline

src/Application/Actions/Hooks/StripePayoutUpdate.php
NoelLH: @bdsl Ideally yes, but building custom DI support for specific settings is a bigger undertaking.

What I'd like to do is refactor to use a proper settings class as introduced in newer Slim skeleton versions, but I don't have time to do this today. For example, ID is using this at https://github.com/thebiggive/identity/blob/develop/app/settings.php#L13

bdsl (approved):

PR #515, opened 2023-04-03
ID-38: Add UpdatePaymentMethod controller

bdsl:
Based on a copy of the DeletePaymentMethod controller

### DIFF
--- a/src/Application/Actions/UpdatePaymentMethod.php
+++ b/src/Application/Actions/UpdatePaymentMethod.php
@@ -52,14 +52,24 @@ class UpdatePaymentMethod extends Action
             );
         }
 
+        $body = (string)$request->getBody();
+
+        // We don't need to know the details inside the billing details - we are just a thin layer between the front end
+        // and Stripe here.
+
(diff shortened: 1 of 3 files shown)


src/Application/Actions/UpdatePaymentMethod.php
```diff
@@ -0,0 +1,77 @@
...
+
+        /** @var array<array{id: string}> $allPaymentMethodsOfCustomer */
+        $allPaymentMethodsOfCustomer = $this->stripeClient->customers->allPaymentMethods(
+            $customerId,
+        )->toArray()['data'];
+
+        $allCustomersPaymentMethodIds = array_map(
+            static fn(array $pm): string => $pm['id'],
+            $allPaymentMethodsOfCustomer
+        );
```
bdsl, line 40: Quite a bit of this is duplicating what the DeletePaymentMethod controller does - not sure where to move it to to avoid duplication. Maybe we can live with the duplication for now, and then think about adding a service layer or something to the architecture before making a third duplicate.

PR #517, opened 2023-04-11
MAT-295, MAT-258 – manage old payment methods; tidy up existing Stripe things

NoelLH:
Specifically, auxiliary changes are:
* use auto-pagination to reduce complexity in our code
* drop _"The "_ from statement descriptors' prefix


NoelLH:
A couple of the new assertions to make psalm happy are also making unit tests fail.

More importantly, the actual logic doesn't work with the real Stripe behaviour because `payment_method` turns out not to be an allowed filter for charges' `all()`. We might need to do something more convoluted like check all charges for a customer and see whether any match the found payment intents.

src/Application/Actions/Donations/Update.php
```diff
@@ -215,6 +215,9 @@ private function addData(Donation $donation, HttpModels\Donation $donationData,
         );
 
         if ($donation->getPsp() === 'stripe') {
+            $transactionId = $donation->getTransactionId();
+            \assert(is_string($transactionId));
```
bdsl, line 219: It's not immediately obvious to me how we know at this point that the donation has a transaction ID, but its also not clear if there's a reason for this class to change in this PR.

Since it looks like this pattern is repeated a few times it might be worth adding a method to Donation that returns the transaction ID as a string and throws if it's null, with docs and a name to make it clear it should only be called when we know there's a transaction ID.

src/Application/Commands/DeleteStalePaymentDetails.php
```diff
@@ -0,0 +1,85 @@
...
+        $oneDayAgo = $this->initDate
+            ->sub(new \DateInterval('P1D'))
+            ->getTimestamp();
+
+        // Get all Stripe customers without the password set metadata field, who are over 24h old.
+        // The metadata restriction lets us better leave people with passwords since April 2023,
+        // so this `query` covers conditions (1) and (2) from the class doc block.
+        $customers = $this->stripeClient->customers->search([
+            'query' => "created<$oneDayAgo and metadata['hasPasswordSince']:null",
+        ]);
```
bdsl, line 55: can / should we limit this to a manageable batch size instead of getting all? Or does it do that anyway?

src/Application/Commands/DeleteStalePaymentDetails.php
```diff
@@ -0,0 +1,85 @@
...
+use Stripe\StripeClient;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Deletes payment methods from Stripe if:
+ *   1. they (or more specifically, their Customer) are 24h+ old; and
+ *   2. they do not belong to a Person with a password set after mid-April 2023; and
+ *   3. they are a credit/debit card based method; and
+ *   4. they have not been used for any complete donations.
```
bdsl, line 19: what's the reason for condition 4?

Also is there any way a person account could be more than 24 hours old but still have a valid JWT if it's been renewed a few times? Not sure if it's possible to renew the JWT without a password.

src/Application/Actions/Donations/Update.php
bdsl: If we're going to start using `assert` outside of test code it would be good to know how it's configured in production. For options see https://www.php.net/manual/en/ini.core.php#ini.zend.assertions and https://www.php.net/manual/en/info.configuration.php#ini.assert.exception

Or we could use a library like https://github.com/beberlei/assert or instead https://github.com/webmozarts/assert . These both come with annotations that Psalm understands on the assertion methods, and will always throw if the assertion is false.

NoelLH: @bdsl The reason for changing this class is that the continued use of `StripeClientInterface` is IMO a trap which makes writing future psalm-failing code and then having to fix it again more likely.

src/Application/Commands/DeleteStalePaymentDetails.php
NoelLH: It defaults to 10, there are no infinite pages. https://stripe.com/docs/api/charges/search

I guess we might as well make it 100 like the others.

src/Application/Actions/Donations/Update.php
bdsl: Right. We're not the first to have this issue - see https://github.com/stripe/stripe-php/issues/970 from three years ago. Seems like `StripeClientInterface` may be of dubious value. I think if we just change the type from StripeClientInterface to StripeClient and that makes the psalm error appear it's fine to add it to the baseline and leave the code `$this->stripeClient->paymentIntents->confirm($donation->getTransactionId());` as it was, since it's not really a new error it's just an old one in a different form.

src/Application/Commands/DeleteStalePaymentDetails.php
bdsl: Right. I think we should be deleting even if they were used for completed (or paid) donations, but it makes sense to test it out first.

src/Application/Actions/Donations/Update.php
NoelLH: @bdsl I was mainly following your lead in introducing `\assert` to fix errors – it's already in Production or en route via `UpdatePaymentMethod` & co.

I'm 99% sure our current config changes no defaults, so asserts will be evaluated and give warnings but not terminate. This seems fine to me as a starting point: it means that introducing stricter code won't break things unexpectedly but that we will have a clear way to monitor the AWS dashboards and search logs for unexpected assert failures.

NoelLH: > Right. We're not the first to have this issue - see [stripe/stripe-php#970](https://github.com/stripe/stripe-php/issues/970) from three years ago. Seems like `StripeClientInterface` may be of dubious value. I think if we just change the type from StripeClientInterface to StripeClient and that makes the psalm error appear it's fine to add it to the baseline and leave the code `$this->stripeClient->paymentIntents->confirm($donation->getTransactionId());` as it was, since it's not really a new error it's just an old one in a different form.

Think we overlapped commenting here. As the pre-existing class is now green with the stricter approach, I suggest we keep the changes.

NoelLH:
I've realised that we don't really need to check the method ID _specifically_. If we narrow the charges search on both the card fingerprint and the customer ID, it's sufficient to be sure that a match is the same card, without the extra overhead of manually mapping two lists to check for matches.

app/dependencies.php
```diff
@@ -304,17 +303,15 @@ function (ContainerInterface $c): Auth\DonationPublicAuthMiddleware {
             return new Serializer($normalizers, $encoders);
         },
 
+        // StripeClientInterface doesn't have enough properties documented to keep
+        // psalm happy.
```
bdsl, line 307: I don't think this is just as Psalm issue, interfaces can't really have properties. Yes they could be documented on the interface but there's no system AFAIK to make implementers provide them.

```suggestion
        // Previously we provided StripeClientInterface but we need to use the properties declared by
        // StripeClient. @see https://github.com/stripe/stripe-php/issues/970
```

matchbot-cli.php
```diff
@@ -48,6 +50,11 @@
         $psr11App->get(LoggerInterface::class),
         [$messengerReceiverKey],
     ),
+    new DeleteStalePaymentDetails(
+        new \DateTimeImmutable('now'),
```
bdsl, line 54: Liking the date DI

src/Application/Commands/DeleteStalePaymentDetails.php
bdsl: On point 2 - we want need to keep payment methods around if they belong to a person with a password set in the past, don't we? Not only keep them for people who set passwords after mid-april 2023. I know it's harder from matchbot to detect which accounts had passwords set in the past.

src/Application/Commands/DeleteStalePaymentDetails.php
```diff
@@ -0,0 +1,100 @@
...
+            ->sub(new \DateInterval('P1D'))
+            ->getTimestamp();
+
+        // Get all Stripe customers without the password set metadata field, who are over 24h old.
+        // The metadata restriction lets us better leave people with passwords since April 2023,
+        // so this `query` covers conditions (1) and (2) from the class doc block.
+        $customers = $this->stripeClient->customers->search([
+            'query' => "created<$oneDayAgo and metadata['hasPasswordSince']:null",
+            'limit' => $stripePageSize,
+        ]);
```
bdsl, line 51: is this not going to find the same donors every time the command is run? How will we ever move beyond the first 100 donors?

src/Application/Messenger/Handler/StripePayoutHandler.php
```diff
@@ -52,53 +52,37 @@ public function __invoke(StripePayout $payout): void
...
-            if ($hasMore && $lastBalanceTransactionId !== null) {
-                $attributes['starting_after'] = $lastBalanceTransactionId;
-                $this->logger->debug(sprintf(
-                    'Stripe Balance Transaction for Payout ID %s will next use starting_after: %s',
-                    $payoutId,
-                    $lastBalanceTransactionId,
-                ));
-            }
+        // Auto page, iterating in reverse chronological order. https://stripe.com/docs/api/pagination/auto?lang=php
+        foreach ($balanceTransactions->autoPagingIterator() as $balanceTransaction) {
```
bdsl, line 82: nice simplification

tests/Application/StripeFormattingTrait.php
```diff
@@ -0,0 +1,44 @@
...
+        $itemsArray = json_decode($json, false);
+        /** @var StripeObject[] $itemData */
+        $itemData = $itemsArray->data;
+
+        $iterableCollection = $this->prophesize(Collection::class);
+        $iterableCollection->autoPagingIterator()->willReturn(new ArrayIterator($itemData));
+        $iterableCollection->count()->willReturn(count($itemData));
+
+        return $iterableCollection->reveal();
+    }
```
bdsl, line 27: Slight simplifications:

```suggestion
    private function buildAutoIterableCollection(string $json): Collection
    {
        $itemData = json_decode($json, false, 512, JSON_THROW_ON_ERROR)->data;

        $iterableCollection = $this->prophesize(Collection::class);
        $iterableCollection->autoPagingIterator()->willReturn($itemData);
        $iterableCollection->count()->willReturn(count($itemData));

        return $iterableCollection->reveal();
    }
```

- Private functions in a trait can still be accessed by classes that use the trait
- Returning the ArrayIterator doesn't seem to add anything over just iterating the array directly
- This function always returns a Collection, never an ObjectProphecy.

Similar applies to functions below.

tests/Application/Commands/DeleteStalePaymentDetailsTest.php
```diff
@@ -0,0 +1,178 @@
...
+use Stripe\StripeClient;
+use Symfony\Component\Console\Tester\CommandTester;
+use Symfony\Component\Lock\LockFactory;
+
+class DeleteStalePaymentDetailsTest extends TestCase
+{
+    use DonationTestDataTrait;
+    use StripeFormattingTrait;
+
+    public function testDeletingOneStaleCard(): void
```
bdsl, line 26: These are long test functions, I think tricky to read through and follow. Is it possible to split them up, maybe just with `// arrange` `// act` `// assert` or `// given` `// when` `// then` comments? Or if each function is testing more than one thing could it be split into multiple test functions?

bdsl: Right I see each test is just testing one thing - there's just quite a bit of code. May be worth extracting some into functions. Can't quite do AAA in the standard order because mocks always switch around the last two A's.

tests/Application/Commands/DeleteStalePaymentDetailsTest.php
```diff
@@ -0,0 +1,178 @@
...
+            'matchbot:delete-stale-payment-details complete!',
+        ];
+        $this->assertEquals(implode("\n", $expectedOutputLines) . "\n", $commandTester->getDisplay());
+
+        $this->assertEquals(0, $commandTester->getStatusCode());
+    }
+
+    public function testNotDeletingOneUsedCard(): void
+    {
+        $initDate = new \DateTimeImmutable('now');
```
bdsl, line 99: Maybe better to use hard coded dates in tests to make it clear that it doesn't matter what time we run the test at? This still passes:

```suggestion
        $initDate = new \DateTimeImmutable('2023-01-01 00:00:00');
```

Same applies to second test.

tests/Application/Commands/DeleteStalePaymentDetailsTest.php
```diff
@@ -0,0 +1,178 @@
...
+        ])
+            ->shouldBeCalledOnce()
+            ->willReturn($this->buildCollectionFromSingleObjectFixture(
+                $this->getStripeHookMock('ApiResponse/pm'),
+            ));
+
+        // One PM should be detached i.e. soft deleted.
+        $stripePaymentMethodsProphecy->detach($testPaymentMethodId)
+            ->shouldBeCalledOnce()
+            ->willReturn($this->getStripeHookMock('ApiResponse/pm'));
```
bdsl, line 59: The code under test doesn't use the return value - any reason for providing it?
```suggestion
        $stripePaymentMethodsProphecy->detach($testPaymentMethodId)->shouldBeCalledOnce();
```

tests/Application/Commands/DeleteStalePaymentDetailsTest.php
```diff
@@ -0,0 +1,178 @@
...
+            $initDate,
+        ));
+        $commandTester->execute([]);
+
+        $expectedOutputLines = [
+            'matchbot:delete-stale-payment-details starting!',
+            'Deleted 0 payment methods from Stripe, having checked 1 customers',
+            'matchbot:delete-stale-payment-details complete!',
+        ];
+        $this->assertEquals(implode("\n", $expectedOutputLines) . "\n", $commandTester->getDisplay());
```
bdsl, line 161:
```suggestion
        $this->assertSame(<<<EXPECTED
            matchbot:delete-stale-payment-details starting!
            Deleted 1 payment methods from Stripe, having checked 1 customers
            matchbot:delete-stale-payment-details complete!

            EXPECTED,
            $commandTester->getDisplay()
        );
```

tests/Application/Commands/DeleteStalePaymentDetailsTest.php
```diff
@@ -0,0 +1,178 @@
...
+
+        $this->assertEquals(0, $commandTester->getStatusCode());
+    }
+
+    private function getCommand(
+        ObjectProphecy $stripeClientProphecy,
+        \DateTimeImmutable $initDate,
+    ): DeleteStalePaymentDetails {
+        $stripeClient = $stripeClientProphecy->reveal();
+        \assert($stripeClient instanceof StripeClient);
```
bdsl, line 171: You can replace the assert with a docblock if you like:

```suggestion
    /**
     * @param ObjectProphecy<StripeClient> $stripeClientProphecy
     */
    private function getCommand(
        ObjectProphecy $stripeClientProphecy,
        \DateTimeImmutable $initDate,
    ): DeleteStalePaymentDetails {
        $stripeClient = $stripeClientProphecy->reveal();
```

bdsl:
Added comments inline, otherwise I think it all LGTM.

tests/Application/Commands/DeleteStalePaymentDetailsTest.php
NoelLH: My thinking was to match the return behaviour of the real library by default. I still think there's maybe some small value in that as protection against missed regressions from future changes in the code under test, but don't feel strongly about it.

src/Application/Commands/DeleteStalePaymentDetails.php
NoelLH: @bdsl I think those methods are essential to keep only when (4) is also true, so I believe we're pretty much covered with these conditions. This is what I was getting at re. only cards which have been used exclusively for _failed_ donations being at risk of deletion when we wouldn't ideally have it – which IMO is not a big deal.

src/Application/Commands/DeleteStalePaymentDetails.php
NoelLH: @bdsl See the next line and the auto paging Stripe docs link in my other comment

src/Application/Actions/Donations/Update.php
NoelLH: Used a getter instead

tests/Application/Commands/DeleteStalePaymentDetailsTest.php
bdsl: I feel like its better to make the test only as complex as required to document the behaviour of our code under test. Any changes in what the real library does or doesn't return can't break our code while our code doesn't use that return value.

src/Application/Commands/DeleteStalePaymentDetails.php
bdsl: Right, I agree that it's not to be a problem to detach payment methods that were never used for a successful donation. But I do think eventually we should be detaching all payment methods from all old passwordless accounts. (and also probably removing them from id DB as afaik they have no purpose there, but that's another question)

So yes these four conditions together I think are good for now, we just need to know that we can't get rid of #4 without also changing #2.

matchbot-cli.php
bdsl: (on the other hand in the unit tests you've repeated the same code `new \DateTimeImmutable('now');` which means we aren't actually getting any benefit from DI-ing it. As mentioned below I think the tests could be quite a bit simpler and less like the production code by using a hard-coded date.

src/Application/Commands/DeleteStalePaymentDetails.php
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    protected function configure(): void
+    {
+        $this->setDescription('Deletes unused and likely-non-customer-deletable payment methods from Stripe');
+    }
+
+    protected function doExecute(InputInterface $input, OutputInterface $output): int
+    {
+        $stripePageSize = 100; // Maximum allowed. Iterators page through automatically.
```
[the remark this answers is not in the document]
bdsl: I see this is done, thanks.

src/Application/Commands/DeleteStalePaymentDetails.php
bdsl: I don't follow. The state of the auto paging iterator is held in PHP memory isn't it? But we need persistent state somewhere to tell us which people have or haven't already had their stale methods detached so we can work on detaching methods from new people every time. Maybe we can discuss tomorrow.

tests/Application/Commands/DeleteStalePaymentDetailsTest.php
bdsl: oops sorry looks like I changed the 0 to 1 by accident here. Not sure how that happened.

src/Application/Commands/DeleteStalePaymentDetails.php
NoelLH: @bdsl Right, with you. It's a bit wasteful to keep calling out for the same empty list of attached payment methods for old customers already processed – but if the total list was short enough to complete in an hour and we did this for a few days max. I don't think it's out of the question.

So 2 options I can think of so far:
1. release this as an MVP, then once we've tested it and seen live customers tidied up, add a constraint so that we look at only customers who are also _newer than_ (e.g.) 1 day for potential payment method detachment; or
2. set an additional metadata field on a customer to indicate when we last detached stale payment methods, and update this query to also only look at customers where that is null

NoelLH: Yeah, unfortunately the query-able fields are very few. https://stripe.com/docs/search#query-fields-for-customers

tests/Application/Commands/DeleteStalePaymentDetailsTest.php
NoelLH: As discussed, work to address this is started for inclusion on another PR shortly

bdsl (approved):
Approving as discussed on slack, will make further PR to try and simplify tests.

PR #520, opened 2023-04-18
Store old passworded status in stripe

bdsl:


integrationTests/MarkOldPasswordedAccountsInStripeTest.php
```diff
@@ -0,0 +1,68 @@
...
+
+        $lockFactoryProphecy = $this->prophesize(\Symfony\Component\Lock\LockFactory::class);
+        $lockFactoryProphecy->createLock('matchbot:mark-old-passworded-accounts-in-stripe', \Prophecy\Argument::cetera())
+            ->willReturn(new \Symfony\Component\Lock\NoLock());
+
+        $sut->setLockFactory($lockFactoryProphecy->reveal());
+
+        $tester = new CommandTester($sut);
+
+        $stripeCustomersProphecy->update('cus_NHDQHoQP8Abgbp', ['metadata' => ['hasPasswordSince' => '2023-02-01 12:01:26']])
```
bdsl, line 53: Noel suggested storing the customer email address in stripe along with the password.

src/Application/Commands/MarkOldPasswordedAccountsInStripe.php
```diff
@@ -0,0 +1,42 @@
...
+    ) {
+        $this->logger = $logger;
+        $this->setLogger($logger);
+        parent::__construct();
+    }
+
+    protected function doExecute(InputInterface $input, OutputInterface $output): int
+    {
+        $personRows = $this->identityDBConnection->fetchAllAssociative(
+            'SELECT BIN_TO_UUID(id) as uuid, updated_at, stripe_customer_id from Person where Password IS NOT NULL AND created_at < "2023-03-01" LIMIT 1'
```
bdsl, line 32: needs an 'order' by' so we start with the oldest customers. Created at should be '=>'  the cutoff, not '<'

NoelLH, line 32: As discussed, should make sure the date range is inclusive (i.e. do the same second an additional time to be safe) and the order is explicit, when we use Redis-stored datetimes

dee-jo (approved):
LGTM

PR #527, opened 2023-04-20
MAT-297 – add closed EntityManager retry support for `refresh()`

NoelLH:


NoelLH:
There are 2 remaining Psalm errors where I've specified more constrained types than the lib we depend upon, to fix other errors. Not immediately sure what we can do about that. @bdsl any ideas?

bdsl:
> There are 2 remaining Psalm errors where I've specified more constrained types than the lib we depend upon, to fix other errors. Not immediately sure what we can do about that. @bdsl any ideas?

You could suppress the error, or you could leave the param type as it is in the parent class and then in the body of the method throw if the parameter passed isn't an instance of `Model`. But Psalm is raising the question of whether you know or whether its safe to assume that this method will never be called with an argument that isn't a Model? The caller might not have seen our method definition, it could be called by code that only knows the parent method definition, maybe library code will call it.

NoelLH:
@bdsl I can't use a native type because the method signature is then incompatible with its superclass.

If I remove the doc block override I get the error I started with due to the lack of type in the library:
```
ERROR: MissingParamType - src/Application/Persistence/RetrySafeEntityManager.php:101:29 - Parameter $object has no provided type (see https://psalm.dev/154)
    public function refresh($object, ?int $lockMode = null): void
```

bdsl:
@NoelLH ah I see. I can give it a go, but maybe just make the code what you think it should be and use @psalm-suppress to suppress the error.

src/Application/Persistence/RetrySafeEntityManager.php
```diff
@@ -92,14 +93,31 @@ public function persist($object): void
         }
     }
 
+
+    /**
+     * @param Model $object
```
bdsl, line 98:
```suggestion
     * @param object $object
```

bdsl:
@NoelLH I made a PR to your PR - https://github.com/thebiggive/matchbot/pull/528 fixes the Psalm issues.

tests/Application/Persistence/RetrySafeEntityManagerTest.php
```diff
@@ -152,6 +152,65 @@ public function testFlushWithRetry(): void
...
+
+        $retrySafeEntityManagerReflected = new ReflectionClass($this->retrySafeEntityManager);
+
+        $emProperty = $retrySafeEntityManagerReflected->getProperty('entityManager');
+        $emProperty->setValue($this->retrySafeEntityManager, $underlyingEmProphecy->reveal());
+
+        $app = $this->getAppInstance();
+        /** @var Container $container */
+        $container = $app->getContainer();
+        $container->set(EntityManager::class, $underlyingEmProphecy->reveal());
```
bdsl, line 170:
```suggestion
```

Is there any reason for these lines? The test passes without them.

tests/Application/Persistence/RetrySafeEntityManagerTest.php
```diff
@@ -152,6 +152,65 @@ public function testFlushWithRetry(): void
...
     }
 
+    public function testRefresh(): void
+    {
+        $underlyingEmProphecy = $this->prophesize(EntityManager::class);
+        $underlyingEmProphecy
+            ->refresh(Argument::type(Donation::class),LockMode::PESSIMISTIC_WRITE)
+            ->shouldBeCalledOnce();
+
+        $retrySafeEntityManagerReflected = new ReflectionClass($this->retrySafeEntityManager);
```
bdsl, line 162: I know this is following the pattern of the existing `testPersist`, but needing reflection in tests is often a sign that the subject under test could be better designed. Any reason why RetrySafeEntityManager shouldn't take its underlying entityManger via constructor injection so that it would be straight forward to replace in tests?

tests/Application/Persistence/RetrySafeEntityManagerTest.php
```diff
@@ -152,6 +152,65 @@ public function testFlushWithRetry(): void
         $this->retrySafeEntityManager->flush();
     }
 
+    public function testRefresh(): void
```
bdsl, line 155: IMHO tests get easier to read if we use full sentance names, e.g. `testRefreshCallsUnderlyingRefresh`

But honestly the code under test here seem so simple I'm not sure it's really worth having a unit test for it - what plausible mistakes would the unit test help us catch that won't be caught just as well by existing integration tests and/or Psalm? This isn't testing the case where the underlying throws, so it's really just testing:

```
    public function refresh($object, ?int $lockMode = null): void
    {
            $this->entityManager->refresh($object, $lockMode);
     }

File: tests/Application/Persistence/RetrySafeEntityManagerTest.php [hunk 5]
```diff
@@ -152,6 +152,65 @@ public function testFlushWithRetry(): void
...
+            $underlyingEmProphecy1->reveal(),
+            $underlyingEmProphecy2->reveal(),
+        );
+
+        $container->set(EntityManager::class, $underlyingEmProphecy1->reveal());
+        $container->set(RetrySafeEntityManager::class, $this->retrySafeEntityManager);
+        $container->set(EntityManagerInterface::class, $this->retrySafeEntityManager);
+
+        $this->retrySafeEntityManager->refresh(new Donation(), LockMode::PESSIMISTIC_WRITE);
+    }
```
bdsl, line 212: I simplified this test quite a lot - I think it still this version still tests what you want to test, and is easier to understand and less coupled to various things. Let me know what you think?

```suggestion
    public function testRefreshRetriesOnEmClosed(): void
    {
        // First underlying EM should throw closed error.
        $underlyingEmProphecy1 = $this->prophesize(EntityManager::class);
        $underlyingEmProphecy1
            ->refresh(Argument::type(Donation::class), LockMode::PESSIMISTIC_WRITE)
            ->willThrow(new EntityManagerClosed());

        // Second should work.
        $underlyingEmProphecy2 = $this->prophesize(EntityManager::class);
        $underlyingEmProphecy2
            ->refresh(Argument::type(Donation::class), LockMode::PESSIMISTIC_WRITE)
            ->shouldBeCalledOnce();

        $container = $this->getAppInstance()->getContainer();
        assert($container instanceof Container);
        $retrySafeEntityManager = $this->getRetrySafeEntityManagerPartialMock(
            $this->getConfiguration($container),
            ['driver' => 'pdo_mysql'],
            $underlyingEmProphecy1->reveal(),
            $underlyingEmProphecy2->reveal(),
        );

        $retrySafeEntityManager->refresh(new Donation(), LockMode::PESSIMISTIC_WRITE);
    }
```

bdsl:
Production code LGTM if you merge my PR to your PR to get Psalm to pass. I think its worth simplifying the tests quite a bit, I added suggestions.

tests/Application/Persistence/RetrySafeEntityManagerTest.php
NoelLH: I don't recall really as it was 2019, but think my thinking was probably that we want to ensure the first build, on construct, is guaranteed to happen exactly the same way as much rarer resets when they call `buildEntityManager()`. If the EM was created first that might not be true.

tests/Application/Persistence/RetrySafeEntityManagerTest.php
NoelLH: I think it's still better to have a simple "sanity check" test than no test, as this also protects against future changes to the code under test.

tests/Application/Persistence/RetrySafeEntityManagerTest.php
NoelLH: Nope, good catch

tests/Application/Persistence/RetrySafeEntityManagerTest.php
NoelLH: Much better, thanks!

PR #537, opened 2023-05-12
MAT-293 – support over-refunds

NoelLH:
And tidy up our #stripe Slack notification config a bit

### DIFF
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -23,6 +23,7 @@ use MatchBot\Application\Messenger\Handler\GiftAidResultHandler;
 use MatchBot\Application\Messenger\Handler\StripePayoutHandler;
 use MatchBot\Application\Messenger\StripePayout;
 use MatchBot\Application\Messenger\Transport\ClaimBotTransport;
+use MatchBot\Application\Notifier\StripeChatterInterface;
 use MatchBot\Application\Persistence\RetrySafeEntityManager;
 use MatchBot\Application\SlackChannelChatterFactory;
 use MatchBot\Client;
@@ -91,9 +92,26 @@ return function (ContainerBuilder $containerBuilder) {
             return new Chatter($transport);
         },
 
-        SlackChannelChatterFactory::class => static function (ContainerInterface $c): SlackChannelChatterFactory {
+        StripeChatterInterface::class => static function (ContainerInterface $c): ChatterInterface {
+            /**
+             * @var array{
+             *    notifier: array{
+             *      slack: array{
+             *        stripe_channel: string
+             *      }
+             *    }
+             *  } $settings
+             */
             $settings = $c->get('settings');
-            assert(is_array($settings));
+            $stripeChannel = $settings['notifier']['slack']['stripe_channel'];
+            /** @var SlackChannelChatterFactory $chatterFactory */
+            $chatterFactory = $c->get(SlackChannelChatterFactory::class);
+            return $chatterFactory->makeChatter($stripeChannel);
+        },
+
+        SlackChannelChatterFactory::class => static function (ContainerInterface $c): SlackChannelChatterFactory {
+            /** @var array $settings */
+            $settings = $c->get('settings');
             /** @psalm-suppress MixedArrayAccess $token */
             $token = $settings['notifier']['slack']['api_token'];
             assert(is_string($token));
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -110,8 +110,8 @@ return function (ContainerBuilder $containerBuilder) {
         },
 
         SlackChannelChatterFactory::class => static function (ContainerInterface $c): SlackChannelChatterFactory {
-            /** @var array $settings */
             $settings = $c->get('settings');
+            assert(is_array($settings));
             /** @psalm-suppress MixedArrayAccess $token */
             $token = $settings['notifier']['slack']['api_token'];
             assert(is_string($token));
--- a/src/Application/Actions/Hooks/StripeChargeUpdate.php
+++ b/src/Application/Actions/Hooks/StripeChargeUpdate.php
@@ -221,6 +221,14 @@ class StripeChargeUpdate extends Stripe
         $charge = $event->data->object;
         $amountRefunded = $charge->amount_refunded; // int: pence.
 
+        // Available status' (pending, succeeded, failed, canceled),
+        // see: https://stripe.com/docs/api/refunds/object.
+        // For now we support the successful refund path (inc. partial refund IF it's for the tip amount),
+        // converting status to the one MatchBot + SF use.
+        if ($charge->status !== 'succeeded') {
+            return $this->validationError($response, sprintf('Unsupported Status "%s"', $charge->status));
+        }
+
         $this->entityManager->beginTransaction();
 
         /** @var Donation $donation */
@@ -242,15 +250,6 @@ class StripeChargeUpdate extends Stripe
             $donation->getAmountFractionalIncTip() === $charge->amount_captured &&
             $amountRefunded > $donation->getAmountFractionalIncTip()
         );
-
-        // Available status' (pending, succeeded, failed, canceled),
-        // see: https://stripe.com/docs/api/refunds/object.
-        // For now we support the successful refund path (inc. partial refund IF it's for the tip amount),
-        // converting status to the one MatchBot + SF use.
-        if ($charge->status !== 'succeeded') {
-            $this->entityManager->rollback();
-            return $this->validationError($response, sprintf('Unsupported Status "%s"', $charge->status));
-        }
 
         if ($isTipRefund) {
             $this->logger->info(sprintf(
--- a/src/Application/Actions/Hooks/StripeChargeUpdate.php
+++ b/src/Application/Actions/Hooks/StripeChargeUpdate.php
@@ -296,10 +296,7 @@ class StripeChargeUpdate extends Stripe
     }
 
     /**
-     * @param string $eventType
-     * @param Donation $donation
      * @param int $refundedOrDisputedAmount In small currency unit, e.g. pence or cents.
-     * @param string $refundedOrDisputedCurrencyCode
      */
     private function warnAboutOverRefund(
         string $eventType,
--- a/src/Application/Actions/Hooks/StripeChargeUpdate.php
+++ b/src/Application/Actions/Hooks/StripeChargeUpdate.php
@@ -6,6 +6,7 @@ namespace MatchBot\Application\Actions\Hooks;
 
 use Doctrine\ORM\EntityManagerInterface;
 use MatchBot\Application\Actions\ActionPayload;
+use MatchBot\Application\Notifier\StripeChatterInterface;
 use MatchBot\Domain\Donation;
 use MatchBot\Domain\DonationRepository;
 use MatchBot\Domain\DonationStatus;
@@ -17,6 +18,10 @@ use Stripe\Charge;
 use Stripe\Dispute;
 use Stripe\Event;
 use Stripe\StripeClient;
+use Symfony\Component\Notifier\Bridge\Slack\Block\SlackHeaderBlock;
+use Symfony\Component\Notifier\Bridge\Slack\Block\SlackSectionBlock;
+use Symfony\Component\Notifier\Bridge\Slack\SlackOptions;
+use Symfony\Component\Notifier\Message\ChatMessage;
 
 /**
  * Handle charge.succeeded, charge.refunded and charge.dispute.closed events from a Stripe Direct webhook.
@@ -28,6 +33,7 @@ class StripeChargeUpdate extends Stripe
     public function __construct(
         protected DonationRepository $donationRepository,
         protected EntityManagerInterface $entityManager,
+        protected StripeChatterInterface $chatter,
         protected StripeClient $stripeClient,
         ContainerInterface $container,
         LoggerInterface $logger,
@@ -170,7 +176,8 @@ class StripeChargeUpdate extends Stripe
             return $this->respond($response, new ActionPayload(204));
         }
 
-        if ($donation->getAmountFractionalIncTip() !== $dispute->amount) {
+        if ($donation->getAmountFractionalIncTip() > $dispute->amount) {
+            // Less than the original amount was returned/refunded, which we don't expect.
             $this->logger->error(sprintf(
                 'Skipping unexpected dispute lost amount %s pence for donation %s based on Payment Intent ID %s',
                 $dispute->amount, // int: pence / cents.
@@ -180,6 +187,18 @@ class StripeChargeUpdate extends Stripe
             $this->entityManager->rollback();
 
             return $this->respond($response, new ActionPayload(204));
+        }
+
+        if ($donation->getAmountFractionalIncTip() < $dispute->amount) {
+            // More than the original amount was returned/refunded, which we don't
+            // *really* expect but are best to continue processing as a full refund
+            // with a warning.
+            $this->warnAboutOverRefund(
+                'charge.dispute.closed (lost)',
+                $donation,
+                $dispute->amount,
+                $dispute->currency,
+            );
         }
 
         $this->logger->info(sprintf(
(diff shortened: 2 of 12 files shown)


tests/Application/Actions/Hooks/StripeChargeUpdateTest.php
```diff
@@ -367,6 +372,64 @@ public function testSuccessfulFullRefund(): void
...
+
+        $request = $this->createRequest('POST', '/hooks/stripe', $body)
+            ->withHeader('Stripe-Signature', $this->generateSignature($time, $body, $webhookSecret));
+
+        $response = $app->handle($request);
+
+        $this->assertEquals('ch_externalId_123', $donation->getChargeId());
+        $this->assertEquals(DonationStatus::Refunded, $donation->getDonationStatus());
+        $this->assertEquals('1.00', $donation->getTipAmount());
+        $this->assertEquals(200, $response->getStatusCode());
```
bdsl, line 408: this test has now got quite long at imho at 42 lines, its hard to see exactly what it's testing - especially when the name is just `testSuccessfulFullRefund`. May be worth splitting it up into multiple tests, extracting functions, or even just breaking it up into with given/when/then or arrange/act/assert comments?

bdsl: I could comment on details above but probably not worth adding more before I get your thoughts on the first comment @NoelLH . Happy to pair on splitting it / making it clearer.

tests/Application/Actions/Hooks/StripeChargeUpdateTest.php
```diff
@@ -367,6 +372,64 @@ public function testSuccessfulFullRefund(): void
...
+    }
+
+    public function testSuccessfulOverRefund(): void
```
bdsl, line 411: `Successful` feels like a bit of a meaninguless word to me. Would it be possible to make the test name a full sentence that describes the main thing we're asserting? Looking at the test I think it might be `testOverRefundSendsSlackNoticeAndUpdatesRecord`. But the fact that there's an `and` in there makes me think maybe it should be two tests.

tests/Application/Actions/Hooks/StripeChargeUpdateTest.php
```diff
@@ -452,7 +516,15 @@ public function testUnsupportedRefundAmount(): void
 
         $body = $this->getStripeHookMock('ch_unsupported_partial_refund');
         $donation = $this->getTestDonation();
-        $webhookSecret = $container->get('settings')['stripe']['accountWebhookSecret'];
+        /**
+         * @var array{
+         *    stripe: array{
+         *      accountWebhookSecret: string
```
bdsl, line 522: The need to repeat this and similar makes me think it might be worth changing the value of 'settings' in the container from an array to an object with an actual class. Could be read-only with all public properties. Not a blocker here, but might be worth doing before repeating this sort of thing too many times. Using `@var` is error-prone, there's no check that the var actually hold what we say it holds.

app/dependencies.php
```diff
@@ -91,9 +92,26 @@ function (ContainerInterface $c): Auth\DonationPublicAuthMiddleware {
...
+        StripeChatterInterface::class => static function (ContainerInterface $c): ChatterInterface {
+            /**
+             * @var array{
+             *    notifier: array{
+             *      slack: array{
+             *        stripe_channel: string
+             *      }
+             *    }
+             *  } $settings
+             */
```
bdsl, line 104: See https://github.com/thebiggive/matchbot/pull/537/files#r1192272402

app/dependencies.php
```diff
@@ -91,9 +92,26 @@ function (ContainerInterface $c): Auth\DonationPublicAuthMiddleware {
...
+            $settings = $c->get('settings');
+            $stripeChannel = $settings['notifier']['slack']['stripe_channel'];
+            /** @var SlackChannelChatterFactory $chatterFactory */
+            $chatterFactory = $c->get(SlackChannelChatterFactory::class);
+            return $chatterFactory->makeChatter($stripeChannel);
+        },
+
         SlackChannelChatterFactory::class => static function (ContainerInterface $c): SlackChannelChatterFactory {
+            /** @var array $settings */
             $settings = $c->get('settings');
```
bdsl, line 114: Why the change from `assert` to `@var`? Seem to be going in the wrong direction, away from type safety.

src/Application/Actions/Hooks/StripeChargeUpdate.php
```diff
@@ -182,6 +189,18 @@ private function handleChargeDisputeClosed(Event $event, Response $response): Re
             return $this->respond($response, new ActionPayload(204));
         }
 
+        if ($donation->getAmountFractionalIncTip() < $dispute->amount) {
```
bdsl, line 192: minor point but since we have more than in the comments imho it's easier a bit easier to follow if we use the more than symbol in the code too?
```suggestion
        if ($dispute->amount > $donation->getAmountFractionalIncTip()) {
```

src/Application/Actions/Hooks/StripeChargeUpdate.php
```diff
@@ -214,6 +234,14 @@ private function handleChargeRefunded(Event $event, Response $response): Respons
 
         $isTipRefund = $donation->getTipAmountFractional() === $amountRefunded;
         $isFullRefund = $donation->getAmountFractionalIncTip() === $amountRefunded;
+        // If things disagree about the original captured amount, this is still uncharted
+        // territory and we should refrain from processing. If that matches but there is an
+        // over-refund, we'll want to notify the team but we know we are best off handling it
+        // like a simple refund from a donation status perspective.
+        $isOverRefund = (
+            $donation->getAmountFractionalIncTip() === $charge->amount_captured &&
+            $donation->getAmountFractionalIncTip() < $amountRefunded
```
bdsl, line 243: IMHO easier to read this way - we think of it as an over-refund, not an under-donation:
```suggestion
             $amountRefunded > $donation->getAmountFractionalIncTip()
```

bdsl, line 243: IMHO easier to read this way - we think of it as an over-refund, not an under-donation:
```suggestion
             $amountRefunded > $donation->getAmountFractionalIncTip()
```

tests/Application/Actions/Hooks/StripeChargeUpdateTest.php
NoelLH: @bdsl I think we talked about this a couple of weeks ago. The latest Slim Skeleton and newer apps do this so we have a model to follow, but it's not a super quick change.

app/dependencies.php
NoelLH: I guess no great reason except that it was closer to the other psalm-placating changes I was thinking about, and as discussed `$settings` being a generic array will likely be temporary anyway.

tests/Application/Actions/Hooks/StripeChargeUpdateTest.php
NoelLH: For now I've given it this name and added a comment. I think that releasing match funds is important too – adding this test helped me spot that the logic for that needed to change.

src/Application/Actions/Hooks/StripeChargeUpdate.php
```diff
@@ -182,6 +189,18 @@ private function handleChargeDisputeClosed(Event $event, Response $response): Re
...
+        if ($donation->getAmountFractionalIncTip() < $dispute->amount) {
+            // More than the original amount was returned/refunded, which we don't
+            // *really* expect but are best to continue processing as a full refund
+            // with a warning.
+            $this->warnAboutOverRefund(
+                'charge.dispute.closed (lost)',
+                $donation,
+                $dispute->amount,
+                $dispute->currency,
+            );
```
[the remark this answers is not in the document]
NoelLH: Yep true – I'm working on this

src/Application/Actions/Hooks/StripeChargeUpdate.php
```diff
@@ -249,11 +291,48 @@ private function handleChargeRefunded(Event $event, Response $response): Respons
...
 
         return $this->respondWithData($response, $charge);
     }
 
+    /**
+     * @param string $eventType
+     * @param Donation $donation
+     * @param int $refundedOrDisputedAmount In small currency unit, e.g. pence or cents.
+     * @param string $refundedOrDisputedCurrencyCode
+     */
```
bdsl, line 304: IMHO easier to see the important parts if we leave out the docblock part that jsut repeats what's in the code:
```suggestion
     * @param int $refundedOrDisputedAmount In small currency unit, e.g. pence or cents.
     */
```

I appreciate the specification that $refundedOrDisputedAmount is in a small currency unit - but also the fact that we need to specify it makes me think it'd would be good to introduce a value object type and use that instead of `int`. I'd probably combine the amount with the currency So the calling code could look something like:

```php
            $this->warnAboutOverRefund(
                'charge.dispute.closed (lost)',
                $donation,
                Money::of(currencySymbol: $dispute->currency, minorUnits: $dispute->amount)
            );
```

src/Application/Actions/Hooks/StripeChargeUpdate.php
```diff
@@ -214,6 +234,14 @@ private function handleChargeRefunded(Event $event, Response $response): Respons
...
+        // territory and we should refrain from processing. If that matches but there is an
+        // over-refund, we'll want to notify the team but we know we are best off handling it
+        // like a simple refund from a donation status perspective.
+        $isOverRefund = (
+            $donation->getAmountFractionalIncTip() === $charge->amount_captured &&
+            $donation->getAmountFractionalIncTip() < $amountRefunded
+        );
 
         // Available status' (pending, succeeded, failed, canceled),
         // see: https://stripe.com/docs/api/refunds/object.
```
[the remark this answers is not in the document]
NoelLH: I guess we don't – will update

src/Application/Notifier/StripeChatterInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace MatchBot\Application\Notifier;
+
+use Symfony\Component\Notifier\ChatterInterface;
+
+interface StripeChatterInterface extends ChatterInterface
+{}
```
bdsl, line 8: Feels strange to extend the interface without adding anything. This is just so that the container can distinguish the two instances and know where to inject each one right? Not a blocker but I feel like there should be a nicer way to do that.

And actually if we are gong the way of extending it don't we really need to give it two children - one `StripeChatterInterface` and one `MatchbotChatterInterface`? Otherwise it looks like anything that declares a dependency on the parent interface doesn't care which channel its messages go to.

src/Application/Actions/Hooks/StripeChargeUpdate.php
bdsl: on the other hand as this is a private function it's not a big issue to manually check that the usage semantics are correct. Just a more general point.

src/Application/Notifier/StripeChatterInterface.php
NoelLH: Yes, that's why. I think it's OK for consumers to essentially "not care" for now, it's not the end of the world if we have a default channel that doesn't require much thought. I'd be interested in other options than extending the interface but couldn't immediately think of one that would work well with Slim DI. This felt like an improvement on essentially doing DI stuff inside the payout handler as we were before.

tests/Application/Actions/Hooks/StripeChargeUpdateTest.php
NoelLH: I've added grouping comments for now.

tests/Application/Actions/Hooks/StripeChargeUpdateTest.php
```diff
@@ -367,6 +372,70 @@ public function testSuccessfulFullRefund(): void
         $webhookSecret = $container->get('settings')['stripe']['accountWebhookSecret'];
         $time = (string) time();
 
+        $chatterProphecy = $this->prophesize(StripeChatterInterface::class);
+        // Double-check that the normal success case isn't messaging Slack.
+        $chatterProphecy->send(Argument::type(ChatMessage::class))->shouldNotBeCalled();
```
bdsl, line 434: Can make a slightly stronger / more general assertion.
```suggestion
        $chatterProphecy->send(Argument::cetera())->shouldNotBeCalled();
```
I.e. we don't care what the arguments are or aren't, this method should just not be called at all.

src/Application/Notifier/StripeChatterInterface.php
bdsl: It's not that we shouldn't have a default channel, it's that the the consumers don't specify that they want the default channel. If the parent interface promises to give them a default channel then to be a Liskov-subtype the child interface should give them the default channel too. Can leave it for now, although I think making the third interface would make it easier to understand.

tests/Application/Actions/Hooks/StripeChargeUpdateTest.php
```diff
@@ -367,6 +428,73 @@ public function testSuccessfulFullRefund(): void
...
+        $this->assertEquals(DonationStatus::Refunded, $donation->getDonationStatus());
+        $this->assertEquals('1.00', $donation->getTipAmount());
+        $this->assertEquals(200, $response->getStatusCode());
+    }
+
+    /**
+     * Verifies that:
+     * * Slack gets a warning with the correct details
+     * * The donation record is updated with the new status, and
+     * * Match funds are released.
```
bdsl, line 475: :+1:

tests/Application/Actions/Hooks/StripeChargeUpdateTest.php
```diff
@@ -367,6 +428,73 @@ public function testSuccessfulFullRefund(): void
...
+     */
+    public function testOverRefundSendsSlackNoticeAndUpdatesRecord(): void
+    {
+        // arrange
+        $app = $this->getAppInstance();
+        /** @var Container $container */
+        $container = $app->getContainer();
```
bdsl, line 482: Probably not worth changing now, but I wonder if using the container and the app instance actually gives us much in a unit test? I feel like its sometimes easier to see exactly what the code being tested is if the test directly instantiates e.g. the `StripeChargeUpdate` class and passes in the dependencies.  Or if we want to test the app as a whole maybe go the other way and do an integration test with the real DB then we test the repository code too.

tests/Application/Actions/Hooks/StripeChargeUpdateTest.php
```diff
@@ -367,6 +428,73 @@ public function testSuccessfulFullRefund(): void
...
+        $time = (string) time();
+
+        $chatterProphecy = $this->prophesize(StripeChatterInterface::class);
+
+        $options = (new SlackOptions())
+            ->block((new SlackHeaderBlock('[test] Over-refund detected')))
+            ->block((new SlackSectionBlock())->text('Over-refund detected for donation 12345678-1234-1234-1234-1234567890ab based on charge.refunded hook. Donation inc. tip was 124.45 GBP and refund or dispute was 134.45 GBP'))
+            ->iconEmoji(':o');
+        $message = (new ChatMessage('Over-refund detected'))
+            ->options($options);
```
bdsl, line 496: Could rename `$message` to `$expectedMessage` to make it clearer - but I think even better is just to inline all this stuff so it appears only in the assert section, e.g.:

```php
        $chatterProphecy->send((new ChatMessage('Over-refund detected'))
            ->options((new SlackOptions())
                ->block((new SlackHeaderBlock('[test] Over-refund detected')))
                ->block((new SlackSectionBlock())->text('Over-refund detected for donation 12345678-1234-1234-1234-1234567890ab based on charge.refunded hook. Donation inc. tip was 124.45 GBP and refund or dispute was 134.45 GBP'))
                ->iconEmoji(':o')))->shouldBeCalledOnce();
```

Also TIL that prophecy will let you do `shouldBeCalled` after the the action, using the revelation as a spy instead of a mock. Nice!

NoelLH: Yeah, in theory that would be nicer. I tried moving some of the calls to `// assert` but of course if they need to define things about the prophesised method's behaviour in the same chained call then to do it with the fluent interface you need both things before you `// act`

bdsl:
Initial review done, all LGTM other than where I've commented. Just going through updates now.

bdsl (approved):
LGTM - thanks for the adjustments.

tests/Application/Actions/Hooks/StripeChargeUpdateTest.php
NoelLH: I think it was moving donation repo calls at the same time which broke things, but I didn't dig into it further

PR #545, opened 2023-05-17
MAT-295: Detatch payment cards from passwordless accounts

bdsl:
Changes the logic of the payment method detacher to work based on newly complete metadata in stripe - and also adds an extra metadata field to stripe to track which customer accounts the command has operated on them so that it won't find the same account in search next time it runs.

### DIFF
--- a/tests/Application/Commands/DeleteStalePaymentDetailsTest.php
+++ b/tests/Application/Commands/DeleteStalePaymentDetailsTest.php
@@ -32,13 +32,6 @@ class DeleteStalePaymentDetailsTest extends TestCase
         $testCustomerId = 'cus_aaaaaaaaaaaa11';
         $testPaymentMethodId = 'pm_aaaaaaaaaaaa13';
 
-        $stripeChargesProphecy = $this->prophesize(ChargeService::class);
-        $stripeChargesProphecy->search([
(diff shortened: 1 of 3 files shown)


tests/Application/Commands/DeleteStalePaymentDetailsTest.php
```diff
@@ -51,7 +44,7 @@ public function testDeletingOneStaleCard(): void
 
         $stripeCustomersProphecy = $this->prophesize(CustomerService::class);
         $stripeCustomersProphecy->search([
-            'query' => "created<{$previousDay->getTimestamp()} and metadata['hasPasswordSince']:null",
+            'query' => "created<{$previousDay->getTimestamp()} and metadata['hasPasswordSince']:null and metadata['paymentMethodsCleared']:null",
```
bdsl, line 47: If there's a way to write a search query that will only return accounts with attached payment methods it would be simpler and a bit more elegant, but I think this should work fine and is good enough.

dee-jo (approved):
LGTM

tests/Application/Commands/DeleteStalePaymentDetailsTest.php
NoelLH: @bdsl I'm pretty sure theere isn't. https://stripe.com/docs/search#query-fields-for-customers

NoelLH (approved):

PR #547, opened 2023-05-20
MAT-302 – fix Update lock wait recovery

NoelLH:
Also:
* Reduce `Update`'s database access when we can fail validation earlier
* Use UUID over ID a bit more clearly & consistently in logs etc.

### DIFF
--- a/src/Application/Actions/Donations/Update.php
+++ b/src/Application/Actions/Donations/Update.php
@@ -59,22 +59,6 @@ class Update extends Action
             throw new DomainRecordNotFoundException('Missing donation ID');
         }
 
-        /** @var Donation $donation */
-        $this->entityManager->beginTransaction();
-
-        try {
-            $donation = $this->getDonationWithRetry($args['donationId']);
-        } catch (LockWaitTimeoutException $e) {
-            $this->logger->error("Lock Wait Timeout for donationId '{$args['donationId']}' even after retry");
-            throw $e;
-        }
-
-        if (!$donation) {
-            $this->entityManager->rollback();
-
-            throw new DomainRecordNotFoundException('Donation not found');
-        }
-
         $body = (string) $request->getBody();
 
         try {
@@ -91,8 +75,6 @@ class Update extends Action
             $message = 'Donation Update data deserialise error';
             $exceptionType = get_class($exception);
 
-            $this->entityManager->rollback();
-
             return $this->validationError($response,
                 "$message: $exceptionType - {$exception->getMessage()}",
                 $message,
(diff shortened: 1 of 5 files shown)


src/Application/Actions/Donations/Update.php
```diff
@@ -59,22 +59,6 @@ protected function action(Request $request, Response $response, array $args): Re
...
-
-        try {
-            $donation = $this->getDonationWithRetry($args['donationId']);
-        } catch (LockWaitTimeoutException $e) {
-            $this->logger->error("Lock Wait Timeout for donationId '{$args['donationId']}' even after retry");
-            throw $e;
-        }
-
-        if (!$donation) {
-            $this->entityManager->rollback();
```
bdsl, line 73: Not a problem since this is deleted, but I'm confused now about why we ever did rollback here, since it seems like no commands have been issued that would require rolling back.

src/Application/Actions/Donations/Update.php
```diff
@@ -101,21 +83,27 @@ protected function action(Request $request, Response $response, array $args): Re
         }
 
         if (!isset($donationData->status)) {
-            $this->entityManager->rollback();
```
bdsl, line 104: same here and above - what changes needed rolling back? So deleting these lines looks good.

src/Application/Actions/Donations/Update.php
```diff
@@ -101,21 +83,27 @@ protected function action(Request $request, Response $response, array $args): Re
...
         while ($retryCount < self::MAX_UPDATE_RETRY_COUNT) {
+            $this->entityManager->beginTransaction();
+
             try {
-                if ($retryCount > 0) {
-                    $this->entityManager->refresh($donation, LockMode::PESSIMISTIC_WRITE);
+                $donation = $this->donationRepository->findAndLockOneBy(['uuid' => $args['donationId']]);
+
+                if (!$donation) {
+                    $this->entityManager->rollback();
```
bdsl, line 102: as with the deleted rollbacks, I'm confused about what we're actually rolling back, since it seems like we've only queried the DB so far, not commanded it. The only time we send a command the the DB seems to be via the call to `addData` on line 120.

bdsl: @bdsl No, we have sent a command to the DB - we have commanded it to not allow any concurrent read or write operations on this row of the donations table when we called `\MatchBot\Domain\DonationRepository::findAndLockOneBy` Rolling back the transaction unlocks the row and does make sense. I think it would happen naturally at the end of the request but it makes sense to do it early.

tests/Application/Actions/Donations/UpdateHandlesLockExceptionTest.php
```diff
@@ -184,12 +193,11 @@ public function setExpectationsForPersistAfterRetry(string $donationId, Donation
             return null;
```
bdsl, line 193: I know this is outside the diff, and probably my mistake originally, but it seems weird that the `$newStatus` param is unused. Maybe we should assert on it like this?

```suggestion
            TestCase::assertEquals($newStatus, $donation->getDonationStatus());
            return null;
```
Need to also `use` $newStatus and $donation at line 188.

bdsl (approved):
LGTM - validate input before starting the DB transaction & lock, no need to lock up the row if the input is unusable. Good rule to follow generally, and If we've got significant spam traffic that doesn't pass the validation then it may make a big difference.

Added one suggestion on the tests.

bdsl (approved):

PR #548, opened 2023-05-23
MAT-300 – update some CC22 donations for internal Gift Aid processing

NoelLH:


src/Migrations/Version20230523154501.php
```diff
@@ -0,0 +1,36 @@
...
+    public function getDescription(): string
+    {
+        return 'Include some CC22 donations for Gift Aid claims';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql(<<<SQL
+            UPDATE Donation SET tbgShouldProcessGiftAid = 1
+            WHERE campaign_id = 4216 AND
```
bdsl, line 24: Yes, I've checked in the DB, this is the ID of the most recent campaign for the relevant charity.

PR #550, opened 2023-05-30
MAT-301 – customer balance auto-confirm edge case fixes

NoelLH:


src/Application/Actions/Donations/Update.php
```diff
@@ -113,6 +113,20 @@ protected function action(Request $request, Response $response, array $args): Re
                     );
                 }
 
+                if ($donationData->autoConfirmFromCashBalance && $donation->getPaymentMethodType() !== 'customer_balance') {
```
bdsl, line 116: not necessarily now, but at some point it might be good to make payment method type an enum, with just the two cases `card` and `customerBalance` - and we might be able to share the enum between the domain model and the DTO. Or we could more cheaply leave it as-is at runtime but type it statically as `'card'|'customer_balance'` instead of `string`. Could also make it an enum in the db scehma.

bdsl: Why would a non customer balance donation have `autoConfirmFromCashBalance` set? Would it have to be a bug in the client? I see our frontend only sets this in [finaliseCashBalancePurchase](https://github.com/thebiggive/donate-frontend/blob/853dd7101148255a57e0118a3e36e682f277994b/src/app/donation.service.ts#L135)

src/Application/Actions/Donations/Update.php
```diff
@@ -113,6 +113,20 @@ protected function action(Request $request, Response $response, array $args): Re
...
+                    $this->entityManager->rollback();
+
+                    // Log a warning to more easily spot occurrences in dashboards.
+                    $this->logger->warning(
+                        "Donation ID {$args['donationId']} auto-confirm attempted with '{$donation->getPaymentMethodType()}' payment method",
+                    );
+
+                    return $this->validationError($response,
+                        "Donation ID {$args['donationId']} could not be auto-confirmed",
+                        'Processing incomplete. Please refresh and check your donation funds balance'
```
bdsl, line 126: If this indicates a bug in the client maybe the message needs to be different, e.g. 
```suggestion
                        "Processing incomplete. Auto-confirm is not possible for a '{$donation->getPaymentMethodType()}' donation"
```

src/Application/Actions/Donations/Update.php
```diff
@@ -258,10 +272,35 @@ private function addData(Donation $donation, HttpModels\Donation $donationData,
...
+                } catch (InvalidRequestException $exception) {
+                    // Currently a typical Update call which auto-confirms is being made for just
+                    // that purpose. So our safest options are to return a 500 and roll back any
+                    // database changes.
+                    $this->entityManager->rollback();
+
+                    // To help analyse it quicker we handle the specific auto-confirm API failure we've
+                    // seen before with a distinct message, but both options give the client an HTTP 500,
+                    // as we expect neither with our updated guard conditions.
+                    if (str_contains($exception->getMessage(), 'missing a payment method')) {
```
bdsl, line 286: Maybe worth making this a bit more specific, to make sure we don't catch any similar but different errors here?

```suggestion
                    if (str_starts_with($exception->getMessage(), "You cannot confirm this PaymentIntent because it's missing a payment method")) {

```

src/Application/Actions/Donations/Update.php
```diff
@@ -258,10 +272,35 @@ private function addData(Donation $donation, HttpModels\Donation $donationData,
...
+                            'Stripe Payment Intent for donation ID %s was missing a payment method, so we could not confirm it',
+                            $donation->getUuid(),
+                        ));
+                        $error = new ActionError(ActionError::SERVER_ERROR, 'Could not confirm Stripe Payment Intent');
+
+                        return $this->respond($response, new ActionPayload(500, null, $error));
+                    }
+
+                    throw $exception;
+                }
```
bdsl, line 297: Is it worth adding a second catch to catch any other species of `\Stripe\Exception\ApiErrorException` and wrapping it in our own? Or just adding the `@throws` tags on this method and on `action`?

src/Application/Actions/Donations/Update.php
```diff
@@ -258,10 +272,35 @@ private function addData(Donation $donation, HttpModels\Donation $donationData,
...
             }
         }
 
+        // We hold off finishing the transaction for now, because the current logic
+        // involves rolling back our whole Update in *some* Stripe error scenarios.
+        // We don't have enough info to detect these without making the Stripe API call.
```
bdsl, line 303: I'm not sure I follow - if the transaction isn't finished here where does it get committed? I don't see any opportunity for a stripe error to happen after this line.

tests/Application/Actions/Donations/UpdateTest.php
```diff
@@ -1657,24 +1657,13 @@ public function testAddDataSuccessWithCashBalanceAutoconfirm(): void
...
-        $donation->setDonorHomeAddressLine1('99 Updated St');
-        $donation->setDonorHomePostcode('X1 1XY');
-        $donation->setDonorFirstName('Saul');
-        $donation->setDonorLastName('Williams');
-        $donation->setDonorEmailAddress('saul@example.com');
-        $donation->setTbgComms(true);
-        $donation->setCharityComms(false);
-        $donation->setChampionComms(false);
-        $donation->setDonorBillingAddress('Y1 1YX');
+        $donation = $this->prepareDonationBasics($donation);
```
bdsl, line 1660: Liking moving all these setter calls into a function, but:
- it seems to be overwriting many of the properties set in `\MatchBot\Tests\Application\DonationTestDataTrait::getTestDonation` making me thing we shouldn't need to use both.
- it needlessly returns the same object that was passed to it, which I think is a bit misleading. Better to either have it create an object (maybe by calling getTestDonation inside the function) or operate on a passed-in object and return void.

tests/Application/Actions/Donations/UpdateTest.php
```diff
@@ -1765,4 +1754,242 @@ public function testAddDataSuccessWithCashBalanceAutoconfirm(): void
...
+        $this->assertJson($payload);
+        $this->assertEquals(400, $response->getStatusCode());
+
+        $expectedPayload = new ActionPayload(400, ['error' => [
+            'type' => 'BAD_REQUEST',
+            'description' => 'Processing incomplete. Please refresh and check your donation funds balance',
+        ]]);
+        $expectedSerialised = json_encode($expectedPayload, JSON_PRETTY_PRINT);
+
+        $this->assertEquals($expectedSerialised, $payload);
```
bdsl, line 1826: I see what this is doing, and maybe we discussed this before, but it feels like this test is quite long and fragile and it's a bit hard to see which code is the actual subject under test. As the business logic is mixed with the database stuff a bit in the Update controller it seems like the only way to test it actually works would be to do a test with a real DB. But I also think the controller is getting a bit long and unwieldy. We don't really have any service layer at the moment, but maybe it would be worth creating one now, and then we could have a much simpler test that just expects the service to throw an exception when there's a request to `autoConfirmFromCashBalance` on a card payment.

tests/Application/Actions/Donations/UpdateTest.php
```diff
@@ -1765,4 +1754,242 @@ public function testAddDataSuccessWithCashBalanceAutoconfirm(): void
...
+        $donationInRepo->setPaymentMethodType('customer_balance');
+
+        $donationRepoProphecy
+            ->findAndLockOneBy(['uuid' => '12345678-1234-1234-1234-1234567890ab'])
+            ->willReturn($donationInRepo)
+            ->shouldBeCalledOnce();
+        $donationRepoProphecy
+            ->deriveFees(Argument::type(Donation::class), null, null)
+            ->willReturn($donation) // Actual fee calculation is tested elsewhere.
+            ->shouldBeCalledOnce();
```
bdsl, line 1853: not a new issue, but noticing that the return value of deriveFees is always identical to the first argument - so it might be worth making it return void which would make it a little bit simpler to use and to test with.

Even better but for more work, I think deriveFees doesn't do any I/O so it wouldn't it make sense to hide it inside the Donation model and make the donation derive its own fees whenever necessary?

tests/Application/Actions/Donations/UpdateTest.php
```diff
@@ -1765,4 +1754,242 @@ public function testAddDataSuccessWithCashBalanceAutoconfirm(): void
...
+        // We re-throw this unknown type and leave it to be handled "downstream" in
+        // the shutdown handler.
+        $this->expectException(InvalidRequestException::class);
+        $this->expectExceptionMessage('Not the one we know!');
+
+        // act
+        $app->handle($request->withAttribute('route', $route));
+    }
+
+    private function prepareDonationBasics(Donation $donation): Donation
```
bdsl, line 1976: See https://github.com/thebiggive/matchbot/pull/550/files?diff=unified&w=0#r1211422742

bdsl:
Added comments inline - no real blockers, my main thought is that `\MatchBot\Application\Actions\Donations\Update` is starting to get unwieldy at nearly 500 lines, and the tests take quite a long time to read as they have to set up so many dependencies of the code under test.

src/Application/Actions/Donations/Update.php
NoelLH: @bdsl I'm not totally sure. I think probably the best guess I have right now is the 3rd of the scenarios I suggested testing:

> 1. donating logged in in 2 different browsers on a Customer with credit: setting up a donation in each (using the same balance), completing 1 (which should work) and then attempting to complete the 2nd (which should give an error)
> 2. starting a donation logged out, then [if possible] returning to the start to log into an account with a balance before completing the original donation
> 3. starting a donation logged in but with £0 balance, then funding the account in Stripe test mode, refreshing and then choosing to continue the previously set up donation [which I think may also error at the end]

src/Application/Actions/Donations/Update.php
NoelLH: Yep, not sure. I had this in my local uncommited work from before this week, and missed that it no longer makes sense with the current logic. Removed.

src/Application/Actions/Donations/Update.php
NoelLH: Did the latter for now

tests/Application/Actions/Donations/UpdateTest.php
NoelLH: I think it's true that we could return `void` but I'm not sure it unlocks a great simplification for any tests.

`deriveFees()` does set 2 mapped properties on Donation which we need denormalised after each change in card country/brand etc., so I'm not sure it would work to make it synthetic like that. Most fundamentally, we need to support changes in fee pricing over time.

src/Application/Actions/Donations/Update.php
NoelLH: > not necessarily now, but at some point it might be good to make payment method type an enum, with just the two cases `card` and `customerBalance` - and we might be able to share the enum between the domain model and the DTO. Or we could more cheaply leave it as-is at runtime but type it statically as `'card'|'customer_balance'` instead of `string`. Could also make it an enum in the db scehma.

I'm nervous about the testing and work to move an already-live field to an enum at DB level, unless there is a substantial benefit relevant to this change.

I've started using a PHP `enum` for the property, along with some possible or necessary related simplifications, on https://github.com/thebiggive/matchbot/pull/550/commits/2670d7d922754e93678e2ceb7b1a67e8a7193127

I had to `@psalm-suppress MissingConstructor` on the HTTP model in order to use an enum in an object without a constructor. Not sure if there's a more elegant way to handle that?

tests/Application/Actions/Donations/UpdateTest.php
NoelLH: Good spot – I obviously misdiagnosed the problems with tests early on and massively overestimated how much of this we needed. I guess some of the existing`UpdateTest` scenarios also pre-dated us having a helper that's as complete as the one we do now, so I started off copying work which already had redundant stuff.

Simplified it significantly & removed some redundant assertions on https://github.com/thebiggive/matchbot/pull/550/commits/9ac27612d06da705e672a915244e5022be5a6bce

Unfortunately psalm's baseline updates have gone into a tailspin with my local so none of the commands give me a way to update it, and I've run out of time for the week to try to placate it.

src/Application/Actions/Donations/Update.php
bdsl: Will have to continue this tomorrow - the missing constructor complaint isn't because https://github.com/thebiggive/matchbot/commit/2670d7d922754e93678e2ceb7b1a67e8a7193127 changes the string to an enum, it's because it removes the default value from the declaration. Declaring the property as `public PaymentMethodType $paymentMethodType == PaymentMethodType::card`; should make that go away.

Whether or not it should have that default is a separate question to whether it should be an enum instead of a string.

Btw I think it will need to be a string-backed enum to let doctrine persist it to the db.

src/Domain/DonationRepository.php
```diff
@@ -177,7 +171,7 @@ public function buildFromApiRequest(DonationCreate $donationData): Donation
 
         $donation = new Donation();
         $donation->setPsp($donationData->psp);
-        $donation->setPaymentMethodType($donationData->paymentMethodType);
+        $donation->setPaymentMethodType($donationData->paymentMethodType->name);
```
bdsl, line 174: If we move the enum into the domain model (or even if we don't) we could share it between there and the HTTP model. Then this can be simplified to

```suggestion
        $donation->setPaymentMethodType($donationData->paymentMethodType);
```

I think that's a lot better. The enum has to be string-backed, and doctrine can be configured to persist it with an annotation like we have for the status field:

https://github.com/thebiggive/matchbot/blob/38dc7445c67ee1ccc763476046477227d56ef97a/src/Domain/Donation.php#L146

It doesn't mean it has to be an enum in MySQL.

tests/Application/Actions/Donations/UpdateTest.php
bdsl: Happy to have a look at it with you whenever you're next working. If there's ever an issue updating the baseline feel free to make the PR with the baseline not updated and ask me to update it. Another option is to ssh into a CircleCI executor and run the update there. We could also change the Psalm config to not require the baseline to be updated all the time, and just update it when we feel like it - although it is nice to see issues disappearing from the basline as we work on stuff.

tests/Application/Actions/Donations/UpdateTest.php
bdsl: Not any great simplification, just a small simplification in many places - test and production code. I might make a separate PR to change it to void.

I see what you mean about about fee pricing over time - the `deriveFees` is run at a specific time, applying the policy we have at that time.

src/Application/HttpModels/DonationCreate.php
```diff
@@ -6,6 +6,7 @@
 
 /**
  * Request-only payload for setting up new donations.
+ * @psalm-suppress MissingConstructor Seems we can't use an enum w/o default without hitting this.
```
bdsl, line 9: True we can't use an enum w/o default without hitting this, Psalm wants to make sure we never hit any Uninitialized property errors, which I think is a good thing. I'm not sure why you don't want to let it default to card as it was.

bdsl:
ah I've noticed I accidentally put comments on a commit instead of on the PR. Will copy them to here.

src/Application/HttpModels/PaymentMethodType.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MatchBot\Application\HttpModels;
+
+enum PaymentMethodType
+{
+    case card;
```
bdsl, line 9: (copied from commit comments)  can we name the cases in PascalCase? I went with that for the status because it's recommended in the PER standard.

src/Domain/DonationRepository.php
```diff
@@ -72,7 +73,7 @@ public function doCreate(SalesforceWriteProxy $donation): bool
     public function doUpdate(SalesforceWriteProxy $donation): bool
     {
         if ($donation->getPaymentMethodType() === null) {
-            $donation->setPaymentMethodType('card');
+            $donation->setPaymentMethodType(PaymentMethodType::card->name);
```
bdsl, line 76: (copied from commit comments) 

I think it's generally better not to use the `name` of an enum for anything except debugging - it's sort of reflectiony. If you need to convert an enum to string make it a backed enum and use the `value` property. But as I've commented elsewhere I think this can just change to `$donation->setPaymentMethodType(PaymentMethodType::card);`

bdsl:
I think I've reviewed all I can for now - happy to work together on this or re-review after changes.

src/Application/Actions/Donations/Update.php
NoelLH: @bdsl I think I realised last week that there were already errors in the baseline for other properties on the same model. So effectively we were suppressing it on each other property already.

src/Application/HttpModels/DonationCreate.php
bdsl: I think we probably shouldn't suppress the MissingConstructor error - I think making sure we initialise all required properties in constructors is a really good thing Psalm can do for us. Can we either set the default to card to match what it was doing before, or if you want to improve it add a constructor, e.g. as below and pass the arguments to it from tests? 

That will also make the symfony seralizer ensure those three arguments are always provided:

https://symfony.com/doc/current/components/serializer.html#handling-constructor-arguments

```
    public function __construct(
        string $donationAmount,
        PaymentMethodType $paymentMethodType,
        string $projectId,
        string $psp
    ) {
        $this->donationAmount = $donationAmount;
        $this->paymentMethodType = $paymentMethodType;
        $this->projectId = $projectId;
        $this->psp = $psp;
    }
    ```

bdsl (approved):
Still think it's worth tidying up the test(s) but we could do that in a separate PR.

PR #552, opened 2023-06-06
MAT-301 – fix missing normalizer

NoelLH:
And re-apply changes from https://github.com/thebiggive/matchbot/pull/550


bdsl (dismissed):
Adding the normalizer LGTM but it doesn't look like you're re-applying the changes from the reverted PR here.

PR #559, opened 2023-06-07
DON-800: Increase donation limit for donation account users to £200k

bdsl:
Fairly substantial refactoring involved in this PR, must of it done with the PhpStorm tools, and actually the changes in production code are minimal so it's pretty safe, it's mostly changes in test code. The main reason is that the new requirement has the maximum donation amount dependant on the payment method type, and so the donation class can only enforce that if it doesn't allow them to be set separately, or at least doesn't allow the method to be changed after the amount is set.

Our site doesn't actually support changing the amount on an existing donation, so there's no need for the class to support it. Now amount, currency code and payment method type are all required params of the domain donation constructor. In production we only ever construct a donation by calling `\MatchBot\Domain\Donation::fromApiModel` (or letting Doctrine instantiate it from the DB, which doesn't run use any constructor function).

The refactors take us most of the way towards https://github.com/thebiggive/matchbot/pull/509

I'm assuming we want the maximum tip to be the same as the maximum donation, as it has been up to now.

### DIFF
--- a/src/Domain/Donation.php
+++ b/src/Domain/Donation.php
@@ -298,15 +298,16 @@ class Donation extends SalesforceWriteProxy
      * @deprecated but retained for now as used in old test classes. Not recommend for continued use - either use
      * fromApiModel or create a new named constructor that takes required data for your use case.
      */
-    public static function emptyTestDonation(string $amount): self
+    public static function emptyTestDonation(string $amount, PaymentMethodType $paymentMethodType = PaymentMethodType::Card): self
     {
-        return new self($amount, 'GBP');
+        return new self($amount, 'GBP', $paymentMethodType);
     }
 
-    private function __construct(string $amount, string $currencyCode)
+    private function __construct(string $amount, string $currencyCode, PaymentMethodType $paymentMethodType)
     {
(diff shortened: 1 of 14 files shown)


src/Domain/Donation.php
```diff
@@ -291,9 +293,62 @@ class Donation extends SalesforceWriteProxy
...
+    {
+
+        $psp = $donationData->psp;
+        assert($psp === 'stripe');
+
+        $donation = new self($donationData->donationAmount, $donationData->currencyCode, $donationData->paymentMethodType);
+
+        $donation->setPsp($psp);
+        $donation->setDonationStatus(DonationStatus::Pending);
+        $donation->setUuid(Uuid::uuid4());
```
bdsl, line 323: Previously we had the more complex `$donation->setUuid((new UuidGenerator())->generateId($this->getEntityManager(), $donation));` but all that was doing in practice was calling `Uuid::uuid4()` for us in more polymorphic way.

dee-jo (approved):
LGTM!

dee-jo (commented):
LGTM

PR #563, opened 2023-06-16
DON-772 – Fix checking for customer balance tips under £1

NoelLH:
And tidy up comparison operator + types

Effectively post-merge review suggestions from a quick skim of https://github.com/thebiggive/matchbot/pull/560/files

### DIFF
--- a/src/Domain/Donation.php
+++ b/src/Domain/Donation.php
@@ -841,11 +841,9 @@ class Donation extends SalesforceWriteProxy
         return $this->tipAmount;
     }
 
-    /**
-     * @param string $tipAmount Numeric string
-     */
     public function setTipAmount(string $tipAmount): void
     {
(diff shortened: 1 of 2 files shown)


src/Domain/Donation.php
```diff
@@ -843,11 +843,10 @@ public function getTipAmount(): string
 
     public function setTipAmount(string $tipAmount): void
     {
-        /** @var numeric-string $tipAmount */
-
+        /** @psalm-var numeric-string $tipAmount */
         if (
-            $this->paymentMethodType == PaymentMethodType::CustomerBalance &&
-            bccomp($tipAmount, '0') !== 0
+            $this->paymentMethodType === PaymentMethodType::CustomerBalance &&
```
bdsl, line 848: Agreed we should use `===` whenever possible, although I don't think in this case it could do anything different to `==`.

PR #567, opened 2023-06-21
BG2-2306: Prepare to send refund timestamp to salesforce

bdsl:
Because the salesforce part isn't done yet, (see https://github.com/thebiggive/salesforce/pull/1316) the one line here that does the actual sending of new data to SF is commented out. But the code to pull it from Stripe into Matchbot's DB etc should be fine to start running whenever we want to deploy.

### DIFF
--- a/integrationTests/DonationPersistenceTest.php
+++ b/integrationTests/DonationPersistenceTest.php
@@ -15,6 +15,7 @@ class DonationPersistenceTest extends IntegrationTest
         $em = $this->getService(EntityManagerInterface::class);
         $connection = $em->getConnection();
         $donation = $this->makeDonationObject();
+        $donation->recordRefundAt(new \DateTimeImmutable('2023-06-22 10:00:00'));
 
         // act
         $em->persist($donation);
@@ -39,7 +40,7 @@ class DonationPersistenceTest extends IntegrationTest
         $donation = $donationRepo->findOneBy(['uuid' => $donationRow['uuid']]);
 
         $this->assertNotNull($donation);
-        $this->assertSame(DonationStatus::Collected, $donation->getDonationStatus());
+        $this->assertEquals(DonationStatus::Refunded, $donation->getDonationStatus());
     }
 
     /**
@@ -62,7 +63,7 @@ class DonationPersistenceTest extends IntegrationTest
             'uuid' => Uuid::uuid4(),
             'transactionId' => NULL,
             'amount' => '1.00',
-            'donationStatus' => 'Collected',
+            'donationStatus' => 'Refunded',
             'charityComms' => NULL,
             'giftAid' => NULL,
             'tbgComms' => NULL,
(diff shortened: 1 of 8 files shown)


src/Domain/Donation.php
```diff
@@ -390,6 +402,7 @@ public function toHookModel(): array
         }
 
         $data['updatedTime'] = $this->getUpdatedDate()->format(DateTimeInterface::ATOM);
+        $data['refundedTime'] = $this->refundedAt?->format(DateTimeInterface::ATOM);
```
bdsl, line 405: We could just comment out this line until SF is ready to recieve the refund timestamps so that this can be merged, and will start recording donation times to the MySQL db.

src/Application/Actions/Hooks/StripeChargeUpdate.php
```diff
@@ -288,7 +288,7 @@ private function handleChargeRefunded(Event $event, Response $response): Respons
                 $donation->getUuid(),
                 $charge->id,
             ));
-            $donation->setDonationStatus(DonationStatus::Refunded);
+            $donation->recordRefundAt(new \DateTimeImmutable());
```
bdsl, line 291: We probably could take all the timestamps above from stripe, but I'm not sure if there's any particular advantage for that over just taking them from the server clock. Depends how precise we need the date to be. @DomTBG do you have a view?

DomTBG:
Thanks @bdsl The datetimes need to match Stripe, as we do reconciliations between Salesforce and Stripe to check the accuracy of the date. This might not seem so important, except for the fact that we upload the day's transactions as one figure into QuickBooks, so it's important that anything that happens at 23:59:59 on 30 June is recognised in Salesforce as also being on 30 June and not at 00:00:01 on 1 July.

bdsl:
Thanks, I will take the tiemstamp from Stripe then and also pay attention to timezones.

bdsl:
@DomTBG I've updated this to use the time stripe sends us for the refund **event**.

src/Application/Actions/Hooks/StripeChargeUpdate.php
```diff
@@ -10,6 +10,7 @@
 use MatchBot\Domain\Donation;
 use MatchBot\Domain\DonationRepository;
 use MatchBot\Domain\DonationStatus;
+use Monolog\DateTimeImmutable;
```
dee-jo, line 13:
```suggestion
```

dee-jo (approved):

integrationTests/DonationPersistenceTest.php
```diff
@@ -39,7 +40,7 @@ public function testItLoadsADonationToDB(): void
         $donation = $donationRepo->findOneBy(['uuid' => $donationRow['uuid']]);
 
         $this->assertNotNull($donation);
-        $this->assertSame(DonationStatus::Refunded, $donation->getDonationStatus());
+        $this->assertEquals(DonationStatus::Refunded, $donation->getDonationStatus());
```
bdsl, line 43: just changed this because PHPUnit gives a much more useful failure message with assertEquals on an enum than assertSame. If assertSame failed it just says two objects are not identical and doesn't elaborate. if assertEquals fails we actually get to see what the expected and actual enum are.

dee-jo (approved):
lgtm

PR #570, opened 2023-07-04
BG2-2306 – set `refundedTime` with outbound Donation webhooks when applicable

NoelLH:


dee-jo (approved):
LGTM, and tests passing: Screenshot from 2023-07-05 14-12-03

PR #577, opened 2023-07-20
DON-706: Allow setting donor name and email on new donations

dorota-joanna:
Currently the donation has to be created without these details and then they can be added later in a PUT request.

### DIFF
--- a/src/Application/HttpModels/DonationCreate.php
+++ b/src/Application/HttpModels/DonationCreate.php
@@ -32,9 +32,9 @@ class DonationCreate
         public ?bool $optInTbgEmail = null,
         public ?string $pspCustomerId = null,
         public ?string $tipAmount = '0.00',
-        public ?string $donorFirstName = null,
-        public ?string $donorLastName = null,
-        public ?string $donorEmail = null
+        public ?string $firstName = null,
+        public ?string $lastName = null,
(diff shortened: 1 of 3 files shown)


src/Application/Actions/Donations/Get.php
```diff
@@ -8,6 +8,7 @@
 use MatchBot\Application\Actions\Action;
 use MatchBot\Domain\DomainException\DomainRecordNotFoundException;
 use MatchBot\Domain\Donation;
+use MatchBot\Domain\DonationStatus;
```
bdsl, line 11:
```suggestion
```

src/Application/Actions/Donations/Get.php
```diff
@@ -39,7 +40,7 @@ protected function action(Request $request, Response $response, array $args): Re
         if (!$donation) {
             throw new DomainRecordNotFoundException('Donation not found');
         }
-
+        $donation->setDonationStatus(DonationStatus::Paid);
```
bdsl, line 43:
```suggestion
```

These lines must have been committed by mistake. I tend to do the same on my local when I'm testing frontend stuff to make the thank you page load. But probably we should have a unit test that will fail when these lines are here - I might add one.

NoelLH (approved):

PR #581, opened 2023-08-01
MAT-308: Resubmit gift aid claims

bdsl:


src/Migrations/Version20230801110114.php
```diff
@@ -0,0 +1,43 @@
...
+                tbgGiftAidRequestConfirmedCompleteAt = NULL,
+                tbgGiftAidRequestCorrelationId = NULL,
+                tbgGiftAidResponseDetail = NULL
+            WHERE
+                campaign_id = 4644 AND
+                giftAid = 1 AND
+                tbgShouldProcessGiftAid = 1 AND
+                tbgGiftAidRequestCorrelationId IS NOT NULL AND
+                donationStatus = 'Paid'
+            LIMIT 224;
```
NoelLH, line 34: @bdsl What about the limit vs. eligible donations on this campaign?

src/Migrations/Version20230801110114.php
```diff
@@ -0,0 +1,43 @@
...
+namespace MatchBot\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20230801110114 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Resumbit Gift Aid claims (MAT-308)';
```
NoelLH, line 14:
```suggestion
        return 'Resubmit Gift Aid claims (MAT-308)';
```

src/Migrations/Version20230801110114.php
bdsl: Ah right that's copied over wrongly. Looks like there are 97 rows in the DB that match this where clause.
```suggestion
            LIMIT 97;
```

NoelLH (approved):

PR #583, opened 2023-08-03
MAT-309: detect stripe initiated payment intent cancellation

bdsl:


src/Application/Actions/Hooks/StripePayoutUpdate.php
```diff
@@ -7,11 +7,15 @@
 use MatchBot\Application\Actions\ActionPayload;
 use MatchBot\Application\Messenger\StripePayout;
 use MatchBot\Application\Notifier\StripeChatterInterface;
+use MatchBot\Domain\DonationRepository;
+use MatchBot\Domain\DonationStatus;
```
bdsl, line 11:
```suggestion
```

src/Application/Actions/Hooks/StripePayoutUpdate.php
```diff
@@ -7,11 +7,15 @@
...
 use Psr\Container\ContainerInterface;
 use Psr\Http\Message\ResponseInterface as Response;
 use Psr\Http\Message\ServerRequestInterface as Request;
 use Psr\Log\LoggerInterface;
+use Stripe\Charge;
 use Stripe\Event;
+use Stripe\PaymentIntent;
```
bdsl, line 18:
```suggestion
use Stripe\Event;
```

src/Application/Actions/Hooks/StripePayoutUpdate.php
```diff
@@ -130,4 +134,5 @@ private function handlePayoutPaid(Request $request, Event $event, Response $resp
 
         return $this->respondWithData($response, $event->data->object);
     }
+
```
bdsl, line 137:
```suggestion
```

app/routes.php
```diff
@@ -63,7 +63,7 @@ function (RouteCollectorProxy $paymentMethodsGroup) {
     });
 
     // Authenticated through Stripe's SDK signature verification
-    $app->post('/hooks/stripe', Hooks\StripeChargeUpdate::class);
+    $app->post('/hooks/stripe', Hooks\StripePaymentsUpdate::class);
```
bdsl, line 66: renamed the class as we're now going to be taking events related to both charges and payment intents.

src/Application/Actions/Hooks/StripePaymentsUpdate.php
```diff
@@ -30,7 +31,7 @@
  *
  * @return Response
  */
-class StripeChargeUpdate extends Stripe
+class StripePaymentsUpdate extends Stripe
```
NoelLH, line 34: It might be helpful to have a class comment here which separately lists the two Stripe objects involved / listened for? If I was new to Stripe I think I might assume `Payment` was a first class object in their domain based on this class name.

In any case, we should add to the list of specific events above.

src/Application/Actions/Donations/Update.php
```diff
@@ -142,11 +142,37 @@ protected function action(Request $request, Response $response, array $args): Re
                     $retryCount,
                 ));
 
-                \usleep(100_000 * (2 ** $retryCount)); // pause for 0.1, 0.2, 0.4 and then 0.8s before giving up.
+                $microseconds = (int)(100_000 * (2 ** $retryCount)); // pause for 0.1, 0.2, 0.4 and then 0.8s before giving up.
+                \assert($microseconds >= 0);
+                \usleep($microseconds);
```
bdsl, line 147: I think Psalm isn't clever enough to know that $microseconds is always positive without us asserting it, there was something in the baseline but because of a mistake I made psalm stopped looking at that and I removed it from the baseline. May as well leave this with the assert now.

src/Application/Actions/Donations/Update.php
```diff
@@ -142,11 +142,37 @@ protected function action(Request $request, Response $response, array $args): Re
...
+                        $args['donationId'],
+                    ));
+                    $this->entityManager->rollback();
+
+                    return $this->validationError(
+                        $response,
+                        "Donation ID {$args['donationId']} could not be updated",
+                        'This donation payment intent has been cancelled. You may wish to start a fresh donation.'
+                    );
+                }
```
bdsl, line 171: I don't think this really needs tests

src/Application/Actions/Hooks/StripePaymentsUpdate.php
```diff
@@ -315,6 +318,33 @@ private function handleChargeRefunded(Event $event, Response $response): Respons
         return $this->respondWithData($response, $charge);
     }
 
+    private function handlePaymentIntentCancelled(Request $request, Event $event, Response $response): Response
```
NoelLH, line 321: I think we should test receiving this hook. Since it was previously used to check for unsupported actions, we already have a likely-suitable simulated payload ready in `pi_canceled.json`

src/Application/Actions/Hooks/StripePaymentsUpdate.php
```diff
@@ -315,6 +318,33 @@ private function handleChargeRefunded(Event $event, Response $response): Respons
...
+        $donation = $this->donationRepository->findOneBy(['transactionId' => $paymentIntent->id]);
+
+        if ($donation === null) {
+            return $this->respond($response, new ActionPayload(404));
+        }
+
+        if ($donation->getDonationStatus() !== DonationStatus::Cancelled) {
+            $this->logger->info(sprintf(
+                'Received Stripe cancellation request, Donation ID %s, payment intent ID %s',
+                $donation->getId() ?? throw new \Exception("Missing ID on donation"),
```
NoelLH, line 334: Including donation UUID in all logs – instead of or in addition to numeric ID – makes it much easier to search all relevant activity in CloudWatch Logs Insights

src/Application/Actions/Donations/Update.php
```diff
@@ -142,11 +142,37 @@ protected function action(Request $request, Response $response, array $args): Re
...
+                if (
+                    str_starts_with(
+                        haystack: $invalidRequestException->getMessage(),
+                        needle: "This PaymentIntent's amount could not be updated because it has a status of canceled",
+                    )
+                ) {
+                    $this->logger->warning(sprintf(
+                        'Stripe rejected payment intent update as PI was cancelled, presumably by stripe' .
+                        ' itself very recently. Donation # %s',
+                        $args['donationId'],
```
NoelLH, line 162: As in the hook receiver, it's good to have UUID as well/instead

integrationTests/IntegrationTest.php
```diff
@@ -4,17 +4,51 @@
...
+        $container->set(RateLimitMiddleware::class, $noOpMiddlware);
+        $container->set(\Psr\Log\LoggerInterface::class, new \Psr\Log\NullLogger());
+
+        AppFactory::setContainer($container);
+        $app = AppFactory::create();
+
+        $routes = require __DIR__ . '/../app/routes.php';
+        $routes($app);
+
+        self::setApp($app);
```
bdsl, line 49: moved this from the bootstrap to the setup because it seemed like some stuff was persisting between tests causing a duplicate db key error

src/Application/Actions/Donations/Update.php
```diff
@@ -97,15 +97,14 @@ protected function action(Request $request, Response $response, array $args): Re
         while ($retryCount < self::MAX_UPDATE_RETRY_COUNT) {
             $this->entityManager->beginTransaction();
 
-            try {
-                $donation = $this->donationRepository->findAndLockOneBy(['uuid' => $args['donationId']]);
-
-                if (!$donation) {
-                    $this->entityManager->rollback();
+            $donation = $this->donationRepository->findAndLockOneBy(['uuid' => $args['donationId']]);
```
bdsl, line 100: moved this outside the following try/catch, because otherwise Psalm points out that $donation is possibly undefined when we try to use it in the catch block to the the uuid.

src/Application/Actions/Donations/Update.php
bdsl: Thanks, changed.

src/Application/Actions/Hooks/StripePaymentsUpdate.php
bdsl: Thanks, done.

NoelLH (approved):
Looks good!

PR #586, opened 2023-08-30
DON-855: Allow setting up payment intent suitable for use with new st…

bdsl:
…ripe payment element

This will have to be deployed to prod *before* the change in frontend.

### DIFF
--- a/src/Application/Actions/Donations/Create.php
+++ b/src/Application/Actions/Donations/Create.php
@@ -56,6 +56,9 @@ class Create extends Action
 
         $body = (string) $request->getBody();
 
+        // Frontend just has to send this temporarily so we know when its switched over from the card element to the new payment element.
+        $forNewPaymentElement = ($request->getQueryParams()['forNewPaymentElement'] ?? '') === 'true';
+
         try {
             /** @var DonationCreate $donationData */
             $donationData = $this->serializer->deserialize($body, DonationCreate::class, 'json');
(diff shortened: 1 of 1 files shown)


src/Application/Actions/Donations/Create.php
```diff
@@ -147,8 +150,20 @@ protected function action(Request $request, Response $response, array $args): Re
...
+            if ($forNewPaymentElement) {
+                // We don't need to set specific payment methods here, they are set on the client side by the payment
+                // element.
+                $stripeMethodProperties = [
+                    'automatic_payment_methods' => [
+                    'enabled' => true,
+                    'allow_redirects' => 'always',
+                    ]
+                ];
+            } else {
```
bdsl, line 162: but note that 153 and 163 do continue to be covered by the existing tests. If 153 was the wrong way round 6 tests would error.

dorota-joanna (approved):
LGTM

src/Application/Actions/Donations/Create.php
```diff
@@ -147,8 +150,20 @@ protected function action(Request $request, Response $response, array $args): Re
                 $donation->setCampaign($campaign);
             }
 
+            if ($forNewPaymentElement) {
+                // We don't need to set specific payment methods here, they are set on the client side by the payment
```
NoelLH, line 154: @bdsl This doesn't feel like the best approach to me – what if somebody uses modified frontend code to try out methods we don't expect against our backend? I think we should support just the methods we expect.

src/Application/Actions/Donations/Create.php
```diff
@@ -147,8 +150,20 @@ protected function action(Request $request, Response $response, array $args): Re
...
+                // element.
+                $stripeMethodProperties = [
+                    'automatic_payment_methods' => [
+                    'enabled' => true,
+                    'allow_redirects' => 'always',
```
NoelLH, line 159: @bdsl Does `never` cause any problems here, if our account has no redirect based methods switched on?

bdsl: I'd have to test it to know for sure. But the frontend is configured in the open PR to only do redirects if required, so I don't think there's any harm I can see from allowing them. I've set it to redirect to the thank you page. Is there any advantage to never? Even if does work now it would mean any payment method we try to add that requires a redirect would get filtered out. https://stripe.com/docs/api/errors#errors-setup_intent-automatic_payment_methods-allow_redirects

src/Application/Actions/Donations/Create.php
NoelLH: Cool – presumably we could still do the reverse and allow auto methods everywhere (along with making sure those we don't want are off in the account) @bdsl?

bdsl: See https://github.com/thebiggive/matchbot/pull/588

src/Application/Actions/Donations/Create.php
bdsl: See https://github.com/thebiggive/matchbot/pull/588

PR #587, opened 2023-09-01
MAT-314: Stop logging error on update payment method exception

bdsl:


NoelLH (commented):
The caught exception class is quite general – is it possible we're going to lose visibility on other problems that should remain error level?

bdsl:
@NoelLH Yes and it's an abstract class. We don't have a log which concrete exception was thrown this week, but I'd guess it was `\Stripe\Exception\CardException`. 

I've pushed '[2bfba24](https://github.com/thebiggive/matchbot/pull/587/commits/2bfba24d05578c5f8fab5a8334d468b2d16d52eb)' so we only log 'info' if it's the message we saw in the logs, for any other message we will log an error.

Also logging the class of the exception.

I don't think it's a big issue, the user will see the exception message and can create a case if they want to.

NoelLH (approved):

PR #588, opened 2023-09-01
Don 855 ii

bdsl:
Adjustments to matchbot prep for new Stripe Payment element as discussed with @NoelLH .

### DIFF
--- a/src/Application/Actions/Donations/Create.php
+++ b/src/Application/Actions/Donations/Create.php
@@ -156,7 +156,7 @@ class Create extends Action
                 $stripeMethodProperties = [
                     'automatic_payment_methods' => [
                     'enabled' => true,
-                    'allow_redirects' => 'always',
+                    'allow_redirects' => 'never',
                     ]
                 ];
             } else {
--- a/src/Application/Actions/Donations/Create.php
+++ b/src/Application/Actions/Donations/Create.php
@@ -56,9 +56,6 @@ class Create extends Action
 
         $body = (string) $request->getBody();
 
-        // Frontend just has to send this temporarily so we know when its switched over from the card element to the new payment element.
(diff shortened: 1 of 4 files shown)


src/Application/Actions/Donations/Create.php
```diff
@@ -56,9 +56,6 @@ protected function action(Request $request, Response $response, array $args): Re
 
         $body = (string) $request->getBody();
 
-        // Frontend just has to send this temporarily so we know when its switched over from the card element to the new payment element.
-        $forNewPaymentElement = ($request->getQueryParams()['forNewPaymentElement'] ?? '') === 'true';
```
bdsl, line 60: I've tested with the current and new versions of the frontend - the current version of frontend works with the new style of payment intents, the new one doesn't work with the old style. So we can just always generate new style ones in matchbot and it will be compatible with both.

src/Application/Actions/Donations/Create.php
```diff
@@ -150,20 +147,11 @@ protected function action(Request $request, Response $response, array $args): Re
...
-                    'allow_redirects' => 'always',
-                    ]
-                ];
-            } else {
-                $stripeMethodProperties = $donation->getStripeMethodProperties();
-            }
             $createPayload = [
-                ...$stripeMethodProperties,
+                'automatic_payment_methods' => [
+                    'enabled' => true,
```
bdsl, line 152: The new payment option doesn't give us an option to do anything other than have automatic methods on the frontend, and payment confirmation will then fail if we use it and we don't have automatic methods set in the intent created here. See https://stripe.com/docs/js/elements_object/create_payment_element or it's maybe clearer in the typescript: https://github.com/stripe/stripe-js/blob/cac790b90d6448dccc61fb9ba5299b48fcad4409/types/stripe-js/elements/payment.d.ts#L199

NoelLH (approved):

PR #590, opened 2023-09-06
MAT-316: Prevent donations with negative tip

bdsl:
We've never seen a donation with a negative tip, and the front end doesn't allow it, but best to block it just in case anyone tries or somehow creates one by mistake.


dorota-joanna (approved):
LGTM!

PR #598, opened 2023-09-20
DON-893: Allow confirming payment methods on server side from matchbot

bdsl:
Introduces a new route at `/donations/{donationId:[a-z0-9-]{36}}/confirm`. Frontend should post to this to indicate when a donor is ready to pay, at which point matchbot will pass the message on to Stripe.

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -43,10 +43,9 @@ return function (App $app) {
             ->add(DonationPublicAuthMiddleware::class);
 
         $versionGroup->post('/donations/{donationId:[a-z0-9-]{36}}/confirm', Donations\Confirm::class)
-            // todo - add needed middlewares. Not in right now as interferes with testing.
-//            ->add(DonationRecaptchaMiddleware::class) // Runs last
-//            ->add($ipMiddleware)
-//            ->add(RateLimitMiddleware::class)
+            ->add(DonationRecaptchaMiddleware::class) // Runs last
+            ->add($ipMiddleware)
+            ->add(RateLimitMiddleware::class)
         ;
 
         $versionGroup->post('/people/{personId:[a-z0-9-]{36}}/donations', Donations\Create::class)
--- a/app/routes.php
+++ b/app/routes.php
@@ -43,7 +43,7 @@ return function (App $app) {
             ->add(DonationPublicAuthMiddleware::class);
 
         $versionGroup->post('/donations/{donationId:[a-z0-9-]{36}}/confirm', Donations\Confirm::class)
-            ->add(DonationPublicAuthMiddleware::class);;
+            ->add(DonationPublicAuthMiddleware::class);
 
         $versionGroup->post('/people/{personId:[a-z0-9-]{36}}/donations', Donations\Create::class)
             ->add(PersonManagementAuthMiddleware::class)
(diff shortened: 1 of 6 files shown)


psalm.xml
```diff
@@ -6,7 +6,7 @@
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
     errorBaseline="psalm-baseline.xml"
-    findUnusedBaselineEntry="true"
+    findUnusedBaselineEntry="false"
```
bdsl, line 9: this gets in the way to much I think for little benefit, especially when trying to Psalm to check one file at a time.

app/routes.php
```diff
@@ -42,6 +42,9 @@
         })
             ->add(DonationPublicAuthMiddleware::class);
 
+        $versionGroup->post('/donations/{donationId:[a-z0-9-]{36}}/confirm', Donations\Confirm::class)
+            ->add(DonationPublicAuthMiddleware::class);;
```
NoelLH, line 46:
```suggestion
            ->add(DonationPublicAuthMiddleware::class);
```

src/Application/Actions/Donations/Confirm.php
```diff
@@ -0,0 +1,106 @@
...
+        ]);
+
+        $updatedIntent = $this->stripeClient->paymentIntents->retrieve($paymentIntentId);
+
+        $this->entityManager->flush();
+
+        return new JsonResponse([
+                'paymentIntent' => [
+                    'status' => $updatedIntent->status,
+                    'next_action' => $updatedIntent->next_action
```
NoelLH, line 102: I think based on the docs that we need to include the secret here, and probably don't strictly need the `next_action` name.

```suggestion
                    'client_secret' => $updatedIntent->client_secret,
```

NoelLH (commented):
Looking good! I've added a couple of things inline.

src/Application/Actions/Donations/Confirm.php
bdsl: Right, I thought I remembered putting next_action in because the frontend used that, but looking again it seems like it only uses status: https://github.com/thebiggive/donate-frontend/blob/55f830d9018f82a583ff56947e06e8ef93094b13/src/app/donation-start/donation-start-form/donation-start-form.component.ts#L810

What were you thinking we'd need to do with the secret on the frontend?

Of course we could just return `['paymentIntent' => $updatedIntent->toArray()]` but I feel like we should probably be intentional about which fields we send on the the client, and have them listed in the source so we can find them instead of having to look up stripe's docs.

bdsl: OK looks like on the client we basically need to copy stripe's code: 

```js
  } else if (response.status === "requires_action") {
    // Use Stripe.js to handle the required next action
    const {
      error,
      paymentIntent
    } = await stripe.handleNextAction({
      clientSecret: response.clientSecret
    });
```

Done in 06e87ca

NoelLH: Yup, thanks!

NoelLH (approved):

PR #599, opened 2023-09-25
MAT-317: Remove Donation.clientSecret property

bdsl:
Not needed in future when donations are always confirmed from matchbot not from the client

** DO NOT MERGE UNTIL AFTER https://github.com/thebiggive/donate-frontend/pull/1302 IS DEPLOYED TO PROD **

### DIFF
--- a/integrationTests/DonationPersistenceTest.php
+++ b/integrationTests/DonationPersistenceTest.php
@@ -77,7 +77,6 @@ class DonationPersistenceTest extends IntegrationTest
             'salesforceId' => NULL,
             'tipAmount' => '0.00',
             'psp' => 'stripe',
-            'clientSecret' => NULL,
             'donorHomeAddressLine1' => NULL,
             'donorHomePostcode' => NULL,
             'tipGiftAid' => NULL,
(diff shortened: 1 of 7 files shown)


NoelLH (commented):
Looks like Psalm's still unhappy but the basic idea looks good

bdsl:
> Looks like Psalm's still unhappy but the basic idea looks good

Thanks, fixed the Psalm issue.

NoelLH (approved):
Looks good!

I don't think we made the full scope of the tidying explicit in Jira but perhaps we should also at this point stop calling `$this->donationRepository->deriveFees()` in the `Update` action? Unless that's going to cause more trouble with Donation Funds when we use `autoConfirmFromCashBalance` in an Update.

bdsl:
> perhaps we should also at this point stop calling `$this->donationRepository->deriveFees()` in the `Update` action? Unless that's going to cause more trouble with Donation Funds when we use `autoConfirmFromCashBalance` in an Update.

Yes probably. I've copied your question and this answer to Jira.

PR #601, opened 2023-09-26
MAT-317: Remove never-set card detail properties from donation http m…

bdsl:
…odel

### DIFF
--- a/src/Application/Actions/Donations/Confirm.php
+++ b/src/Application/Actions/Donations/Confirm.php
@@ -7,6 +7,7 @@ use http\Exception\BadMethodCallException;
 use Laminas\Diactoros\Response\EmptyResponse;
 use Laminas\Diactoros\Response\JsonResponse;
 use MatchBot\Application\Actions\Action;
+use MatchBot\Application\Fees\Calculator;
 use MatchBot\Client\NotFoundException;
 use MatchBot\Domain\DonationRepository;
 use Psr\Http\Message\ResponseInterface as Response;
@@ -68,6 +69,9 @@ class Confirm extends Action
         // two letter upper string, e.g. 'GB', 'US'.
         $cardCountry = $paymentMethod->card->country;
         \assert(is_string($cardCountry));
+        if (! in_array($cardBrand, Calculator::STRIPE_CARD_BRANDS, true)) {
+            throw new HttpBadRequestException($request, "Unrecognised card brand");
+        }
 
         // at present if the following line was left out we would charge a wrong fee in some cases. I'm not happy with
         // that, would like to find a way to make it so if its left out we get an error instead - either by having
--- a/src/Application/Actions/Donations/Update.php
+++ b/src/Application/Actions/Donations/Update.php
@@ -274,11 +274,9 @@ class Update extends Action
         $donation->setChampionComms($donationData->optInChampionEmail);
         $donation->setDonorBillingAddress($donationData->billingPostalAddress);
 
-        $this->donationRepository->deriveFees(
-            $donation,
-            $donationData->cardBrand,
-            $donationData->cardCountry
-        );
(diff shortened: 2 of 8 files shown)


src/Application/Actions/Donations/Confirm.php
```diff
@@ -68,6 +69,9 @@ protected function action(Request $request, Response $response, array $args): Re
         // two letter upper string, e.g. 'GB', 'US'.
         $cardCountry = $paymentMethod->card->country;
         \assert(is_string($cardCountry));
+        if (! in_array($cardBrand, Calculator::STRIPE_CARD_BRANDS, true)) {
+            throw new HttpBadRequestException($request, "Unrecognised card brand");
```
bdsl, line 73: the docblock change on `deriveFees` forces this check here.

src/Application/Actions/Donations/Update.php
```diff
@@ -274,11 +274,9 @@ private function addData(Donation $donation, HttpModels\Donation $donationData,
         $donation->setChampionComms($donationData->optInChampionEmail);
         $donation->setDonorBillingAddress($donationData->billingPostalAddress);
 
-        $this->donationRepository->deriveFees(
-            $donation,
-            $donationData->cardBrand,
-            $donationData->cardCountry
-        );
+        // currently this can't change the fee from what it was when donation entity was created, but
+        // we leave it here for consistency with the Create action.
```
bdsl, line 278:
```suggestion
        // we call deriveFees here for consistency with the Create action, in case the derivation logic changes to depend on something we do mutate in the donation.
```

src/Domain/DonationRepository.php
```diff
@@ -544,7 +544,15 @@ public function abandonOldCancelled(): int
...
-    public function deriveFees(Donation $donation, ?string $cardBrand = null, ?string $cardCountry = null): void
+    /**
+     * Updates a donation to set the appropriate fees. If card details are null then we assume for now that a card with
+     * the lowest possible fees will be used, and this should be called again with the details of the selected card
+     * when confirming the payment.
+     *
+     * @psalm-param value-of<Calculator::STRIPE_CARD_BRANDS>|null $cardBrand
+     * @param string|null $cardCountry ISO two letter uppercase code
+     */
+    public function deriveFees(Donation $donation, ?string $cardBrand, ?string $cardCountry): void
```
bdsl, line 555: I thought it worth removing the default nulls so that we're forced to look at the nulls and think about them when we call this.

src/Application/HttpModels/Donation.php
```diff
@@ -34,6 +34,4 @@ class Donation
     public string $projectId;
     public ?float $tipAmount = null;
     public ?bool $tipGiftAid = null;
-    public ?string $cardBrand = null;
-    public ?string $cardCountry = null;
```
bdsl, line 38: @NoelLH while I'm here do you think I should remove `$donationMatched` and `$projectId`? PhpStorm says they're unused, I'm not sure what they were fore.

bdsl: These properties are already removed from the client side model in https://github.com/thebiggive/donate-frontend/commit/5dbf0aa5ced which is in prod.

NoelLH: I'm struggling to quickly figure this one out. To be clear, `donationMatched` and `projectId` are _not_ removed from the general client side Donate model and are used there, but it does seem correct that in this specific PHP model they're not sent. (They are in `DonationCreate` though.) I'm not sure if they were ever needed or if we got 2 similar models conflated.

NoelLH (approved):

src/Application/HttpModels/Donation.php
bdsl: OK, lets leave them here for now.

PR #602, opened 2023-09-27
MAT-313: Don't generate 404 error when stripe tries to cancel non-exi…

bdsl:
…stent donation in non-prod environment


dorota-joanna (approved):
LGTM

PR #603, opened 2023-09-28
MAT-318, MAT-319 – distinguish between Gift Aid onboarding & agent approval

NoelLH:


integrationTests/DonationRepositoryTest.php
```diff
@@ -0,0 +1,92 @@
...
+use MatchBot\Domain\DonationRepository;
+use MatchBot\Domain\DonationStatus;
+use MatchBot\Domain\PaymentMethodType;
+
+class DonationRepositoryTest extends IntegrationTest
+{
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->clearPreviousCampaignsCharitiesAndRelated(); // Avoid e.g. HMRC ref dupes
```
NoelLH, line 19: It feels a bit odd to do this specifically for _this_ repo test. But I also feel like because the fixture charity statuses here are very specific to what this class/method test is targeting, it's easier to understand if the test data is set up in this class too.

Very open to any ideas on how to get the best of both approaches.

src/Domain/CampaignRepository.php
```diff
@@ -110,10 +117,16 @@ private function pullCharity(
...
-            in_array($giftAidOnboardingStatus, static::$giftAidOnboardedStatuses, true)
+            in_array($giftAidOnboardingStatus, self::GIFT_AID_ONBOARDED_STATUSES, true)
+        );
+        $tbgApprovedToClaimGiftAid = (
+            !empty($hmrcReferenceNumber) &&
+            in_array($giftAidOnboardingStatus, self::GIFT_AID_APPROVED_STATUSES, true)
         );
 
         $charity->setTbgClaimingGiftAid($tbgCanClaimGiftAid);
+        $charity->setTbgApprovedToClaimGiftAid($tbgApprovedToClaimGiftAid);
```
bdsl, line 128: Maybe at this point it's worth moving this code into the Charity class - something like https://github.com/thebiggive/matchbot/commit/11db11a63b36f6a8637440ab81874a5a7f876b0a . That would allow making several of the setters private or removing them.

Or would it make sense now that we pull quite a lot from SF to convert CharityRepository into a SalesforceReadyProxyRepository? Maybe the reasoning at https://github.com/thebiggive/matchbot/blob/1dd8998c64f0c81d4a25c80694cb1062a579edb9/src/Domain/CharityRepository.php#L10 is out of date.

src/Domain/Charity.php
```diff
@@ -7,6 +7,9 @@
 use Doctrine\ORM\Mapping as ORM;
 
 /**
+ * @psalm-api Work around Psalm being unhappy with the property or method being unused
```
NoelLH, line 10: This doesn't feel like the most accurate way to make Psalm happy about `$tbgApprovedToClaimGiftAid`. The real reason it is declared and not used in PHP is that it's used via DQL / queries built into DQL instead.

(If we add a getter, it just switches to complaining about that being unused.)

bdsl: I'd add a suppression on the specific property, instead of at the class level:

```php
    /**
     * @psalm-suppress PossiblyUnusedProperty - used in a database query
     * @ORM\Column(type="boolean")
     * @var bool    Whether the charity's Gift Aid may NOW be claimed by the Big Give according to HMRC.
     */
    protected bool $tbgApprovedToClaimGiftAid = false;
    ```

File: src/Domain/Charity.php [hunk 4]
```diff
@@ -57,10 +60,18 @@ class Charity extends SalesforceReadProxy
...
+     *              indicates we should charge a fee and plan to claim, but not that we are necessarily
+     *              set up as an approved Agent yet.
      */
     protected bool $tbgClaimingGiftAid = false;
 
+    /**
+     * @ORM\Column(type="boolean")
+     * @var bool    Whether the charity's Gift Aid may NOW be claimed by the Big Give according to HMRC.
+     */
+    protected bool $tbgApprovedToClaimGiftAid = false;
```
bdsl, line 73: Can we do private unless we have a use case for protected?

Separately, I'm wondering whether this should be a required constructor param, so that we know we always have to know whether we're approved or not. We only actually instantiate Charity once in prod code, but it might also be useful to have it as a required constructor param so we're prompted to pass it for tests and we make a choice of which scenario to test instead of just testing the false scenario unless we think of setting it to true.

integrationTests/DonationRepositoryTest.php
bdsl: What if we randomise the ref and anything else that could conflict? Maybe then we wouldn't need to delete anything from the DB.

I think we don't have a separate database for manual testing and automated testing in dev environments, so clearing this stuff might cause problems for manual testing.

src/Domain/Donation.php
```diff
@@ -63,9 +63,9 @@ class Donation extends SalesforceWriteProxy
 
     /**
      * @ORM\Column(type="datetime", nullable=true)
-     * @var ?DateTime    When the donation first moved to status Collected, i.e. the donor finished paying.
+     * @var ?DateTimeInterface  When the donation first moved to status Collected, i.e. the donor finished paying.
      */
-    protected ?DateTime $collectedAt = null;
+    protected ?DateTimeInterface $collectedAt = null;
```
bdsl, line 68: If we're moving away from DateTime could we type it as `DateTimeImmutable`? I think the existence of that interface is a mistake, (and the author of it agrees) since the two classes are not really interchangable.

Would require changing the ORM mapping to `type="datetime_immutable"` but shouldn't need any DB migration.

src/Domain/Charity.php
```diff
@@ -93,17 +104,11 @@ public function setStripeAccountId(?string $stripeAccountId): void
         $this->stripeAccountId = $stripeAccountId;
     }
 
-    /**
-     * @return bool
-     */
```
bdsl, line 98: :+1:

src/Domain/Donation.php
```diff
@@ -316,16 +316,24 @@ private function __construct(string $amount, string $currencyCode, PaymentMethod
...
+            $donationData->paymentMethodType,
+        );
 
         $donation->setPsp($psp);
         $donation->setUuid(Uuid::uuid4());
         $donation->setCampaign($campaign); // Charity & match expectation determined implicitly from this
+
         $donation->setGiftAid($donationData->giftAid);
+        // `DonationCreate` doesn't support a distinct property yet & we only ask once about GA.
+        $donation->setTipGiftAid($donationData->giftAid);
```
bdsl, line 334: I see previously this was only set on `update`, not on donation create, and we're not calling update so much now. Is this fixing a separate bug that resulted from that oversight? 

Also I'm a bit confused about why it's a nullable property - are there clear semantics for the three values, or should we make it a non-null bool?

src/Domain/Donation.php
```diff
@@ -316,16 +316,24 @@ private function __construct(string $amount, string $currencyCode, PaymentMethod
...
+        $donation->setTbgShouldProcessGiftAid($donation->getCampaign()->getCharity()->isTbgClaimingGiftAid());
```
bdsl, line 335: We already have the campaign, don't need to pull it of the donation:

```suggestion
        $donation->setTbgShouldProcessGiftAid($campaign->getCharity()->isTbgClaimingGiftAid());
```

src/Domain/Donation.php
```diff
@@ -478,12 +486,12 @@ public function setDonationStatus(DonationStatus $donationStatus): void
         $this->donationStatus = $donationStatus;
     }
 
-    public function getCollectedAt(): ?DateTime
+    public function getCollectedAt(): ?DateTimeInterface
```
bdsl, line 489: or `?DateTimeImmutable` ?

src/Domain/Donation.php
```diff
@@ -478,12 +486,12 @@ public function setDonationStatus(DonationStatus $donationStatus): void
...
     {
         return $this->collectedAt;
     }
 
-    public function setCollectedAt(?DateTime $collectedAt): void
+    public function setCollectedAt(?DateTimeInterface $collectedAt): void
```
bdsl, line 494: as above

tests/TestCase.php
```diff
@@ -167,4 +169,13 @@ protected function getTestIdentityTokenIncomplete(): string
...
     }
+
+    protected function getMinimalCampaign(): Campaign
+    {
+        $charity = new Charity();
+        $campaign = new Campaign();
+        $campaign->setCharity($charity);
+
+        return $campaign;
+    }
```
bdsl, line 180: If this is the minimal campaign maybe worth doing this in the campaign class? Maybe give Campaign a required but nullable $charity param? 

Ultimately I think we should do something like make the `$proxy` param of `\MatchBot\Domain\SalesforceReadProxyRepository::pull` nullable, or split that into two separate methods, one to pull a new entity and one to update existing, so that we don't need the ability to create a less-than-minimal entity. For pulling new we don't need a whole entity, we just need an SF ID.

tests/Domain/DonationRepositoryTest.php
```diff
@@ -523,16 +521,16 @@ public function testFindReadyToClaimGiftAid(): void
         $queryBuilderProphecy->where('d.donationStatus = :claimGiftAidWithStatus')
             ->shouldBeCalledOnce()->willReturn($queryBuilderProphecy->reveal());
 
-        // 5 `andWhere()`s in all, excluding the first `where()` but including the one
+        // 6 `andWhere()`s in all, excluding the first `where()` but including the one
         // NOT for `$withResends` calls.
         $queryBuilderProphecy->andWhere(Argument::type('string'))
-            ->shouldBeCalledTimes(5)->willReturn($queryBuilderProphecy->reveal());
+            ->shouldBeCalledTimes(6)->willReturn($queryBuilderProphecy->reveal());
```
bdsl, line 527: Is it worthwhile to be counting the `andWheres` etc? I'm not sure how much it really helps us understand what the code's doing or avoid bugs. Maybe worth deleting this test and just testing the DonationRepository through in integration with the DB? This feels more like coding in duplicate than coding and testing.

tests/Domain/DonationTest.php
```diff
@@ -88,7 +89,7 @@ public function test25k1CardIsTooHigh(): void
             projectId: "any project",
             psp:'stripe',
             paymentMethodType: PaymentMethodType::Card
-        ), new Campaign());
+        ), $this->getMinimalCampaign());
```
bdsl, line 92: see comment on getMinimalCampaign function

bdsl (commented):
Reviewed, all LGTM other than where I've commented. Happy to chat or pair if you like to make any adjustments.

src/Domain/CampaignRepository.php
```diff
@@ -110,10 +117,16 @@ private function pullCharity(
 
         $tbgCanClaimGiftAid = (
             !empty($hmrcReferenceNumber) &&
-            in_array($giftAidOnboardingStatus, static::$giftAidOnboardedStatuses, true)
+            in_array($giftAidOnboardingStatus, self::GIFT_AID_ONBOARDED_STATUSES, true)
+        );
+        $tbgApprovedToClaimGiftAid = (
+            !empty($hmrcReferenceNumber) &&
+            in_array($giftAidOnboardingStatus, self::GIFT_AID_APPROVED_STATUSES, true)
```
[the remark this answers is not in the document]
bdsl: Am I right to think this relies on the changes about to go into SF, and therefore this PR shouldn't be merged to develop until after the SF one is in production?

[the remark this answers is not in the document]
NoelLH: @bdsl In https://github.com/thebiggive/matchbot/pull/603/commits/8d247a3750a4e0d2b8de48104709da23163e889b I was thinking that setting a default would make it safe either way round, but I think you're right that there's a risk of subsequent data pulls putting the value back to a not-approved one. I don't think it's super critical though, at worst it might delay some claims slightly.

integrationTests/DonationRepositoryTest.php
```diff
@@ -0,0 +1,92 @@
...
+        $donationsReady = $sut->findReadyToClaimGiftAid(withResends: false);
+
+        $this->assertEquals([], $donationsReady);
+    }
+
+    private function prepareOnboardedCharity(bool $withAgentApproval): Charity
+    {
+        $charity = new Charity();
+        $charity->setName('Charity Name');
+        $charity->setHmrcReferenceNumber($withAgentApproval ? 'YY54321' : 'NN54321');
```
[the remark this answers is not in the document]
NoelLH: @bdsl I was trying to fix duplicate record errors, probably prior to doing wider data resets.

src/Domain/CampaignRepository.php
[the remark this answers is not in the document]
bdsl: Right, was thinking of it like a new field when matchbot might crash if the input doesn't have it, but as its just a new value I guess it should be safe enough, they'll show up as approved in matchbot the first time pull after the SF change is live.

integrationTests/DonationRepositoryTest.php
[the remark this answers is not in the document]
bdsl: Right, I think for integration tests randomising is probably the best way to avoid duplicate errors. We're not passing `true` and `false` within the same test run anyway, so I think this isn't needed to avoid duplicates within a test run, and the `clearPreviousCampaignsCharitiesAndRelated` runs to make duplicates not a problem between different tests anyway.

I just tried and the tests pass if I change this to

```suggestion
        $charity->setHmrcReferenceNumber($_specificRefNumberDoesntMatterHere = 'any_ref');
```

integrationTests/DonationRepositoryTest.php
```diff
@@ -0,0 +1,92 @@
...
+        $campaign->setSalesforceId('ccampaign123');
+        $campaign->setCurrencyCode('GBP');
+        $campaign->setStartDate((new \DateTime())->sub(new \DateInterval('P16D')));
+        $campaign->setEndDate((new \DateTime())->add(new \DateInterval('P15D')));
+        $campaign->setIsMatched(true);
+
+        return $campaign;
+    }
+
+    private function prepareDonation(Charity $charity): Donation
```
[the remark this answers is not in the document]
bdsl: Maybe rename to `prepareDonationInDB` or something?

src/Domain/Charity.php
NoelLH: @bdsl Thanks, yep – that makes more sense. I guess I was hoping there would be an answer that would lean on the Doctrine plugin to make this inferable without suppressing anything. Do you know what that plugin does with Doctrine stuff in relation to psalm? I think it's readme was pretty empty if I was looking in the right place.

integrationTests/DonationRepositoryTest.php
```diff
@@ -0,0 +1,92 @@
...
+
+class DonationRepositoryTest extends IntegrationTest
+{
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->clearPreviousCampaignsCharitiesAndRelated(); // Avoid e.g. HMRC ref dupes
+    }
+
+    public function testItFindsGiftAidSendableDonationsForCharityThatIsReady(): void
```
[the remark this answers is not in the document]
NoelLH: I personally don't find snake case easier, and would slightly favour keeping this consistent across tests

src/Domain/Charity.php
bdsl: I don't remember I'm afraid, I haven't used Psalm with Doctrine much, I think it's not a big issue, probably unusual that we'd have a property that we never read from PHP. Even if we don't strictly need to read something in a lot of cases we can put in a read that just checks that some entity isn't in a bad state for a given operation to happen, even if we know it wouldn't be because we wouldn't have selected it from the DB.

I guess we could add something like that in the Donation class, but I'm not sure I like the law of demeter violation and possible trigger of lazy loading:

image

integrationTests/DonationRepositoryTest.php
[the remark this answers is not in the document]
NoelLH: Yes, the workaround was written prior to introducing that method.

src/Domain/Charity.php
bdsl: Also for static analysis related to stuff in the DB I think we could do quite a bit more using PHPStan and https://github.com/staabm/phpstan-dba rather than only having Psalm. But given that matchbot doesn't have a lot of code changes probably not worth it any time soon.

integrationTests/DonationRepositoryTest.php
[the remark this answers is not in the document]
bdsl: OK, lets keep as-is and maybe at some point I'll propose doing changing all at once, I think there's a tool that will do it. I do think for a long name like this once you're used to it the extra negative space in snake case must make it easier to read. I have to look much more closely at each word to read it with camel case:

```
testItFindsGiftAidSendableDonationsForCharityThatIsReady

vs

test_it_finds_gift_aid_sendable_donations_for_charity_that_is_ready
```

src/Domain/Charity.php
bdsl: ah I see making the property private made it change from possiblyUnused to just unused in Psalm's eyes. Makes sense.

src/Domain/CampaignRepository.php
NoelLH: It's true the comment is out of date in terms of its list of fields, but I think the more important rationale – reducing complexity and load in Salesforce – probably still stands. In particular we don't want to double how many distinct calls to Salesforce it takes to keep MatchBot up to date because that's a high risk area in terms of sort limits that are hard to usefully load test against.

I'll amend the comment but I don't think we'd want to make Charity a totally independent proxy with its own Salesforce endpoint in the near future.

bdsl: @NoelLH That makes sense, updating the comment would be useful.

integrationTests/DonationRepositoryTest.php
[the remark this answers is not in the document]
NoelLH: I've gone with [prepareAndPersistDonation()](https://github.com/thebiggive/matchbot/pull/603/commits/5ef2ffc9e23688127bf51567ffeb81be292747dc) for now – trying to capture that it does some of the same object building as its analogue, plus the extra step of persisting.

src/Domain/CampaignRepository.php
```diff
@@ -13,9 +13,16 @@
...
         'Onboarded',
         'Onboarded & Data Sent to HMRC',
+        'Onboarded & Approved',
+        // We'll always aim to fix data problems with HMRC, so should still plan to claim.
+        'Onboarded but HMRC Rejected',
+    ];
+
+    private const GIFT_AID_APPROVED_STATUSES = [
+        'Onboarded & Approved',
     ];
```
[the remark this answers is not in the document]
NoelLH: Done – put in a separate commit as for some reason in this context I'm not allowed to add suggestions to a batch.

src/Domain/CampaignRepository.php
```diff
@@ -13,9 +13,14 @@
  */
 class CampaignRepository extends SalesforceReadProxyRepository
 {
-    private static array $giftAidOnboardedStatuses = [
+    private const GIFT_AID_APPROVED_STATUS = 'Onboarded & Approved';
+
+    private const GIFT_AID_ONBOARDED_STATUSES = [
         'Onboarded',
         'Onboarded & Data Sent to HMRC',
+        self::GIFT_AID_APPROVED_STATUS,
```
bdsl, line 21: :+1:

integrationTests/DonationRepositoryTest.php
NoelLH: Yeah, I think I'd misremembered what happened after I pointed out the risks with the database not being clearly separated out. I think I thought we'd made a new named schema, or at least restricted the app_env further to `test`, but if that's not the case I agree this could lead to unhelpful dev mishaps.

I'm really not a fan of the test data expanding indefinitely, especially as I have dozens of projects and run out of Docker space fairly regularly. I think if we could isolate the tests better and reset their DB regularly it would make it less likely we were relying on local side effects and would mean better dev quality of life around these tests.

bdsl: Right - I think a bit of expansion is OK, it's sometimes handy to look in the DB after running some tests to see what was left, and the data added here is going not going to take much space at all.

Or we maybe add a WHERE to the delete so we only delete the records created by this test instead of everything?

Not an absolute blocker if you want to merge with deleting everything, but I might change it later if I find it too annoying having the charities cleared on running integration tests.

src/Domain/Donation.php
NoelLH: I think this does need a migration to keep Doctrine happy because it stores a comment about the type to ensure it's handled consistently. That seems fine though – it's cool that this has a native Doctrine type now!

integrationTests/DonationRepositoryTest.php
bdsl: Let's leave as is for now, consider making a separate schema later.

src/Domain/CampaignRepository.php
bdsl: Did the move - may be still worth unit testing though? not sure.

tests/TestCase.php
bdsl: Created charity constructor that takes a nullable charity for now.

bdsl (approved):
LGTM, only thing I might do extra is unit testing \MatchBot\Domain\Charity::updateFromSfPull and maybe removing some API from Charity that's now just used within that method and in tests - but I could potentially do that after this is merged.

src/Domain/Charity.php
```diff
@@ -148,4 +162,44 @@ public function setRegulatorNumber(?string $regulatorNumber): void
...
+        ?string  $hmrcReferenceNumber,
+        ?string  $giftAidOnboardingStatus,
+        ?string  $regulator,
+        ?string  $regulatorNumber,
+        DateTime $time,
+    ): void
+    {
+        $this->setName($charityName);
+        $this->setStripeAccountId($stripeAccountId);
+
```
bdsl, line 183: maybe worth adding in a check that `$giftAidOnboardingStatus` is a status we've heard of and throwing if not - either with `in_array` and throw or by calling `Assertion::inArray(mixed $value, array $choices);` from https://github.com/beberlei/assert

PR #605, opened 2023-10-02
MAT-320 – handle `CardError`s on Confirm

NoelLH:


src/Application/Actions/Donations/Confirm.php
```diff
@@ -92,9 +91,28 @@ protected function action(Request $request, Response $response, array $args): Re
...
+                $exception->getMessage(),
+            ));
+
+            $this->entityManager->rollback();
+
+            return new JsonResponse([
+                'error' => [
+                    'message' => $exception->getMessage(),
+                    'code' => $exception->getStripeCode(),
+                    'decline_code' => $exception->getDeclineCode(),
```
bdsl, line 112: Depending whether the messages stripe generates are in the language we want to present, it might be worth matching on the decline code and presenting our own copy, at least for the ones we expect will be most common. Listed at https://stripe.com/docs/declines/codes?locale=en-GB

src/Application/Actions/Donations/Confirm.php
```diff
@@ -92,9 +91,28 @@ protected function action(Request $request, Response $response, array $args): Re
...
+                ],
+            ], 402);
+        }
```
bdsl, line 115: Maybe worth also catching `\Stripe\Exception\ApiErrorException`, in case it wasn't a CardException and loggin details from that, particularly the result of `getStripeCode`

src/Application/Actions/Donations/Confirm.php
```diff
@@ -92,9 +91,28 @@ protected function action(Request $request, Response $response, array $args): Re
             // Note that `on_behalf_of` is set up on create and is *not allowed* on update.
         ]);
 
-        $this->stripeClient->paymentIntents->confirm($paymentIntentId, [
-            'payment_method' => $pamentMethodId,
-        ]);
+        try {
+            $this->stripeClient->paymentIntents->confirm($paymentIntentId, [
+                'payment_method' => $pamentMethodId,
+            ]);
```
bdsl, line 97: I should have really thought that confirm might thrown and put in a try catch myself when I added this function call.

src/Application/Actions/Donations/Confirm.php
NoelLH: @bdsl The ones we've seen this AM say _Your card was declined._ I think simply presenting them to the donor will roughly mirror what Stripe.js did before.

src/Application/Actions/Donations/Confirm.php
NoelLH: Yep, good call. I think we probably still want this to be an `error` level log and a 500 code for the client – what do you reckon?

bdsl: Yes I think that's right, matchbot relies on stripe so if stripe is broken matchbot is broken and it should own up to being broken. Hopefully it'll never happen but if it does logs will help us.

bdsl (approved):
LGTM. Would be good to also test this with frontend if you haven't already and maybe add a screenshot to show how the decline message appears? I'm happy to do that if it helps.

NoelLH:
> LGTM. Would be good to also test this with frontend if you haven't already and maybe add a screenshot to show how the decline message appears? I'm happy to do that if it helps.

Yep I'm working on this @bdsl

PR #609, opened 2023-10-06
MAT-321 – fix crash reading not-set `Charity.name`s

NoelLH:


src/Application/Actions/Donations/Create.php
```diff
@@ -238,13 +238,21 @@ protected function action(Request $request, Response $response, array $args): Re
 
     private function getStatementDescriptor(Charity $charity): string
     {
+        $prefix = 'Big Give';
+        $prefixIncSpace = "$prefix ";
+
+        $charityName = $charity->getName();
+
+        if ($charityName === null) {
```
bdsl, line 246: Don't we need to put the charity name on the [statement descriptor](https://stripe.com/docs/payments/account/statement-descriptors?locale=en-GB)? I feel like if we don't know the charity's name and we can't generate a useful statement descriptor then we probably shouldn't be taking the donation, and should [let it crash](https://wiki.c2.com/?LetItCrash). (probably by throwing an exception rather than just letting PHP generate the typed property error).

I know on the ticket you said you' weren't sure of the reason for this - maybe we need to work out the reason before fixing it. I think we'd rather have a crash that stops us taking a donation than a donation taken that ends up stored without all the info necessary to deal with it properly later.

Seems like making `name` a required non-nullabale constructor param of `Charity` might be a good start.

bdsl: We only instantiate charity in one place in non-test code, and we already have a `$charityName` variable in scope there, so it should be straightforward to pass it to a constructor.

NoelLH: @bdsl I agree supporting missing names that late was a mistake. Suppressed in https://github.com/thebiggive/matchbot/pull/609/commits/3958c53452c4957dbcbbee4d50a09672c6310198 instead as discussed in standup.

Because there's scope for getting more info by getting something like this out quickly, I'd prefer to release it as is now rather than spend the time refactoring first.

src/Application/Actions/Donations/Create.php
```diff
@@ -238,21 +238,16 @@ protected function action(Request $request, Response $response, array $args): Re
...
-            return $prefix;
-        }
-
         $maximumLength = 22; // https://stripe.com/docs/payments/payment-intents#dynamic-statement-descriptor
+        $prefix = 'Big Give ';
 
-        return $prefixIncSpace . mb_substr(
-            $this->removeSpecialChars($charityName),
+        /**
+         * @psalm-suppress PossiblyNullArgument We don't expect to get here with no charity » let it crash for now.
```
bdsl, line 245: Yep, we're in strict types mode so if we don't have a charity name this [throws type error](https://3v4l.org/vCCDq). LGTM  to merge but IMHO we should keep the ticket where it is in Jira to do a more robust fix later.

bdsl (approved):

src/Application/Actions/Donations/Create.php
bdsl: ah don't know why I was testing passing null to strlen anyway, we're passing null to `removeSpecialChars` which of course will also throw the error.

PR #610, opened 2023-10-06
Book club related: Set Donation#amount and Donation#currencyCode read…

bdsl:
…only

Makes this a bit more 'partially immutable', communicates that we do not ever changes these values once a donation is recorded.

### DIFF
--- a/src/Domain/Donation.php
+++ b/src/Domain/Donation.php
@@ -35,11 +35,7 @@ class Donation extends SalesforceWriteProxy
     public const MAXIMUM_CARD_DONATION = 25_000;
 
     public const MAXIMUM_CUSTOMER_BALANCE_DONATION = 200_000;
-    /**
-     * @var int
-     * @see Donation::$currencyCode
-     */
-    private int $minimumAmount = 1;
+    public const MINUMUM_AMOUNT = 1;
 
     private array $possiblePSPs = ['stripe'];
 
@@ -317,12 +313,12 @@ class Donation extends SalesforceWriteProxy
         $maximumAmount = self::maximumAmount($paymentMethodType);
 
         if (
-            bccomp($amount, (string)$this->minimumAmount, 2) === -1 ||
+            bccomp($amount, (string)self::MINUMUM_AMOUNT, 2) === -1 ||
             bccomp($amount, (string)$maximumAmount, 2) === 1
         ) {
             throw new \UnexpectedValueException(sprintf(
                 'Amount must be %d-%d %s',
-                $this->minimumAmount,
+                self::MINUMUM_AMOUNT,
                 $maximumAmount,
                 $this->currencyCode,
             ));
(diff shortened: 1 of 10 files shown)


src/Domain/Donation.php
```diff
@@ -15,6 +15,8 @@
 use Messages;
```
bdsl, line 15: We just read the section on "Partially Immutable Entities" in book club. I took it as good advice to start by making all properties of everything immutable and then only allow mutations for things that we actually want to change. We never want to change the amount or the currency on any donation in our system.

This is the only file used in in production changed in this PR.

src/Domain/Donation.php
```diff
@@ -303,16 +305,31 @@ class Donation extends SalesforceWriteProxy
...
+            throw new \UnexpectedValueException(sprintf(
+                'Amount must be %d-%d %s',
+                $this->minimumAmount,
+                $maximumAmount,
+                $this->currencyCode,
+            ));
+        }
+
+        $this->amount = $amount;
+        $this->paymentMethodType = $paymentMethodType;
```
bdsl, line 332: Psalm only allows assignment to readonly properties from inside the constructor, hence inlining the `setAmountAndMethod` function.

src/Domain/Donation.php
```diff
@@ -95,7 +97,7 @@ class Donation extends SalesforceWriteProxy
      * @ORM\Column(type="string", length=3)
      * @var string  ISO 4217 code for the currency in which all monetary values are denominated, e.g. 'GBP'.
      */
-    protected string $currencyCode;
+    protected readonly string $currencyCode;
 
     /**
      * Core donation amount excluding any tip.
```
bdsl, line 103:
```suggestion
     * Core donation amount in major currency units (i.e. Pounds) excluding any tip.
```

src/Domain/Donation.php
```diff
@@ -95,16 +93,16 @@ class Donation extends SalesforceWriteProxy
      * @ORM\Column(type="string", length=3)
      * @var string  ISO 4217 code for the currency in which all monetary values are denominated, e.g. 'GBP'.
      */
-    protected string $currencyCode;
+    protected readonly string $currencyCode;
```
bdsl, line 96: Doctrine says "An entity class must not be final nor read-only but it may contain final methods or read-only properties."

src/Domain/Donation.php
```diff
@@ -33,11 +35,7 @@ class Donation extends SalesforceWriteProxy
     public const MAXIMUM_CARD_DONATION = 25_000;
 
     public const MAXIMUM_CUSTOMER_BALANCE_DONATION = 200_000;
-    /**
-     * @var int
-     * @see Donation::$currencyCode
-     */
-    private int $minimumAmount = 1;
```
bdsl, line 40: For some reason Psalm seemed to think this property was unused. I think it might be a Psalm bug but I replaced it with a constant anyway. It can't really be a readonly property as readonly properties aren't allowed to have values set on declaration (since doing that would make them just like constants)

NoelLH (approved):

PR #611, opened 2023-10-06
MAT-321: Fix crash during Donation Create persistence retry, due to m…

bdsl:
…issing charity name

Introducing a constructor for Charity that requires all the parameters that we would pass to the updateFromSfPull function

### DIFF
--- a/src/Domain/CampaignRepository.php
+++ b/src/Domain/CampaignRepository.php
@@ -91,26 +91,33 @@ class CampaignRepository extends SalesforceReadProxyRepository
         ?string $stripeAccountId,
         ?string $giftAidOnboardingStatus,
         ?string $hmrcReferenceNumber,
-        ?string $regulator,
+        string $regulator,
         ?string $regulatorNumber,
     ): Charity {
         $charity = $this->getEntityManager()
             ->getRepository(Charity::class)
             ->findOneBy(['salesforceId' => $salesforceCharityId]);
         if (!$charity) {
-            $charity = new Charity();
-            $charity->setSalesforceId($salesforceCharityId);
+            $charity = new Charity(
+                charityName: $charityName,
+                stripeAccountId: $stripeAccountId,
+                hmrcReferenceNumber: $hmrcReferenceNumber,
+                giftAidOnboardingStatus: $giftAidOnboardingStatus,
+                regulator: $this->getRegulatorHMRCIdentifier($regulator),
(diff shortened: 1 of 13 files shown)


src/Domain/CampaignRepository.php
```diff
@@ -91,27 +91,34 @@ private function pullCharity(
         ?string $stripeAccountId,
         ?string $giftAidOnboardingStatus,
         ?string $hmrcReferenceNumber,
-        ?string $regulator,
```
bdsl, line 94: this may as well be non-null because it is being passed to `getRegulatorHMRCIdentifier` anyway, and that already requires a non-null value.

bdsl:
There are 13 files changed here, but only four in `src`, and not many lines changed. The changes in tests are mostly one search and replace.

image

src/Domain/Charity.php
```diff
@@ -87,6 +87,31 @@ class Charity extends SalesforceReadProxy
      */
     private bool $tbgApprovedToClaimGiftAid = false;
 
+    public function __construct(
+        string $charityName,
```
bdsl, line 91: this seems like a good and standard way to ensure every charity has a name - but I'm looking at the current version of the code that we had before deployment today (commit # 663b30e6eaec28b ) and I'm confused about how that caused a crash - there doesn't seem to be any opportunity for anything to need to read the charity name in the time between when we instantiate the charity and when we set the name to a non-null string:

image

`setSalesforceId` didn't do any funny business.

bdsl: Maybe it's related to this - Doctrine somehow creates a proxy object from nothing based on the relationship? https://github.com/doctrine/orm/issues/3945

bdsl: and I think https://github.com/thebiggive/matchbot/blob/70e2cfa735f7aaf39e924069e043f666f19bf24f/src/Domain/FundRepository.php#L66 means a Campaign can get persisted with no charity. In the DB campaign.charity_id is nullable, but in our Campaign php class $charity is non-nullable. So it seams the campaign gets persisted with no charity, and then when it gets loaded up Doctrine hallucinates a Charity with no properties.

So maybe the fix for this actually isn't giving Charity a no-nullable name, its making Campaign enforce that it must have a Charity, via the DB column and a Campaign constructor.

dorota-joanna (approved):
LGTM

PR #612, opened 2023-10-07
Log less noisily if attempts are made to reuse stale Payment Methods

NoelLH:


src/Application/Actions/Donations/Confirm.php
```diff
@@ -114,6 +115,28 @@ protected function action(Request $request, Response $response, array $args): Re
                 ],
             ], 402);
         } catch (ApiErrorException $exception) {
+            $paymentMethodReuseAttempted = (
+                $exception instanceof InvalidRequestException &&
+                str_contains($exception->getMessage(), 'The provided PaymentMethod was previously used')
```
bdsl, line 120: Can you add a comment about why we get this message and maybe an example of the full message with any specific details redacted?

Thanks

bdsl (approved):

PR #613, opened 2023-10-09
MAT-321: Fix bug introduced in commit 037a7cf22

bdsl:
SF ID has to be set for all charities. The DB doesn't enforce this, but if we save a charity with no SF ID then we won't find it when we need it and we'll try to pull it from SF again - and get a constraint violation error on the duplicate Stripe ID.

This bug has not been in production - I found it on my local

### DIFF
--- a/src/Domain/CampaignRepository.php
+++ b/src/Domain/CampaignRepository.php
@@ -94,7 +94,6 @@ class CampaignRepository extends SalesforceReadProxyRepository
         string $regulator,
         ?string $regulatorNumber,
     ): Charity {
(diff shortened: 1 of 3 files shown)


src/Domain/CampaignRepository.php
```diff
@@ -94,11 +94,13 @@ private function pullCharity(
         string $regulator,
         ?string $regulatorNumber,
     ): Charity {
+        $this->logger->info("Searching for charity by sf ID: $salesforceCharityId");
```
bdsl, line 97:
```suggestion
```

PR #615, opened 2023-10-10
MAT-321: Prevent persisting any Campaign without a Charity ID

bdsl:
Having a campaign without a charity might be the cause of a crash on attempt to persist a donation. This may bring any crash earlier and make it easier to understand deal with.

I have checked the production database, there are currently no campaigns without charities, i.e. the following returns the empty relation:

```
   SELECT * from Campaign where Campaign.charity_id IS NULL;
```

### DIFF
--- a/src/Domain/Campaign.php
+++ b/src/Domain/Campaign.php
@@ -24,6 +24,7 @@ class Campaign extends SalesforceReadProxy
 
     /**
      * @ORM\ManyToOne(targetEntity="Charity", cascade={"persist"})
+     * @ORM\JoinColumn(name="charity_id", referencedColumnName="id", nullable=false)
      * @var Charity
      */
     protected Charity $charity;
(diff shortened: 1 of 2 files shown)


bdsl:
I have just confirmed that I can still make a donation on my local with this branch after clearing out the Matchbot database so it has to pull my selected charity and campaign from SF again.

src/Domain/Campaign.php
```diff
@@ -24,6 +24,7 @@ class Campaign extends SalesforceReadProxy
 
     /**
      * @ORM\ManyToOne(targetEntity="Charity", cascade={"persist"})
+     * @ORM\JoinColumn(name="charity_id", referencedColumnName="id", nullable=false)
```
bdsl, line 27: see https://stackoverflow.com/a/9674597/2526181

dorota-joanna (approved):
LGTM

PR #618, opened 2023-10-11
Mat 123 persistence crash

bdsl:
This doesn't fix the crash but adds a test that demonstrates how the crash happens and asserts that it does in fact happen when the underlying EM closes and our system auto retries. This is basically just adding new test code, there are only very minor changes to production code either to improve logging by adding __toString methods or to make testing possible to allow injecting the EM factory.

May be better to squash merge isntead or regular merge this - I think there's a lot of changes to production code that would be confusing if seen in isolation and I have reverse.

There's a lot of back and forth work in the commits that IMHO isn't worth reviewing - I would just review the overall PR diff.

### DIFF
--- a/integrationTests/CreateDonationEvenWhenEntityManagerClosesAfterFirstPersistsTest.php
+++ b/integrationTests/CreateDonationEvenWhenEntityManagerClosesAfterFirstPersistsTest.php
@@ -40,7 +40,7 @@ use Ramsey\Uuid\Uuid;
 class CreateDonationEvenWhenEntityManagerClosesAfterFirstPersistsTest extends IntegrationTest
 {
     private int $entityManagerWillCloseAfterNPersists = 3;
-    private $campaignSfID;
+    private string $campaignSfID;
 
     public function setUp(): void
     {
@@ -145,7 +145,6 @@ class CreateDonationEvenWhenEntityManagerClosesAfterFirstPersistsTest extends In
     }
 
     /**
-     * @return EntityManager
      * @throws \Doctrine\ORM\Exception\ORMException
      * @throws \Psr\Container\ContainerExceptionInterface
      * @throws \Psr\Container\NotFoundExceptionInterface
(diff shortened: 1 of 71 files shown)


integrationTests/CreateDonationWithPersistErrorTest.php
```diff
@@ -0,0 +1,141 @@
...
+         *
+         * @psalm-suppress DeprecatedMethod
+         * @psalm-suppress MixedArrayAccess
+         * @psalm-suppress MixedArgument
+         * We're using EntityManager::create already inside RetrySafeEntityManager - while we have it there we may as
+         * well have it in this test. Also probably not worthwhile fixing the Mixed issues here before they're
+         * fixed in the similar prod code.
+         */
+        $this->setInContainer(EntityManagerInterface::class, EntityManager::create(
+            // for this test we don't need the RetrySafeEntityManager - using the standard EM makes things simpler.
```
bdsl, line 97: actually this is misleading - in this test the crash only happens because have two different entity manages due to this line. In real scenarios we probably have a replaced underlying em because the original one closed. I been testing and trying to fix that scenario on my local.

integrationTests/CreateDonationEvenWhenEntityManagerClosesAfterFirstPersistsTest.php
```diff
@@ -0,0 +1,176 @@
...
+        //persist operations for entity
+
+        $this->createDonation(withPremadeCampaign: false);
+        $donationFetchedFromDB = $this->db()->fetchAssociative(
+            "SELECT * from Donation
+                   JOIN Campaign on Donation.campaign_id = Campaign.id
+         where Campaign.salesforceId = '{$this->campaignSfID}';"
+        );
+
+        $this->assertSame([], $donationFetchedFromDB);
```
dorota-joanna, line 142:
```suggestion
```

dorota-joanna (approved):
LGTM

PR #622, opened 2023-10-12
Fix TypeError on attempt to submit donation with null amount

bdsl:


app/dependencies.php
```diff
@@ -323,7 +324,7 @@ function (ContainerInterface $c): Auth\DonationPublicAuthMiddleware {
             $encoders = [new JsonEncoder()];
             $normalizers = [
                 new BackedEnumNormalizer(),
-                new ObjectNormalizer(),
+                new ObjectNormalizer(propertyTypeExtractor: new ReflectionExtractor()),
```
bdsl, line 327: without the ReflectionExtractor the constructor throws a TypeError, so matchbot returns 500 and alarms us if the browser sends a donation with null amount.  With this change and catching the `NotNormalizableValueException` we return a 400 error to the browser instead.

dorota-joanna (approved):
LGTM

PR #624, opened 2023-10-13
MAT-306: Allow creating donor accounts

bdsl:
Once this is in Matchbot, frontend should be able to call it to create a donor account when required. Then stripe can call matchbot to say cash has been transferred to an account, matchbot will be able to look up the account and finally use mailer to email the account holder.

### DIFF
--- a/src/Application/Auth/PersonWithPasswordAuthMiddleware.php
+++ b/src/Application/Auth/PersonWithPasswordAuthMiddleware.php
@@ -13,6 +13,8 @@ use Psr\Http\Message\ServerRequestInterface;
  * Permits only true "complete" value in token claims. Use this for anything that should
  * be returned only to those who've set a password and so expect e.g. their saved payment
  * methods to be available longer term.
+ *
+ * // note use this for making donation funds account.
  */
 class PersonWithPasswordAuthMiddleware extends PersonManagementAuthMiddleware
 {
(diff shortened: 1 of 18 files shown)


src/Application/Auth/PersonWithPasswordAuthMiddleware.php
```diff
@@ -13,6 +13,8 @@
  * Permits only true "complete" value in token claims. Use this for anything that should
  * be returned only to those who've set a password and so expect e.g. their saved payment
  * methods to be available longer term.
+ *
+ * // note use this for making donation funds account.
```
bdsl, line 17:
```suggestion
```

src/Domain/EmailAddress.php
```diff
@@ -0,0 +1,29 @@
...
+        string $value
+    ){
+        $this->email = $value;
+        Assertion::email($this->email);
+    }
+
+    /**
+     * @param string $emailAddress - must be a valid email address.
+     */
+    public static function of(string $emailAddress): self
```
bdsl, line 25: having a constructor called `of` is conventional for value objects - see e.g. https://github.com/brick/date-time/blob/master/src/Year.php#L37

src/Domain/EmailAddress.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace MatchBot\Domain;
+
+use Doctrine\ORM\Mapping\Column;
+use Doctrine\ORM\Mapping\Embeddable;
+use MatchBot\Application\Assertion;
+
+/** @Embeddable */
```
bdsl, line 9: slight issue with using an Embeddable value objects - in some cases the ORM can change the value even if we thought they were immutable, see https://github.com/doctrine/orm/issues/7539 . That would be a problem if the VO was shared between two things, which is fine for a genuinely immutable object. But that's only an issue when use `EntityManager#refresh`, which we don't use much. 

Or we only use for Donations: https://github.com/thebiggive/matchbot/blob/bd90fab20a39b9f6a00dd5f5e0709e586274067a/src/Domain/DonationRepository.php#L698

dorota-joanna (approved):
LGTM

PR #631, opened 2023-10-18
Refactor - reduce duplication of string literal status names

bdsl:
By referencing the enum each time instead of repeating the literal we allow static analysers to check that we haven't typo'd, get earlier runtime crashes in case of error, and allow clicking through to read documentation of each status case from anywhere it's referenced.

The one place were I haven't done this is in migration files, as those need to keep the same literal string value even if we change or remove the DonationStatus case in future - migrations are basically write-only files and shouldn't depend on things that we might change.


dorota-joanna (approved):
LGTM

PR #632, opened 2023-10-18
MAT-306 - improve handling of bad requests

bdsl:
This does two things, both about rejecting unwanted requests for creating donor accounts:

- If we get a request without all required data, we now return a 400 response instead of an alarming 500.
- If we get a request to create a duplicate account (i.e. with a stripe customer ID we've already got an account for) we reject that as well, with a 400, where previously we would have put the dupe into our DB.

### DIFF
--- a/src/Application/Actions/DonorAccount/Create.php
+++ b/src/Application/Actions/DonorAccount/Create.php
@@ -5,7 +5,9 @@ declare(strict_types=1);
 namespace MatchBot\Application\Actions\DonorAccount;
 
 use MatchBot\Application\Actions\Action;
+use MatchBot\Application\AssertionFailedException;
 use MatchBot\Application\Auth\PersonManagementAuthMiddleware;
+use MatchBot\Application\LazyAssertionException;
 use MatchBot\Domain\DonorAccount;
 use MatchBot\Domain\DonorAccountRepository;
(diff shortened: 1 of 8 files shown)


src/Application/Actions/DonorAccount/Create.php
```diff
@@ -45,21 +48,40 @@ protected function action(Request $request, Response $response, array $args): Re
...
 
-        /** @var array{firstName: string, lastName: string} $donorName */
-        $donorName = $requestBody['donorName'];
+        /** @var array{firstName: string, lastName: string} $donorNameArray */
+        $donorNameArray = $requestBody['donorName'] ?? throw new HttpBadRequestException($request, 'Expected donorName');
+
+        try {
+            $emailAddress = EmailAddress::of($emailAddressString);
+            $donorName = DonorName::of($donorNameArray['firstName'], $donorNameArray['lastName']);
+            $stripeCustomerId = StripeCustomerId::of($stripeCustomerIdString);
```
bdsl, line 63: A nicer implementation would collect all exceptions and tell the client about them, instead of bailing out on the first one. We could look at valinor for that in future. But this is good enough for now I think. These should never happen unless there is a bug in frontend or someone is using different software to connect to matchbot.

dorota-joanna (approved):
LGTM

PR #635, opened 2023-10-20
MAT-306: Take stripe webhooks for cash balance updates and just send …

bdsl:
…to Slack for now.

### DIFF
--- a/src/Application/Actions/Hooks/Stripe.php
+++ b/src/Application/Actions/Hooks/Stripe.php
@@ -15,7 +15,9 @@ use Stripe\Event;
 
 abstract class Stripe extends Action
 {
-    protected Event $event;
+    protected ?Event $event = null;
+
+    /** @var array{connectAppWebhookSecret: string, accountWebhookSecret: string} */
     protected array $stripeSettings;
 
     public function __construct(ContainerInterface $container, LoggerInterface $logger)
--- a/src/Application/Actions/Hooks/StripeCashBalanceUpdate.php
+++ b/src/Application/Actions/Hooks/StripeCashBalanceUpdate.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MatchBot\Application\Actions\Hooks;
+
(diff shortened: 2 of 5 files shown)


src/Application/Actions/Hooks/Stripe.php
```diff
@@ -15,7 +15,9 @@
 
 abstract class Stripe extends Action
 {
-    protected Event $event;
+    protected ?Event $event = null;
```
bdsl, line 18: As I created a new child of this class Psalm complained that Event was not set in constructor. So my solution was to initialise it to null here and insert null checks everywhere it's used.

Probably it would be better to remove the property entirely and replace it with local variables, but adding null checks seemed like an easier change.

src/Application/Actions/Hooks/StripeCashBalanceUpdate.php
```diff
@@ -0,0 +1,70 @@
...
+            $event->account,
+        ));
+
+        if (getenv('APP_ENV') === 'production') {
+            return $this->respond($response, new ActionPayload(200));
+        } else {
+            // for now, and outside of production, just send the webhook to slack so we can see if we like it.
+            $this->chatter->send(new ChatMessage(
+                "StripeCashBalanceUpdate in development - recieved webhook from stripe: \n\n" .
+                $event->toJSON()
```
bdsl, line 65: I'm setting up the webhook in the test stripe account. Searched for stripe balance and selected all four matching events. Stripe then gives us the following suggested PHP code to use:

I was planning on just listening for the `cash_balance.funds_available` but perhaps we might as well listen to all four here even if we do no more than log some of them.

```php
switch ($event->type) {
  case 'cash_balance.funds_available':
    $cashBalance = $event->data->object;
  case 'customer_cash_balance_transaction.created':
    $customerCashBalanceTransaction = $event->data->object;
  case 'refund.created':
    $refund = $event->data->object;
  case 'refund.updated':
    $refund = $event->data->object;
  // ... handle other event types
  default:
    echo 'Received unknown event type ' . $event->type;
}
```

PR #639, opened 2023-10-20
MAT-306: Add client to connect to mailer service

bdsl:


matchbot
bdsl, line 1: This is just here because I made a throwaway command class on my local to test ending email and see that it works. The `SendEmail` class wasn't worth committing but I think this is worth keeping for some extra convenience.

matchbot-cli.php
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env php
```
bdsl, line 1: see https://github.com/thebiggive/matchbot/pull/639/files#r1367150515 . Changed file mode to executable.

dorota-joanna (approved):
LGTM

PR #641, opened 2023-10-23
Mat 306 send emails about funds

bdsl:


src/Application/Actions/Hooks/StripePaymentsUpdate.php
```diff
@@ -458,15 +465,36 @@ private function doPostMarkRefundedUpdates(Donation $donation, bool $isCoreDonat
...
     {
-        if (getenv('APP_ENV') === 'production') {
+        Assertion::eq('customer_cash_balance_transaction.created', $event->type);
+
+        /** @var array{customer: string, currency: string, net_amount: int, ending_balance: int, type: string} $webhookObject */
+        $webhookObject = $event->data->toArray()['object'];
+
+        $stripeAccountId = StripeCustomerId::of($webhookObject['customer']);
+
+        $this->chatter->send(new ChatMessage(
```
NoelLH, line 475: Should we think about switching to an info log for the "happy path", so we can check what notifications we sent out on-demand if there are queries, while not having the noise of a Slack alert on every top up?

NoelLH (commented):
Looks good! Just one note inline about monitoring / logs

src/Application/Actions/Hooks/StripePaymentsUpdate.php
bdsl: I'd like to have some slack alerts for the first few times it happens, but then yes it would make sense to delete it. I'll add the info log now, if you think we should delete the send in the happy path I'll delete that as well.

Although this probably should be moved underneath the check for whether the type is 'funded' - otherwise this would give us alerts on bottom downs as well as top ups.

bdsl: Actually done in 6977c8b.

src/Domain/Money.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace MatchBot\Domain;
+
+use MatchBot\Application\Assertion;
+
+/**
+ * @psalm-immutable
+ */
+class Money
```
bdsl, line 10: We could use a third party library like https://github.com/brick/money or https://github.com/moneyphp/money to save us some work, either instead of this class or inside of it. Might very well be worth it before we make this class more complicated, but as for now we have just one very basic use case it seemed simpler to write this class than make a decision about what third partly library to use if any.

PR #646, opened 2023-10-26
Mat 322 fix annotation

bdsl:
This was part of the much bigger PR #645 where I was trying to stop letting Doctrine manage the association between Donation and Campaign - but that's several kinds of broken. So I've just salvaged the first three commits from it here.


dorota-joanna (approved):
LGTM

PR #650, opened 2023-10-27
MAT-322 – `allocateMatchFunds()`: save funding withdrawals later, wit…

bdsl:
…hout a DB transaction

And remove likely-unhelpful exception recovery


bdsl (commented):
LGTM

PR #651, opened 2023-10-27
BG-2297: Report errors to Slack

bdsl:


public/index.php
```diff
@@ -4,6 +4,7 @@
 
 use MatchBot\Application\Handlers\HttpErrorHandler;
 use MatchBot\Application\Handlers\ShutdownHandler;
+use MatchBot\Application\SlackChannelChatterFactory;
```
bdsl, line 7:
```suggestion
```

src/Application/Handlers/HttpErrorHandler.php
```diff
@@ -7,7 +7,10 @@
 use Exception;
 use MatchBot\Application\Actions\ActionError;
 use MatchBot\Application\Actions\ActionPayload;
+use MatchBot\Application\SlackChannelChatterFactory;
+use Psr\Http\Message\ResponseFactoryInterface;
 use Psr\Http\Message\ResponseInterface as Response;
+use Psr\Log\LoggerInterface;
 use Slim\Exception\HttpBadRequestException;
 use Slim\Exception\HttpException;
 use Slim\Exception\HttpForbiddenException;
```
bdsl, line 16:
```suggestion
use Psr\Http\Message\ResponseInterface as Response;
use Slim\Exception\HttpBadRequestException;
use Slim\Exception\HttpException;
use Slim\Exception\HttpForbiddenException;
```

src/Application/Handlers/HttpErrorHandler.php
```diff
@@ -16,6 +19,7 @@
 use Slim\Exception\HttpNotImplementedException;
 use Slim\Exception\HttpUnauthorizedException;
 use Slim\Handlers\ErrorHandler as SlimErrorHandler;
+use Slim\Interfaces\CallableResolverInterface;
```
bdsl, line 22:
```suggestion
```

bdsl:
I've tested what I can on my local. There's no slack connection there, but the "Please throw" bit works. The rest I think we can just test on staging.

src/Monolog/Handler/SlackHandler.php
```diff
@@ -0,0 +1,66 @@
...
+     * @psalm-suppress PossiblyUnusedReturnValue - used by Monolog.
+     */
+    public function handle(array $record): bool
+    {
+        if ($record['level'] < Logger::ERROR) {
+            return false; // allows another handle to handle this.
+         }
+
+        $message = $record['message'];
+        $levelName = $record['level_name'];
```
bdsl, line 46: We could also serialise `$record['context']` and include it in the slack message - but I thought maybe better to leave that out. If we want those details we can find them in the logs elsewhere, and we might choose to put more sensitive info in the context than we do in the message sometimes. Although we probably just avoid putting sensitive info in either and I think we don't actually use context much if at all.

bdsl: See docs for Record type at https://github.com/Seldaek/monolog/blob/2.x/doc/message-structure.md

src/Application/Actions/Donations/Update.php
```diff
@@ -124,7 +132,8 @@ protected function action(Request $request, Response $response, array $args): Re
                         "Donation ID {$args['donationId']} auto-confirm attempted with '$methodSummary' payment method",
                     );
 
-                    return $this->validationError($response,
+                    return $this->validationError(
+                        $response,
```
bdsl, line 136: I pressed the button to fix the whole file according to style rules.

src/Monolog/Handler/SlackHandler.php
NoelLH, line 1: This could perhaps use Monolog's own https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/SlackHandler.php as discussed, but either is OK I think

NoelLH (approved):

src/Monolog/Handler/SlackHandler.php
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Sends error messages to our slack alarm channel so we deal with the errors without unnecessary delay.
+ *
+ * @phpstan-import-type Record from \Monolog\Logger
+ * @phpstan-import-type Level from \Monolog\Logger
+ * @psalm-import-type Record from \Monolog\Logger
+ * @psalm-import-type Level from \Monolog\Logger
+ *
+ */
+class SlackHandler implements HandlerInterface
```
bdsl, line 19: As discussed this class is probably unnecessary - may well delete on Monday and replace with https://github.com/Seldaek/monolog/blob/2.x/src/Monolog/Handler/SlackHandler.php

PR #653, opened 2023-10-30
BG2-2297: Add erroring file and line number to error logs

bdsl:
`getTraceAsString` only gives us all the *previous* stack frames before the one that actually has the `throw` statement. Using `getFile` and `getLine` will tell us exactly where the throw expression is that threw the exception.

### DIFF
--- a/src/Application/Handlers/HttpErrorHandler.php
+++ b/src/Application/Handlers/HttpErrorHandler.php
@@ -67,7 +67,7 @@ class HttpErrorHandler extends SlimErrorHandler
 
         if (!($this->exception instanceof HttpException)) {
             $this->logError(sprintf(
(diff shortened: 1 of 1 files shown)


src/Application/Handlers/HttpErrorHandler.php
```diff
@@ -67,9 +67,10 @@ protected function respond(): Response
 
         if (!($this->exception instanceof HttpException)) {
             $this->logError(sprintf(
-                '%s: %s - %s',
+                "%s: %s \n\n %s \n %s",
```
bdsl, line 70:
```suggestion
                "%s: %s \n#\n %s \n %s",
```

NoelLH (approved):

PR #654, opened 2023-10-30
MAT-306: Add 30 second delay to let Stripe update balance before send…

bdsl:
…ing email notification of transferred funds

This may look like a lot of code just to add a 30 second delay. But it's more complicated than that because:

- It's not simply adding a delay before work we were already doing. Currently stripe sends us the transfer amount and the new balance together, and we use both. Now we have to ignore the balance stripe sends us initially, and query stripe for a balance after the delay.
- We definitely don't want to have a real 30 second delay every time we run the automated tests. There are hundreds of tests and we want to run them many times a day so we can't afford even one second per test. Therefore the code that actually does the delay has to be separated from the code we test.

### DIFF
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -151,11 +151,38 @@ return function (ContainerBuilder $containerBuilder) {
             return new Client\Mailer($settings, $c->get(LoggerInterface::class));
         },
 
+        \Symfony\Component\Clock\ClockInterface::class =>
+            function (ContainerInterface $_c): \Symfony\Component\Clock\ClockInterface {
+                return new \Symfony\Component\Clock\Clock();
+                },
+
+        Client\Stripe::class => function (ContainerInterface $c): Client\Stripe {
+            $offTheShelfStripeClient = $c->get(StripeClient::class);
+
+            \assert($offTheShelfStripeClient instanceof StripeClient);
+
+            return new Client\Stripe($offTheShelfStripeClient);
+        },
+
         \MatchBot\Domain\DonationFundsNotifier::class => function (ContainerInterface $c): \MatchBot\Domain\DonationFundsNotifier {
             $mailer = $c->get(Client\Mailer::class);
             \assert($mailer instanceof Client\Mailer);
 
-            return new \MatchBot\Domain\DonationFundsNotifier($mailer);
+            $stripeClient = $c->get(Client\Stripe::class);
+            \assert($stripeClient instanceof Client\Stripe);
+
+            $clock = $c->get(\Symfony\Component\Clock\ClockInterface::class);
+            \assert($clock instanceof \Symfony\Component\Clock\ClockInterface);
+
+            $logger = $c->get(LoggerInterface::class);
+            \assert($logger instanceof LoggerInterface);
+
+            return new \MatchBot\Domain\DonationFundsNotifier(
+                $mailer,
+                $stripeClient,
+                $clock,
+                $logger,
+            );
         },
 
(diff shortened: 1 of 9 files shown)


src/Client/Stripe.php
```diff
@@ -0,0 +1,40 @@
...
+use MatchBot\Domain\StripeCustomerId;
+use Stripe\StripeClient;
+
+class Stripe
+{
+    public function __construct(private StripeClient $stripeClient)
+    {
+    }
+
+    public function fetchBalance(StripeCustomerId $stripeCustomerId, Currency $currency): Money
```
bdsl, line 16: Implementation here is largely copied from https://github.com/thebiggive/identity/blob/453cfc343ebabdd738a790d67d1f126a888983f2/src/Application/Actions/Person/Get.php#L94

src/Client/Stripe.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace MatchBot\Client;
+
+use MatchBot\Domain\Currency;
+use MatchBot\Domain\Money;
+use MatchBot\Domain\StripeCustomerId;
+use Stripe\StripeClient;
+
+class Stripe
```
bdsl, line 10: I wanted to make a class just to isolate how we talk to stripe. The off the shelf stripe client is very tedious to double with prophecy, and overly complicated to use. In principle we should also have a unit test for this class, but as the implementation of fetchBalance is largely copied from Identity I'm quite confident that it's correct.

src/Application/Actions/Hooks/StripePaymentsUpdate.php
```diff
@@ -495,7 +495,7 @@ private function handleCashBalanceUpdate(Event $event, Response $response): Resp
             return $this->respond($response, new ActionPayload(200));
         }
 
-        $this->donationFundsNotifier->notifyRecieptOfAccountFunds($donorAccount, $transferAmount, $endingBalance);
+        $this->donationFundsNotifier->notifyRecieptOfAccountFunds($donorAccount, $transferAmount);
```
bdsl, line 499: no longer passing the endingBalance because the notifier will wait 30 seconds and pull a new balance from Stripe.

app/dependencies.php
```diff
@@ -151,11 +151,38 @@ function (ContainerInterface $c): Auth\DonationPublicAuthMiddleware {
             return new Client\Mailer($settings, $c->get(LoggerInterface::class));
         },
 
+        \Symfony\Component\Clock\ClockInterface::class =>
+            function (ContainerInterface $_c): \Symfony\Component\Clock\ClockInterface {
+                return new \Symfony\Component\Clock\Clock();
```
bdsl, line 156: Introducing the [Symofony Clock](https://symfony.com/doc/current/components/clock.html) component here. @NoelLH @dorota-joanna what do you think of making it a rule from now on to use the clock component rather than direct access to PHP functions when we need to deal with the system clock, at least when what it does with that clock is something we care to test? 

IMHO it's often hard to write and read tests for things that directly uses the system clock, so injecting a clock a dependency is a god solution.

NoelLH: Sounds good!

src/Client/Stripe.php
NoelLH: I think this is reasonable, but I'd also be keen for us to add a test in the next couple of weeks while we're in "code slush", since this is newly becoming our abstraction and is a particularly important area.

NoelLH (approved):
Nice!

PR #659, opened 2023-10-31
Give `NotFoundException`s requested ID in their message

NoelLH:


src/Client/Campaign.php
```diff
@@ -19,7 +19,7 @@ public function getById(string $id): array
             $response = $this->getHttpClient()->get("{$this->getSetting('campaign', 'baseUri')}/$id");
         } catch (RequestException $exception) {
             if ($exception->getResponse() && $exception->getResponse()->getStatusCode() === 404) {
-                throw new NotFoundException('Campaign not found'); // may be safely caught in sandboxes
+                throw new NotFoundException(sprintf('Campaign ID %s not found', $id)); // may be safely caught in sandboxes
```
bdsl, line 22: Worth adding this to make clear that its looking in SF not in MySQL? 
```suggestion
                throw new NotFoundException(sprintf('Campaign ID %s not found in SF', $id)); // may be safely caught in sandboxes
```

bdsl (approved):
Added a suggestion, but feel fee to merge with or without taking it.

bdsl (approved):

PR #660, opened 2023-10-31
DON-820: Reject attempt to confirm funds donation payment does not im…

bdsl:
…ediatly succeed

We only want to accept a donation from donor funds if the donor already has enough funds in their account, in which case the payment should succeed immediately.

I missed my chance to take a screenshot but I did test manually with real frontend making two donations in two tabs from each using 1k of a 1k donation fund account. The first donation went through as normal, and the second donation resulted in a spinner spinning for a long time on the donation start page and the error message asking the donor to refresh to see their updated balance. Checking in stripe showed that one of the payment intents was cancelled.

### DIFF
--- a/src/Application/Actions/Donations/Update.php
+++ b/src/Application/Actions/Donations/Update.php
@@ -19,9 +19,11 @@ use Psr\Http\Message\ResponseInterface as Response;
 use Psr\Http\Message\ServerRequestInterface as Request;
 use Psr\Log\LoggerInterface;
 use Psr\Log\LogLevel;
+use Slim\Exception\HttpBadRequestException;
 use Stripe\Exception\ApiErrorException;
 use Stripe\Exception\InvalidRequestException;
 use Stripe\Exception\RateLimitException;
+use Stripe\PaymentIntent;
 use Stripe\StripeClient;
 use Symfony\Component\Serializer\Exception\UnexpectedValueException;
 use Symfony\Component\Serializer\SerializerInterface;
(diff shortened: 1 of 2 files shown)


tests/Application/Actions/Donations/UpdateTest.php
```diff
@@ -1896,4 +1875,82 @@ public function testAddDataAutoconfirmHitsUnknownInvalidRequestException(): void
...
+     * entityManagerProphecy: ObjectProphecy<EntityManagerInterface>,
+     * stripePaymentIntentsProphecy: ObjectProphecy<PaymentIntentService>
+     * }
+     *
+     * @throws \Doctrine\DBAL\Exception\LockWaitTimeoutException
+     * @throws \MatchBot\Application\Matching\TerminalLockException
+     * @throws \MatchBot\Domain\DomainException\DomainLockContentionException
+     * @throws \Stripe\Exception\ApiErrorException
+     */
+    public function setupTestDoublesForConfirmingPaymentFromDonationFunds(string $newPayemtnIntentStatus): array
```
NoelLH, line 1894:
```suggestion
    public function setupTestDoublesForConfirmingPaymentFromDonationFunds(string $newPaymentIntentStatus): array
```
and refs to that var

src/Application/Actions/Donations/Update.php
```diff
@@ -550,4 +553,24 @@ private function handleGeneralStripeError(ApiErrorException $exception, Donation
 
         return null;
     }
+
+    private function ensurePaymentIntentIsGoodOrCancelAndThrow(Donation $donation, PaymentIntent $confirmedPaymentIntent, Request $request): void
```
NoelLH, line 557: I wonder if this var name or a doc block could give a bit more info about what we mean by "good"? In particular, that it's specific to the status of a PI right after `confirm()`ing.

NoelLH (commented):
Couple of minor things inline, looking good!

src/Application/Actions/Donations/Update.php
bdsl: Good question - I think since that isn't so clear it's probably better to move the if condition outside this function. Will refactor.

PR #661, opened 2023-10-31
Fix Error: Typed property MatchBot\Domain\SalesforceProxyRepository::…

bdsl:
…$logger must not be accessed before initialization

The EntityManager does not set the logger in repositories, so when we need a repository we have to take it from the DI container which does inject the logger, not the EM.

### DIFF
--- a/src/Application/Actions/Donations/Create.php
+++ b/src/Application/Actions/Donations/Create.php
@@ -16,7 +16,6 @@ use MatchBot\Application\Auth\PersonWithPasswordAuthMiddleware;
 use MatchBot\Application\HttpModels\DonationCreate;
 use MatchBot\Application\HttpModels\DonationCreatedResponse;
 use MatchBot\Application\Matching\Adapter;
-use MatchBot\Domain\Campaign;
 use MatchBot\Domain\CampaignRepository;
 use MatchBot\Domain\Charity;
 use MatchBot\Domain\DomainException\DomainLockContentionException;
@@ -35,6 +34,7 @@ class Create extends Action
     #[Pure]
     public function __construct(
         private DonationRepository $donationRepository,
+        private CampaignRepository $campaignRepository,
         private EntityManagerInterface $entityManager,
         private SerializerInterface $serializer,
         private StripeClient $stripeClient,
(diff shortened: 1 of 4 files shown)


src/Application/Actions/Donations/Create.php
```diff
@@ -149,9 +149,7 @@ protected function action(Request $request, Response $response, array $args): Re
         if ($donation->getPsp() === 'stripe') {
             if (empty($donation->getCampaign()->getCharity()->getStripeAccountId())) {
                 // Try re-pulling in case charity has very recently onboarded with for Stripe.
-                $repository = $this->entityManager->getRepository(Campaign::class);
```
bdsl, line 152: entityManager doesn't know that our repos need to have a logger

src/Domain/SalesforceProxyRepository.php
```diff
@@ -18,35 +18,17 @@
...
     protected function logError(string $message): void
     {
-        if ($log = $this->getLogger()) {
-            $log->error($message);
-            return;
-        }
-
-        // Fallback if logger not configured
-        echo "ERROR: $message";
+        $this->logger->error($message);
```
bdsl, line 21: the logger property is already typed as Logger so if it isn't there getLogger will just crash with a TypeError it won't return a falsy value. So all this stuff is redundant.

src/Domain/SalesforceProxyRepository.php
```diff
@@ -67,9 +49,4 @@ protected function getClient(): Client\Common
 
         return $this->client;
     }
-
-    protected function getLogger(): ?LoggerInterface
```
bdsl, line 71: property was typed as non-nullable so no reason to make return it as nullable here. And then a protected getter for a protected property is redundant.

tests/TestCase.php
```diff
@@ -27,12 +27,22 @@ class TestCase extends PHPUnitTestCase
...
+     */
+    private array $appInstance = [0 => null, 1 => null];
+
     /**
      * @return App
      * @throws Exception
      */
     protected function getAppInstance(bool $withRealRedis = false): App
     {
+        $memoizedInstance = $this->appInstance[(int)$withRealRedis];
```
bdsl, line 41: Having added a call to getAppInstance in the CreateTest setup there are now some duplicate calls to this function form within the same test. We have to return the same instance in that case.

At the end of each test PHPUnit throws away the test case object so the memoized instance(s) get garbaged.

NoelLH (commented):
Looks good!

PR #665, opened 2023-11-03
MAT-315: Using Stripe CLI, allow connecting local dev environments to real stripe webhooks

bdsl:
…ng display of thanks page

Needs some more work, at the very least adding documentation.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -26,18 +26,44 @@ In advance of the first app run:
   are working with your own Salesforce sandbox you would want to change most of the `SALESFORCE_*`
   variables.
 
+### Set up a connection to the Stripe test environment
+
+In dev environments, we use [Stripe CLI](https://stripe.com/docs/stripe-cli?locale=en-GB) in its own docker container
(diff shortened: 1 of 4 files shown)


Readme.md
```diff
@@ -26,18 +26,44 @@ In advance of the first app run:
   are working with your own Salesforce sandbox you would want to change most of the `SALESFORCE_*`
   variables.
 
+### Set up a connection to the Stripe test environment
```
bdsl, line 29: First time doing this if you have an existing docker setup you will need to rebuild it to bring in the new stripe CLI and its settings. After editing config files run:

```shell
$ docker-compose down
$ docker-compose up --build -d
```

And for testing with this but also generally its helpful to watch the logs - the stripe CLI will print logs as will matchbot. To watch both:

```
$ docker logs matchbot-app-1 --tail 100 --follow &
$ docker logs matchbot-stripe-cli-1 --tail 100 --follow &
```

NoelLH (approved):
Nice! Good that it doesn't require anything bigger that's custom, and the readme update looks great

PR #668, opened 2023-11-03
MAT-322 - Retry donation setup at Action level, and not in EM

NoelLH:


src/Application/Actions/Donations/Create.php
```diff
@@ -33,7 +36,8 @@
 
 class Create extends Action
 {
-    private bool $bypassStripe = false;
+    private const MAX_CREATE_RETRY_COUNT = 4;
+    private bool $bypassStripe;
```
bdsl, line 40: having bypassStripe here seems OK as a temp thing for CC23 but imho we wouldn't want to keep it permanently as it's putting too much complexity into this class - instead we should depend on something that may or may not bypass stripe. Would you agree @NoelLH ?

src/Application/Actions/Donations/Create.php
```diff
@@ -279,4 +281,34 @@ private function removeSpecialChars(string $descriptor): string
...
     }
+
+    /**
+     * It seems like just the *first* persist of a given donation needs to be retry-safe, since there is a small
+     * but non-zero minority of Create attempts at the start of a big campaign which get a closed Entity Manager
+     * and then don't know about the connected #campaign on persist and crash when RetrySafeEntityManager tries again.
+     */
+    private function persistDonationWithRetry(Donation $donation): void
+    {
+        \assert($this->entityManager instanceof RetrySafeEntityManager);
```
bdsl, line 292: any reason not to change the type on the actual property instead of asserting here? Fine if it's just for speed of implementation and we're going to change it later.

src/Application/Actions/Donations/Create.php
```diff
@@ -279,4 +281,34 @@ private function removeSpecialChars(string $descriptor): string
...
+                $this->logger->info(
+                    sprintf(
+                        'Donation Create persist error: %s. Retrying %d of %d.',
+                        $exception->getMessage(),
+                        $retryCount,
+                        self::MAX_CREATE_RETRY_COUNT,
+                    )
+                );
+
+                usleep(random_int(100_000, 1_100_000)); // Wait between 0.1 and 1.1 seconds before retrying
```
bdsl, line 310: So as I understand it the only differences between this retry and the one built into the RetrySafeEntityManager are:
- there's a sleep
- We *do not* replace the underlying entity manager. This means if it's still usable we still have any required related new entities in Unit of Work - but the underlying EM is closed the retry will not work.

Is that right? Maybe worth adding a comment to that effect here?

bdsl (commented):
Added comments inline

bdsl (dismissed):

src/Application/Actions/Donations/Create.php
bdsl: Right I misread this, thought it was a new line here.

src/Application/Actions/Donations/Create.php
NoelLH: No good reason, fixed in https://github.com/thebiggive/matchbot/pull/668/commits/0d135146007c36041ee2401d811034024798d7b5

bdsl (dismissed):

src/Application/Actions/Donations/Create.php
NoelLH: Yes, I think so. I'm still not totally confident about the ability to recover from a lost EM but have tried to combine this explanation with the existing method doc block in https://github.com/thebiggive/matchbot/pull/668/commits/ad6312283e74eca2f85f4bbe33ce0f55d89ed634

src/Application/Actions/Donations/Create.php
bdsl: Looks like that would require setting an entry with `RetrySafeEntityManager::class` as the key in `dependancies.php` and configuring that in tests though. Or alternatively adding `\MatchBot\Application\Actions\Donations\Create` to `dependancies.php` and injecting deps manually to it.

bdsl: ah I see we already have an entry for RetrySafeEntityManager in dependencies.

NoelLH: We already have it in `dependencies.php`, and I've just fixed the test

PR #672, opened 2023-11-09
MAT-326: install infection mutation testing tool

dorota-joanna:


src/Application/SlackChannelChatterFactory.php
```diff
@@ -24,4 +24,19 @@ public function makeChatter(string $channelName): ChatterInterface
...
+     * Copied from https://github.com/Roave/infection-static-analysis-plugin/tree/1.34.x#readme
+     * because I want to see how the plugin is working.
+     *
+     * @psalm-suppress PossiblyUnusedMethod
+     *
+     * @template T
+     * @param array<T> $values
+     * @return list<T>
+     */
+    function makeAList(array $values): array
```
bdsl, line 38: After discussion with people who know more about the plugin than me I found that it only reports mutants as killed by Psalm if:

- they are in code covered by tests and
- the psalm error is located in the same file as the mutation.

This function isn't covered so Psalm sadly doesn't get a chance to kill the mutant. Now that we know that we can delete this function.

src/Application/SlackChannelChatterFactory.php
```diff
@@ -24,4 +24,19 @@ public function makeChatter(string $channelName): ChatterInterface
...
+    {
+        return array_values($values);
+    }
```
bdsl, line 41:
```suggestion
```

dorota-joanna (commented):
LGTM!

PR #674, opened 2023-11-11
MAT-330 - update local copy of fee on Confirm

NoelLH:


src/Application/Actions/Donations/Confirm.php
```diff
@@ -119,6 +119,7 @@ protected function action(Request $request, Response $response, array $args): Re
         // that, would like to find a way to make it so if its left out we get an error instead - either by having
         // derive fees return a value, or making functions like Donation::getCharityFeeGross throw if called before it.
         $this->donationRepository->deriveFees($donation, $cardBrand, $cardCountry);
+        $this->entityManager->persist($donation);
```
bdsl, line 122: AFAIK `persist` is only effective when called on a new entity that the ORM doesn't know about yet. Since $donation here came from the ORM in the first place I think persist does nothing:

https://www.doctrine-project.org/projects/doctrine-orm/en/2.16/reference/working-with-objects.html

> If X is a preexisting managed entity, it is ignored by the persist operation. However, the persist operation is cascaded to entities referenced by X, if the relationships from X to these other entities are mapped with cascade=PERSIST or cascade=ALL (see "[Working with Associations](https://www.doctrine-project.org/projects/doctrine-orm/en/2.16/reference/working-with-associations.html#transitive-persistence)").

The call to `flush` further down should be enough to save any changes to the donation, not sure why it hasn't been working.

src/Application/Actions/Donations/Confirm.php
```diff
@@ -238,9 +238,13 @@ protected function action(Request $request, Response $response, array $args): Re
             ], 500);
         }
 
-        $updatedIntent = $this->stripeClient->paymentIntents->retrieve($paymentIntentId);
-
+        // Assuming Stripe calls worked, commit any changes that `deriveFees()` made to the EM-tracked `$donation`.
         $this->entityManager->flush();
+        $this->entityManager->commit();
```
bdsl, line 243: Looks like adding `commit` will fix the immediate issue. Either here or as a rule in general maybe we should avoid using `beginTransaction` and always use `\Doctrine\ORM\EntityManagerInterface::transactional` instead since that automatically commits if no exception is thrown. Also our own implementation of that, `\MatchBot\Application\Persistence\RetrySafeEntityManager::transactional` auto retries the commit, but we don't have our own implementation of `\Doctrine\ORM\EntityManagerInterface::commit` so that one won't auto retry. It's a bit unintuitive that they behave differently.

NoelLH: I'm not totally sure either way on this. I definitely prefer `transactional()` for simple things with one failure mode, but I think it became a bit unwieldy before in a method where we wanted to have a few distinct exception handle clauses which would each rollback, sometimes conditionally. I'm not sure whether Confirm is simple enough to be in the former category or not.

.circleci/config.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         environment:
           MYSQL_DATABASE: matchbot
           MYSQL_ROOT_PASSWORD: tbgCircle123
-      - image: redis:6.2
+      - image: redis:7.0
```
bdsl, line 34: LGTM if this is matching what we have in prod.

docker-compose.yml
```diff
@@ -22,7 +23,7 @@ services:
       - matchbot
 
   db:
-    image: mysql:8.0
+    image: mariadb:11
```
bdsl, line 26: It's MySQL in prod, isn't it? Can you remind me what version? Would be good to keep dev as closely aligned to prod as we can. Would be good to know also about the reason for changing from MySQL to Maria

docker-compose.yml
```diff
@@ -33,7 +34,8 @@ services:
       - matchbot
 
   redis:
-    image: redis:6.2
+    image: redis:7.0
+    platform: linux/amd64
```
bdsl, line 38: as at https://github.com/thebiggive/matchbot/pull/674/files#r1392423013

integrationTests/CreateDonationTest.php
```diff
@@ -1,19 +1,12 @@
...
-use Prophecy\Argument;
-use Prophecy\Prophecy\ObjectProphecy;
 use Ramsey\Uuid\Uuid;
-use Stripe\PaymentIntent;
-use Stripe\StripeClient;
 
 class CreateDonationTest extends IntegrationTest
 {
-    use \Prophecy\PhpUnit\ProphecyTrait;
-
```
bdsl, line 16: :+1:

integrationTests/CreateDonorAccountTest.php
```diff
@@ -1,21 +1,13 @@
...
-use Prophecy\Prophecy\ObjectProphecy;
-use Ramsey\Uuid\Uuid;
-use Stripe\PaymentIntent;
-use Stripe\StripeClient;
 use MatchBot\Application\Actions\DonorAccount;
 
 class CreateDonorAccountTest extends IntegrationTest
 {
-    use \Prophecy\PhpUnit\ProphecyTrait;
-
```
bdsl, line 18: :+1:

src/Application/Actions/Donations/Confirm.php
```diff
@@ -148,7 +148,7 @@ protected function action(Request $request, Response $response, array $args): Re
         ]);
 
         try {
-            $this->stripeClient->paymentIntents->confirm($paymentIntentId, [
+            $updatedIntent = $this->stripeClient->paymentIntents->confirm($paymentIntentId, [
```
bdsl, line 151: nice, eliminating the call to retrieve this from Stripe later.

bdsl (dismissed):
I tested manually on local and it clearly fixes the bug. Two £100, zero tip donations, one done on develop branch and one on this branch (although without rebuilding docker images of course since both are in the same db)

image

All LGTM other than where I've commented.

docker-compose.yml
bdsl: Could we leave this as mysql 8 for now so we have dev closely aligned to prod? Or if we need to make them different can we add a ticker for bringing them back together soon?

src/Application/Actions/Donations/Confirm.php
bdsl: This commit clearly is the line that fixes the bug.

docker-compose.yml
NoelLH: Done, with the caveat that I've added `platform` – would be good to confirm this doesn't break local runs for you on [x86?] Linux.

bdsl: That platform is the hardware platform I have, so doesn't require any emulation. Also tested and it works fine.

bdsl (approved):

PR #675, opened 2023-11-13
 MAT-330: Delete test that has mysteriously started failing on develop

bdsl:
Actually the mystery might not be so much why it fails but why it
previously passed - even on the very same commit where it now fails.

Two builds of same commit https://github.com/thebiggive/matchbot/commit/eb04052ab8b2a59b596cb290644f7e1705d335cd:
Pass: https://app.circleci.com/pipelines/github/thebiggive/matchbot/2130/workflows/3a5fde3b-a330-42f9-9915-38aee8f5b837/jobs/5833
Failure: https://app.circleci.com/pipelines/github/thebiggive/matchbot/2130/workflows/1ccc6350-33e4-47e0-a594-1aee96f32c6d/jobs/5888

It maybe makes sense that it would fail, as the test was asserting that a bug
existed, and that bug was dealt with https://github.com/thebiggive/matchbot/pull/668

But then I still don't understand how it passes and fails for the same
commit. Is there something cached that affects it? Or something else
varaible that isn't in git?


NoelLH (approved):
I can't answer the 'why' but this seems reasonable for now

PR #677, opened 2023-11-15
MAT-327: Replace stripe client in Update handler with stubbable abstr…

bdsl:
…action

### DIFF
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -152,6 +152,20 @@ return function (ContainerBuilder $containerBuilder) {
             return new Client\Mailer($settings, $c->get(LoggerInterface::class));
         },
 
+        Client\Stripe::class => function (ContainerInterface $c): Client\Stripe {
+
+            $useStubStripe = (getenv('APP_ENV') !== 'production' && getenv('BYPASS_PSP'));
+            $offTheShelfStripeClient = $c->get(StripeClient::class);
+
+            \assert($offTheShelfStripeClient instanceof StripeClient);
+
+            if ($useStubStripe) {
+                return new Client\StubStripeClient();
+            }
+
+            return new Client\LiveStripeClient($offTheShelfStripeClient);
+        },
+
         \MatchBot\Domain\DonationFundsNotifier::class => function (ContainerInterface $c): \MatchBot\Domain\DonationFundsNotifier {
             $mailer = $c->get(Client\Mailer::class);
             \assert($mailer instanceof Client\Mailer);
--- a/src/Application/Actions/Donations/Update.php
+++ b/src/Application/Actions/Donations/Update.php
@@ -11,7 +11,7 @@ use MatchBot\Application\Actions\Action;
 use MatchBot\Application\Actions\ActionError;
 use MatchBot\Application\Actions\ActionPayload;
 use MatchBot\Application\HttpModels;
-use MatchBot\Application\PSPStubber;
+use MatchBot\Client\Stripe;
 use MatchBot\Domain\DomainException\DomainRecordNotFoundException;
 use MatchBot\Domain\Donation;
 use MatchBot\Domain\DonationRepository;
@@ -25,7 +25,6 @@ use Stripe\Exception\ApiErrorException;
 use Stripe\Exception\InvalidRequestException;
 use Stripe\Exception\RateLimitException;
 use Stripe\PaymentIntent;
-use Stripe\StripeClient;
 use Symfony\Component\Serializer\Exception\UnexpectedValueException;
 use Symfony\Component\Serializer\SerializerInterface;
 use TypeError;
@@ -45,7 +44,7 @@ class Update extends Action
         private DonationRepository $donationRepository,
         private EntityManagerInterface $entityManager,
         private SerializerInterface $serializer,
-        private StripeClient $stripeClient,
+        private Stripe $stripe,
         LoggerInterface $logger
     ) {
         parent::__construct($logger);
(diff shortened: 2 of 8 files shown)


src/Application/PSPStubber.php
```diff
@@ -5,6 +5,7 @@
 use Ramsey\Uuid\Uuid;
 
 /**
+ * @deprecated - Will put in a better solution for MAT-327
```
bdsl, line 8: Marking this as deprecated so that any usages get flagged by Psalm. Have added the usages here to the baseline so the PR can be merged but then I can do another PR to remove other usages of this PSPStubber class until it can be deleted.

app/dependencies.php
```diff
@@ -152,6 +152,20 @@ function (ContainerInterface $c): Auth\DonationPublicAuthMiddleware {
             return new Client\Mailer($settings, $c->get(LoggerInterface::class));
         },
 
+        Client\Stripe::class => function (ContainerInterface $c): Client\Stripe {
+
+            $useStubStripe = (getenv('APP_ENV') !== 'production' && getenv('BYPASS_PSP'));
```
bdsl, line 157: For now this is the same logic as we have in PSPStubber that was used for recent load tests. But I think it would make sense to adjust it so it can use a query param and we can have manual QA going on at the same time as load tests.

src/Application/Actions/Donations/Update.php
```diff
@@ -45,7 +44,7 @@ public function __construct(
         private DonationRepository $donationRepository,
         private EntityManagerInterface $entityManager,
         private SerializerInterface $serializer,
-        private StripeClient $stripeClient,
+        private Stripe $stripe,
```
bdsl, line 47: This is now our abstraction of Stripe that is simpler to use than the off the shelf library as well as being sometimes real and sometimes stub depending what we need that day.

src/Client/LiveStripeClient.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        $this->stripeClient->paymentIntents->update($paymentIntentId, $updateData);
+    }
+
+    public function confirmPaymentIntent(string $paymentIntentId): PaymentIntent
+    {
+        return $this->stripeClient->paymentIntents->confirm($paymentIntentId);
+    }
+
+    public function retrievePaymentIntent(string $paymentIntentId): PaymentIntent
```
bdsl, line 36: since I've used it four times already here it's probably worth making a type for PaymentIntentID. Either as a class like `StripePaymentIntentId` or something generic like `StripeId<PaymentIntent>`. Would be good to have thoughts on that from @dorota-joanna and @NoelLH

src/Client/Stripe.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+interface Stripe
+{
+    public function cancelPaymentIntent(string $paymentIntentId): void;
+
+    public function updatePaymentIntent(string $paymentIntentId, array $updateData): void;
+
+    public function confirmPaymentIntent(string $paymentIntentId): PaymentIntent;
+
+    public function retrievePaymentIntent(string $paymentIntentId): PaymentIntent;
```
bdsl, line 24: Since I've used it four times already here it's probably worth making a type for payment intent IDs. Either as a class like `StripePaymentIntentId` or something generic like `StripeId<PaymentIntent>`. Would be good to get @dorota-joanna and @NoelLH 's thoughts on that.

tests/Application/Actions/Donations/UpdateHandlesLockExceptionTest.php
```diff
@@ -9,6 +9,7 @@
 use Doctrine\ORM\EntityManagerInterface;
 use GuzzleHttp\Psr7\ServerRequest;
 use MatchBot\Application\Actions\Donations\Update;
+use MatchBot\Client\Stripe;
 use MatchBot\Domain\Campaign;
 use MatchBot\Domain\Charity;
 use MatchBot\Domain\Donation;
```
bdsl, line 15: this test gets quite a bit simpler as a nice side effect of the work to support load testing.

src/Client/StubStripeClient.php
```diff
@@ -0,0 +1,42 @@
...
+
+use phpDocumentor\Reflection\Types\Never_;
+use Stripe\PaymentIntent;
+
+/**
+ * Does not connect to stripe. For use in load testing to allow testing Matchbot with high traffic levels
+ * without sending all that traffic to Stripe. Test mode stripe does not allow sending high volumes of test traffic
+ * so we have to stub it out.
+ */
+class StubStripeClient implements Stripe
```
bdsl, line 13: and this one perhaps also doesn't need tests of its own as its only for use in testing. Although it might be worth having unit tests to verify that we're don't have a bug that would get in the way of running load tests. If we do need to unit test it then we'll have to get rid of the direct `usleep` call, probably by using the Symfony Clock component again.

src/Client/StubStripeClient.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+namespace MatchBot\Client;
+
+use phpDocumentor\Reflection\Types\Never_;
```
dorota-joanna, line 5:
```suggestion
```

dorota-joanna (approved):
LGTM

src/Client/LiveStripeClient.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public function cancelPaymentIntent(string $paymentIntentId): void
+    {
+        $this->stripeClient->paymentIntents->cancel($paymentIntentId);
+    }
+
+    public function updatePaymentIntent(string $paymentIntentId, array $updateData): void
+    {
+        $this->stripeClient->paymentIntents->update($paymentIntentId, $updateData);
```
NoelLH, line 28: I'm finding it a little counterintuitive that we're not returning the Payment Intent here even though the underlying lib does and we've found that useful in Identity. But as nothing uses it right now, I guess we still slightly benefit from simplifying and not having to think about the return type for the moment.

src/Client/Stripe.php
NoelLH: Yep, think that could be a nice improvement

NoelLH (commented):
Looks good!

src/Client/LiveStripeClient.php
bdsl: Happy to add the return in if you think it's worth it. I guess the advantage of returning void is it's easier to mock. If we type this as returning PaymentIntent then any test double will also have to return a payment intent. Probably that can be as simple as `willReturn($this->createStub(PaymentIntent::class));` but we might think we have to spend time making a more realistic PI object even though it's not used.

NoelLH: Right – probably makes sense to leave it now until we have a real use case.

PR #678, opened 2023-11-16
MAT-326: Stop using Roave/infection-static-analysis-plugin, use infec…

bdsl:
…tion/infection directly instead

I have realised that the Roave plugin is massively overinflating our MSI, making our tests look better than they are, because it thinks Psalm is mutated code when its really just just reporting an issue that already exists in our baseline file.

Infection reports:

Before:
image
https://output.circle-artifacts.com/output/job/29553382-ff6b-4072-992c-614341d5cef8/artifacts/0/reports/infection.html#mutant

----

After:
image
https://output.circle-artifacts.com/output/job/50d5d293-e0cc-4ec8-b4bc-f4b50033d1ba/artifacts/0/reports/infection.html#mutant

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,7 @@ jobs:
     # Working directory must match docker-compose mount / Docker `COPY` for real Doctrine proxy
     # generation-dependent tests to work.
     working_directory: /var/www/html
+    resource_class: large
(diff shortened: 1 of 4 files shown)


dorota-joanna (approved):
LGTM

PR #679, opened 2023-11-16
MAT-327: Use new stubable stripe abstraction in Donation Create action

bdsl:
Commits may be reviewed one by one or together. Uses the new abstraction for stripe everywhere that we have been using the PSPStubber class up to now.


src/Application/Actions/Donations/Confirm.php
```diff
@@ -88,17 +83,10 @@ protected function action(Request $request, Response $response, array $args): Re
...
-            /** @psalm-suppress PropertyTypeCoercion card */
-            $paymentMethod->card = (object)['brand' => 'visa', 'country' => 'GB'];
-        } else {
-            $paymentMethod = $this->stripeClient->paymentMethods->retrieve($pamentMethodId);
-        }
+        $paymentMethod = $this->stripe->retrievePaymentMethod($pamentMethodId);
 
-        if ($paymentMethod->type !== 'card') {
+        $type = $paymentMethod->type;
+        if ($type !== 'card') {
```
bdsl, line 89:
```suggestion
        if ($paymentMethod->type !== 'card') {
```

src/Client/Stripe.php
```diff
@@ -15,11 +17,30 @@
...
+     * @throws ApiErrorException
+     */
     public function retrievePaymentIntent(string $paymentIntentId): PaymentIntent;
+
+    /**
+     * @throws ApiErrorException
+     */
+    public function createPaymentIntent(array $createPayload): PaymentIntent;
+
+    public function retrievePaymentMethod(string $paymentMethodId): PaymentMethod;
```
bdsl, line 45:
```suggestion

    /**
     * @throws ApiErrorException
     */
    public function retrievePaymentMethod(string $paymentMethodId): PaymentMethod;
```

dorota-joanna (approved):
LGTM

src/Client/StubStripeClient.php
```diff
@@ -29,7 +31,7 @@ public function updatePaymentIntent(string $paymentIntentId, array $updateData):
         $this->pause();
     }
 
-    public function confirmPaymentIntent(string $paymentIntentId): Never
+    public function confirmPaymentIntent(string $paymentIntentId, array $params = []): Never
     {
         throw new \Exception("Confirm Payment Intent not implemented in stub - not currently used in load tests");
```
NoelLH, line 36: This seems wrong, I think? The previous logic included inline simulation of card brand etc., and we know that load tests _do_ now simulate the new server-side confirmation approach.

bdsl: Thanks, fixed in later PR.

PR #686, opened 2023-11-20
DON-927 – update `Donation\Create` pspMethodType property name to match other contexts & apps; update API contract

NoelLH:


dorota-joanna (approved):
LGTM

bdsl (approved):

PR #688, opened 2023-11-21
MAT-326: Add some donation model unit tests

bdsl:
Before: https://output.circle-artifacts.com/output/job/7450cbaa-974a-47f0-ba6e-77c3f934e836/artifacts/0/reports/infection.html#mutant/Domain/Donation.php

After (although if this isn't reviewed straight away I might push some more tests into the same branch):
https://output.circle-artifacts.com/output/job/900bc9e3-22f8-4d18-b2c3-c3eb2a07e0bd/artifacts/0/reports/infection.html#mutant/Domain/Donation.php

### DIFF
--- a/tests/Domain/DonationTest.php
+++ b/tests/Domain/DonationTest.php
@@ -23,8 +23,14 @@ class DonationTest extends TestCase
 
     public function testBasicsAsExpectedOnInstantion(): void
     {
-        $donation = $this->getMiminalDonation();
-
+        $donation = Donation::fromApiModel(new DonationCreate(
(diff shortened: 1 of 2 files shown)


tests/Domain/DonationTest.php
```diff
@@ -21,8 +23,14 @@ class DonationTest extends TestCase
 
     public function testBasicsAsExpectedOnInstantion(): void
     {
-        $donation = Donation::emptyTestDonation('1');
```
bdsl, line 24: `emptyTestDonation` is a deprecated method.

PR #690, opened 2023-11-22
Mat 326: More tests for Donation class

bdsl:
I think this is probably enough tests for now in the Donation class. Although there are still quite a few mutants none seem to be in particularly complicated code, and some will naturally be dealt with if we test other classes that depend on Donation.

Only test file changed so no way for this to affect prod directly.

Mutation results as at commit 7abad : https://output.circle-artifacts.com/output/job/c37c9b14-91a2-476e-9804-e8e070b5c380/artifacts/0/reports/infection.html#mutant/Domain/Donation.php

### DIFF
--- a/tests/Domain/DonationTest.php
+++ b/tests/Domain/DonationTest.php
@@ -423,7 +423,7 @@ class DonationTest extends TestCase
     /**
      * @return array<array{0: ?string, 1: string}>
      */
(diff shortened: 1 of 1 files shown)


dorota-joanna (approved):
LGTM, the new tests are passing locally

PR #692, opened 2023-11-23
Mat 331 delete redundant command

bdsl:
Ideally this should wait until the matching infrastructure change is deployed, but it isn't a big problem to deploy them in reverse order - will just generate some log noise.


dorota-joanna (approved):
LGTM

PR #693, opened 2023-11-23
MAT-326: Add unit tests for DonationRepository::allocateMatchFunds

bdsl:
Before: https://output.circle-artifacts.com/output/job/27917661-ff05-4b64-8540-c775fbfdf634/artifacts/0/reports/infection.html#mutant/Domain/DonationRepository.php

After: https://output.circle-artifacts.com/output/job/3de14db4-b4d4-4f7f-9beb-14e232b2d122/artifacts/0/reports/infection.html#mutant/Domain/DonationRepository.php

Comparison:
image

Edit - added more tests in https://github.com/thebiggive/matchbot/pull/693/commits/29d8129f07a4706c924a3f833d80319860528146, not accounted for in above comparison or infection report links.


bdsl:
I wanted to also kill the mutants that relate to the `$allocationTries` and `allocationDone` variables in the code, but I think that may be impossible as it looks like there is some dead code in the function. If I'm right we should delete that code but maybe on the non-CC23 branch.

As far as I can see `$allocationDone` is true any time its read - it's set to true unconditionally before it is read. That means that while loop body can only ever run once, and the values of allocationTries and `$this->maxLockTries` don't matter. Also that the body of the `if (!$allocationDone) {` clause can never be executed.

Would be good to have a 2nd opinion on that in case I missed something.

tests/Domain/DonationRepositoryMatchFundsAllocationTest.php
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+
+    /**
+     * @return list<array{0:string, 1: string, 2: string, 3: string, 4: string, 5: string}>
+     */
+    public function AllocationFromTwoFundingsCases(): array
+    {
+        return [
+            // f0 available, f1 available, donation amount, amount matched, withdrawal0, withdrawal2,
```
bdsl, line 140:
```suggestion
            // f0 available, f1 available, donation amount, amount matched, withdrawal0, withdrawal1,
```

psalm-baseline.xml
```diff
@@ -857,6 +857,9 @@
     <ArgumentTypeCoercion>
       <code>$amount</code>
       <code>$amount</code>
+      <code><![CDATA[$fundingWithdrawal->getAmount()]]></code>
+      <code><![CDATA[$fundingWithdrawal->getAmount()]]></code>
+      <code><![CDATA[$fundingWithdrawal->getAmount()]]></code>
```
bdsl, line 862: These look like new issues but they are really existing issues being made more specific by adding more type info - `MixedAssignment` becomes `ArgumentTypeCoercion` when we unmix stuff.

dorota-joanna (approved):
LGTM

PR #694, opened 2023-11-24
MAT-326: Simplify allocateMatchFunds

bdsl:


src/Domain/DonationRepository.php
```diff
@@ -186,45 +186,32 @@ public function allocateMatchFunds(Donation $donation): string
...
-
-        while (!$allocationDone && $allocationTries < $this->maxLockTries) {
-            try {
-                $likelyAvailableFunds = $this->getEntityManager()
-                    ->getRepository(CampaignFunding::class)
-                    ->getAvailableFundings($donation->getCampaign());
+        try {
+            $likelyAvailableFunds = $this->getEntityManager()
+                ->getRepository(CampaignFunding::class)
+                ->getAvailableFundings($donation->getCampaign());
```
bdsl, line 192: The old version of the code looks like it always executes the loop body exactly once.

NoelLH:
Good find @bdsl!

It looks like this loop's been redundant ever since we [removed the Doctrine matching adapter](https://github.com/thebiggive/matchbot/commit/4748edc8e2e0272cbafac03420fbc43345754163) in 2021.

Since we now use a non-locking optimistic Redis adapter exclusively there is no such thing as a `RetryableLockException` and we already removed the old `catch` block which incremented the attempts count.

I think we can probably tidy up in a similar way for `$releaseTries`, and then delete `DonationRepository::$maxLockTries`.

bdsl:
Thanks @NoelLH - it was just the process of trying to kill mutants led me to it - there was no unit test I could write to distinguish the old version and the new version.

bdsl:
@NoelLH 

> I think we can probably tidy up in a similar way for $releaseTries, and then delete DonationRepository::$maxLockTries.

Done

bdsl:
@NoelLH does this look to merge into `deploy-after-cc23` to you now?

src/Domain/DonationRepository.php
```diff
@@ -179,52 +179,36 @@ public function buildFromApiRequest(DonationCreate $donationData): Donation
...
+            $lockEndTime = microtime(true);
+
+            $this->persistQueuedDonations();
+        } catch (Matching\TerminalLockException $exception) { // Includes non-retryable `DBALException`s
+            $waitTime = round(microtime(true) - $lockStartTime, 6);
+            $this->logError(
+                "Match allocate error: ID {$donation->getUuid()} got " . get_class($exception) .
+                " after {$waitTime}s: {$exception->getMessage()}"
+            );
+            throw $exception; // Re-throw exception after logging the details if not recoverable
```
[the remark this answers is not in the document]
NoelLH: I think the timing info is potentially useful – that might be harder to include "downstream"?

src/Domain/DonationRepository.php
```diff
@@ -297,50 +280,36 @@ public function releaseMatchFunds(Donation $donation): void
...
-            throw new DomainLockContentionException();
+            try {
+                $this->removeAllFundingWithdrawalsForDonation($donation);
+            } catch (DBALException $exception) {
+                $this->logError('Doctrine could not remove withdrawals after maximum tries');
+            }
+        } catch (Matching\TerminalLockException $exception) {
+            $waitTime = round(microtime(true) - $lockStartTime, 6);
+            $this->logError(
+                'Match release FINAL error: ID ' . $donation->getUuid() . ' got ' . get_class($exception) .
```
NoelLH, line 309:
```suggestion
                'Match release error: ID ' . $donation->getUuid() . ' got ' . get_class($exception) .
```

src/Domain/DonationRepository.php
```diff
@@ -179,52 +178,36 @@ public function buildFromApiRequest(DonationCreate $donationData): Donation
...
 
-        if (!$allocationDone) {
-            $this->logger->error(
-                "Match allocate FINAL error: ID {$donation->getUuid()} failed matching after $allocationTries tries"
+            $lockStartTime = microtime(true);
+            $newWithdrawals = $this->matchingAdapter->runTransactionally(
+                fn() => $this->safelyAllocateFunds($donation, $likelyAvailableFunds, $amountMatchedAtStart)
             );
-
-            throw new DomainLockContentionException();
```
NoelLH, line 227: I think we've removed all places we threw `DomainLockContentionException` now. So we can also remove the exception itself and the 1 place we catch it, in `MatchBot\Application\Actions\Donations\Create`.

NoelLH (commented):
Looking good! Only minor things / non-blockers inline.

src/Domain/DonationRepository.php
bdsl: Yeah will look at deleting that on Monday. I wonder if Psalm should have an issue for an unused class. You might have to configure or mark any classes that have usages that Psalm can't detect or that are intended for future use.

src/Domain/DonationRepository.php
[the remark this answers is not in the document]
bdsl: Fair enough. Possibly we could wrap the original exception in a new one that includes the timing info, but I think I'll leave as-is for now.

src/Domain/DonationRepository.php
bdsl: Done in https://github.com/thebiggive/matchbot/pull/694/commits/dd02f5b231a6d35b368116d82753197d254fc4d3

PR #696, opened 2023-11-24
MAT-330 – (re-)set fee on Stripe `charge.succeeded` webhook handler

NoelLH:
In the same request as, and immediately before, we set the status "Collected".

This should guarantee consistent fees regardless of race conditions, as discussed a couple of weeks ago.


NoelLH:
I also gave this a go in `DonationRepository::deriveFees()` – it's the other thing I had a note of that we discussed when looking at the fees.

```
if (in_array($donation->getDonationStatus(), DonationStatus::SUCCESS_STATUSES, true)) {
    throw new \LogicException('Cannot derive fees for a donation previously finalised');
}
```

However we have a handful of unit tests trying to do what this prohibits, so it's a bit more involved than I hoped to add the restriction. It also doesn't directly tighten up the logic as it relates to the world outside MatchBot, whereas the PR change potentially does – it only guards against future code errors.

src/Application/Actions/Hooks/StripePaymentsUpdate.php
```diff
@@ -125,6 +127,16 @@ private function handleChargeSucceeded(Event $event, Response $response): Respon
...
+             */
+            $card = $charge->payment_method_details?->card;
+            if ($card) {
+                /** @psalm-var value-of<Calculator::STRIPE_CARD_BRANDS> $brand */
+                $brand = $card->brand;
+                $this->donationRepository->deriveFees($donation, $brand, $card->country);
+            }
+
             $donation->setDonationStatus(DonationStatus::Collected);
             $donation->setCollectedAt(new \DateTimeImmutable("@{$charge->created}"));
```
bdsl, line 141: If this is for after CC23 so we're not trying to do it in a rush, I wonder if it would be worth going to the more rich domain model style and replacing all this, except getting the card from the stripe object, with a single function call on the Donation object. And the donation object could enforce the rule against setting any of them separately.

DonationRepository::deriveFees could be called from inside Donation, or imho even better it could be moved into donation. The only `$this` in deriveFees is to pass settings to the Calculator, but I think we'd be better off hard-coding all the settings inside the calculator, as changing settings isn't really any easier than editing the code, and we haven't tested the calculator with alternate settings anyway.

bdsl: (possibly isn't clear from the way github shows the comments what I meant by "all this". I mean all these statements that mutate the donation object:

```
$donation->setDonationStatus(DonationStatus::Collected);
$donation->setCollectedAt(new \DateTimeImmutable("@{$charge->created}"));
$this->donationRepository->deriveFees($donation, $brand, $card->country);
$donation->setTransferId($charge->transfer);
$donation->setChargeId($charge->id);
```

IMHO it's much better if you only have to make a single function call to mutate an object, and the object class can can enforce invariants before returning - can't have collectedAt without Collected status and vice versa, can't have Collected status without having calculated fees according the payment type etc.

bdsl:
> However we have a handful of unit tests trying to do what this prohibits, so it's a bit more involved than I hoped to add the restriction.

One thing we could do in that case is log an error, rather than throw an exception. Could even throw an exception unless an env var is set that shows we're running unit tests.

src/Application/Actions/Hooks/StripePaymentsUpdate.php
bdsl: See https://github.com/thebiggive/matchbot/pull/730

src/Domain/Donation.php
```diff
@@ -1321,4 +1316,56 @@ public function cancel(): void
...
+        Assertion::numeric($originalFeeFractional);
+        Assertion::notEmpty($chargeId);
+        Assertion::notEmpty($transferId);
+
+        $this->chargeId = $chargeId;
+        $this->transferId = $transferId;
+
+        if ($cardBrand) {
+            /** @psalm-var value-of<Calculator::STRIPE_CARD_BRANDS> $cardBrand */
+            $this->deriveFees($cardBrand, $cardCountry);
```
bdsl, line 1364: this new call to `deriveFees` is the nugget of the PR.  This function `collectFromStripeCharge` is called from the `\MatchBot\Application\Actions\Hooks\StripePaymentsUpdate` handler referenced in the PR title.

tests/Domain/DonationRepositoryTest.php
```diff
@@ -308,18 +306,19 @@ public function testStripeAmountForCharityWithTipUsingAmex(): void
...
-        // Total fee        = £ 31.80
+        // Total fee ex vat = £ 31.80
+        // Total fee inc vat = £ 31.80 * 1.2
+        // Total fee inc vat = £ 38.16
         // Amount after fee = £955.85
 
         // Deduct tip + fee.
-        $this->assertEquals(4_180, $donation->getAmountToDeductFractional());
-        $this->assertEquals(95_585, $donation->getAmountForCharityFractional());
+        $this->assertEquals(48_16, $donation->getAmountToDeductFractional());
```
bdsl, line 320: total deduction £41.80 to £48.16 here, only because in the old version of the code we didn't apply UK VAT to the fee as it was settings dependant, so applied in prod but not in this test. Now it's applied here, adding £6.36

There's also a formatting change, `48_16` is meant to be a more intuitive way to write a figure in pounds and pence than the version with the underscore before three digits.

tests/Application/Actions/Donations/ConfirmTest.php
```diff
@@ -201,8 +206,6 @@ public function testItReturns500OnApiError(): void
 
         // in reality the fee would be calculated according to details of the card etc. The Calculator class is
         //tested separately. This is just a dummy value.
```
dorota-joanna, line 208:
```suggestion
```

dorota-joanna (approved):
LGTM

PR #697, opened 2023-11-24
MAT-310 – lint config fix; style fixes WIP

NoelLH:


bdsl:
There are a lot of modifications here that won't be very interesting when we look at history. Looks like good changes, but can we add the big commit(s) to a `.git-blame-ignore-revs` file so that it won't show up as the last change for all the lines it touches? Apparently github respects that file automatically, git CLI on local can be configured to respect it.

NoelLH:
Closing in favour of [746](https://github.com/thebiggive/matchbot/pull/746)

PR #703, opened 2023-11-29
MAT-336 – improve donation Create edge case recovery

NoelLH:
Run match fund release in a new "matching adapter transaction". And also log the original \Throwable message earlier, separately, just in case.

### DIFF
--- a/src/Application/Actions/Donations/Create.php
+++ b/src/Application/Actions/Donations/Create.php
@@ -141,7 +141,12 @@ class Create extends Action
 
                 return $this->respond($response, new ActionPayload(503, null, $error));
             } catch (\Throwable $t) {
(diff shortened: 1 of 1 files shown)


bdsl (approved):
LGTM, but see also comments on ticket. I'm not sure the runTransactionally system is very helpful now. All it's actually doing I think is ensuring `\MatchBot\Application\Matching\OptimisticRedisAdapter::saveFundingsToDatabase` is called after anything else we ask the adapter do (but not if the adapter throws), so it would be simpler to just directly call that whenever we need it. Maybe something to change after CC?

NoelLH:
Agreed on all that – I'll make an issue to look at simplifying right after CC23.

PR #706, opened 2023-11-30
MAT-338 – add temporary out of sync funds mode, '2023-11-30-patch'

NoelLH:


src/Application/Commands/HandleOutOfSyncFunds.php
```diff
@@ -74,8 +74,17 @@ protected function doExecute(InputInterface $input, OutputInterface $output): in
         $numFundingsCorrect = 0;
         $numFundingsOvermatched = 0;
         $numFundingsUndermatched = 0;
+
+        $problemIds = [];
         /** @var CampaignFunding[] $fundings */
-        $fundings = $this->campaignFundingRepository->findAll();
+        if ($mode === '2023-11-30-patch') {
+            $problemIds = [23790, 23791];
```
bdsl, line 81: I double checked that these are the funds reported as having wrong data.

src/Application/Commands/HandleOutOfSyncFunds.php
```diff
@@ -117,7 +126,7 @@ protected function doExecute(InputInterface $input, OutputInterface $output): in
 
             $output->writeln("Funding {$funding->getId()} is under-matched by $undermatchAmount. $details");
 
-            if ($mode === 'fix') {
+            if ($mode === 'fix' || ($mode === '2023-11-30-patch' && in_array($funding->getId(), $problemIds, true))) {
```
bdsl, line 129: the in_array check here shouldn't be necessary but is just a nice defense in depth in case `findBy` did something we didn't expect right?

bdsl (approved):
LGTM

src/Application/Commands/HandleOutOfSyncFunds.php
NoelLH: Yep

PR #709, opened 2023-12-01
Add "Your card does not support this type of purchase." to expected p…

bdsl:
…ayment messages

See https://support.stripe.com/questions/your-card-does-not-support-this-type-of-purchase-error-when-processing-payment?locale=en-GB


dorota-joanna (approved):
LGTM

PR #710, opened 2023-12-01
MAT-340 – add detail to a Stripe Payment Method before confirming it

NoelLH:


src/Application/Actions/Donations/Confirm.php
```diff
@@ -83,7 +83,7 @@ protected function action(Request $request, Response $response, array $args): Re
         }
 
         $paymentIntentId = $donation->getTransactionId();
-        $paymentMethod = $this->stripe->retrievePaymentMethod($pamentMethodId);
+        $paymentMethod = $this->stripe->updatePaymentMethodBillingDetail($pamentMethodId, $donation);
```
bdsl, line 86: I've got a bit of a misgiving about this. We're setting the name and email as details of the payment method in stripe, but we weren't given them as payment method details - we were given them as details of the donation. 

It's quite possible that a donor will have a different name and email attached to their card than what they gave us on the donation, or indeed that they'll be using another person's card with permission for the payment.

bdsl: Also if this is intended primarily to ease support tasks at Big Give then it seems better to let more people look into the data already held inside the Matchbot Database, on the Donation record, rather than copying it to something else in a different place, that isn't the Donation, just in case it needs to be seen.

bdsl (approved):
I'm not sure. Approving as the code looks OK, but see comments inline.

src/Application/Actions/Donations/Confirm.php
bdsl: Right, but isn't it attached to the customer record rather than the payment method record in the case of a completed donation? I think that's done by identity server at https://github.com/thebiggive/identity/blob/f546fabdc7f042abb68276932d0a4d39ecaaf91b/src/Application/Actions/Person/Update.php#L205

Not blocking

bdsl: And yes you're right we do already set the donations to be pushed to SF immediately on creation. I think even if they don't have a name or email address: https://github.com/thebiggive/matchbot/blob/1b70e860ea4ffc327c059e8bd6a1104222bb3281/src/Application/Actions/Donations/Create.php#L257

NoelLH: Yep, it is. But as Stripe end up linking the two objects anyway, I don't know how big a practical difference it makes.

PR #712, opened 2023-12-01
Revert "MAT-340 – add detail to a Stripe Payment Method before confirming it"

NoelLH:
Reverts thebiggive/matchbot#710


bdsl:
While investigating MAT-339 I found that this change leads to a stripe exception being thrown from somewhere where we're not catching it or displaying it the donor.

PR #713, opened 2023-12-02
Use psalm plugin psr container

bdsl:
Simplifies our code and enhances the ability of Psalm to catch errors.

### DIFF
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -109,9 +109,8 @@ return function (ContainerBuilder $containerBuilder) {
              */
             $settings = $c->get('settings');
             $stripeChannel = $settings['notifier']['slack']['stripe_channel'];
-            /** @var SlackChannelChatterFactory $chatterFactory */
-            $chatterFactory = $c->get(SlackChannelChatterFactory::class);
-            return $chatterFactory->makeChatter($stripeChannel);
+
+            return $c->get(SlackChannelChatterFactory::class)->makeChatter($stripeChannel);
         },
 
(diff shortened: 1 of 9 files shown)


app/dependencies.php
```diff
@@ -109,7 +109,7 @@ function (ContainerInterface $c): Auth\DonationPublicAuthMiddleware {
              */
             $settings = $c->get('settings');
             $stripeChannel = $settings['notifier']['slack']['stripe_channel'];
-            /** @var SlackChannelChatterFactory $chatterFactory */
+
```
bdsl, line 112: Things like this are no longer needed. With https://github.com/Lctrs/psalm-psr-container-plugin Psalm automatically knows that `$chatterFactory` is an instance of `SlackChannelChatterFactory`.

app/repositories.php
```diff
@@ -27,23 +27,34 @@
 return static function (ContainerBuilder $containerBuilder) {
     $containerBuilder->addDefinitions([
         CampaignFundingRepository::class => static function (ContainerInterface $c): CampaignFundingRepository {
-            return $c->get(EntityManagerInterface::class)->getRepository(CampaignFunding::class);
+            $repository = $c->get(EntityManagerInterface::class)->getRepository(CampaignFunding::class);
```
bdsl, line 30: but here all Psalm knows is that repository is an instance of `EntityManagerInterface<CampaignFunding>`, so if we need it to be more specific than that we have to assert it.

NoelLH (approved):
Nice!

PR #716, opened 2023-12-04
Stop a `Confirm` early if we already know the donation's cancelled

NoelLH:


tests/Application/Actions/Donations/ConfirmTest.php
```diff
@@ -82,6 +79,52 @@ public function test_it_confirms_a_card_donation(): void
         );
     }
 
+    public function testItReturns400OnCancelledDonation(): void
+    {
```
bdsl, line 83: It looks like quite a lot of this is duplicating code that we already have in the same test file - IMHO we should refactor to remove a lot of the duplication before merging - it gets very hard to change the API of the SUT when it's coupled separately to a lot of different tests and they all need updating.

bdsl (commented):
Added comment inline, otherwise all LGTM.

dorota-joanna (dismissed):
LGTM

tests/Application/Actions/Donations/ConfirmTest.php
NoelLH: This hopefully helps a bit @bdsl https://github.com/thebiggive/matchbot/pull/716/commits/3f7b4759d2de04171ec5bf953a309dbee02115d0

I think the biggest copied chunk is input to `fakeStripeClient(...)`, but you're right that a lot of this is repeated and some of it could be anything, so I've tried to capture just the essential difference vs. `test_it_confirms_a_card_donation()`

bdsl (approved):
LGTM

PR #721, opened 2023-12-05
Update dev & CI Redis to 7.0 [as there's no Docker tagged 7.1]

NoelLH:


NoelLH:
Redis seem to have ended up with an 'interesting' release pattern where 7.0 is stable but 'older', 7.2 is stable, and 7.1 tags have been pulled; yet 7.1 is the newest AWS ElastiCache version and is being recommended by AWS as a performance upgrade.

It's probably safe enough to use 7.0 for dev for now.

bdsl (approved):

PR #726, opened 2023-12-08
MAT-340 – add detail to a Stripe Payment Method before confirming it

NoelLH:
With more edge case considerations than #710

### DIFF
--- a/src/Application/Actions/Donations/Confirm.php
+++ b/src/Application/Actions/Donations/Confirm.php
@@ -92,7 +92,7 @@ class Confirm extends Action
         }
 
         $paymentIntentId = $donation->getTransactionId();
-        $paymentMethod = $this->stripe->retrievePaymentMethod($paymentMethodId);
+        $paymentMethod = $this->stripe->updatePaymentMethodBillingDetail($paymentMethodId, $donation);
 
         if ($paymentMethod->type !== 'card') {
             throw new HttpBadRequestException($request, 'Confirm endpoint only supports card payments for now');
--- a/src/Application/Actions/Donations/Confirm.php
+++ b/src/Application/Actions/Donations/Confirm.php
@@ -96,6 +96,8 @@ class Confirm extends Action
         try {
             $paymentMethod = $this->stripe->updatePaymentMethodBillingDetail($paymentMethodId, $donation);
         } catch (CardException $cardException) {
+            $this->entityManager->rollback();
+
             return $this->handleCardException(
                 context: 'updatePaymentMethodBillingDetail',
                 exception: $cardException,
@@ -113,12 +115,7 @@ class Confirm extends Action
 
             $this->entityManager->rollback();
 
-            return new JsonResponse([
-                'error' => [
-                    'message' => $exception->getMessage(),
-                    'code' => $exception->getStripeCode(),
-                ],
-            ], 500);
+            throw $exception;
         }
 
         if ($paymentMethod->type !== 'card') {
@@ -163,6 +160,8 @@ class Confirm extends Action
                 'payment_method' => $paymentMethodId,
             ]);
         } catch (CardException $exception) {
+            $this->entityManager->rollback();
+
             return $this->handleCardException(
                 context: 'confirmPaymentIntent',
                 exception: $exception,
@@ -275,8 +274,6 @@ class Confirm extends Action
             $paymentIntentId,
             $exception->getMessage(),
         ));
-
-        $this->entityManager->rollback();
 
         return new JsonResponse([
             'error' => [
--- a/src/Application/Actions/Donations/Confirm.php
+++ b/src/Application/Actions/Donations/Confirm.php
@@ -8,6 +8,7 @@ use MatchBot\Application\Actions\Action;
 use MatchBot\Application\Fees\Calculator;
 use MatchBot\Client\NotFoundException;
 use MatchBot\Client\Stripe;
+use MatchBot\Domain\Donation;
 use MatchBot\Domain\DonationRepository;
 use MatchBot\Domain\DonationStatus;
 use Psr\Http\Message\ResponseInterface as Response;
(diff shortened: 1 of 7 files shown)


src/Application/Actions/Donations/Confirm.php
```diff
@@ -92,7 +93,33 @@ protected function action(Request $request, Response $response, array $args): Re
...
+            ));
+
+            $this->entityManager->rollback();
+
+            return new JsonResponse([
+                'error' => [
+                    'message' => $exception->getMessage(),
+                    'code' => $exception->getStripeCode(),
+                ],
+            ], 500);
```
bdsl, line 121: Logging the detail is useful, but instead of rolling back and returning we could probably just re-throw the exception, which seems a bit simpler and afaik effectively does the same thing. Not a blocker.

src/Application/Actions/Donations/Confirm.php
```diff
@@ -244,4 +259,31 @@ public function randomString(): string
...
+        $this->logger->info(sprintf(
+            'Stripe %s on Confirm %s for donation %s (%s): %s',
+            $exceptionClass,
+            $context,
+            $donation->getUuid(),
+            $paymentIntentId,
+            $exception->getMessage(),
+        ));
+
+        $this->entityManager->rollback();
```
bdsl, line 279: I think I'd move the rollback call outside, e.g.:

```
$this->entityManager->rollback();
 return $this->handleCardException(
            $this->logger->info(sprintf(
                context: 'confirmPaymentIntent',
                'Stripe %s on Confirm for donation %s (%s): %s',
                exception: $exception,
                $exceptionClass,
                donation: $donation,
                $donation->getUuid(),
                paymentIntentId: $paymentIntentId,
                $paymentIntentId,
            );
```

It's probably a good rule of thumb that `beginTransaction`  and `rollback` should both be called from the same function.

src/Client/LiveStripeClient.php
```diff
@@ -44,8 +42,21 @@ public function createPaymentIntent(array $createPayload): PaymentIntent
...
+        // "A PaymentMethod must be attached a customer to be updated."
+        $this->stripeClient->paymentMethods->attach($paymentMethodId, ['customer' => $donation->getPspCustomerId()]);
+
+        // Address etc. is set up in Stripe.js already. Adding these values which we collect on the
+        // donation separately helps with support queries and maybe with fraud signals.
+        return $this->stripeClient->paymentMethods->update(
+            $paymentMethodId,
+            [
+                'billing_details' => [
+                    'name' => "{$donation->getDonorFirstName()} {$donation->getDonorLastName()}",
```
bdsl, line 56: I think it's good if we can keep this LiveStripeClient class as stupid as possible, so it doesn't need testing, since it's coupled to the Stripe API and a bit awkward to test. So I'd suggest even concatenating the name parts is too much to do here. Can we possibly change the function signature and do one of these:

```suggestion
                    'name' => $donation->getDonorFullName()
```

or 

```suggestion
                    'name' => $fullName // param of the present function
```

or

```suggestion
                    'name' => $donorName->value, // new value object class
```

Or maybe

```
     return $this->stripeClient->paymentMethods->update(
            $paymentMethodId,
            [
                'billing_details' => [
                    'name' => $billingDetails->name,
                    'email' => $billingDetails->email,
                ],
            ],
        );

File: src/Client/LiveStripeClient.php [hunk 4]
```diff
@@ -44,8 +42,21 @@ public function createPaymentIntent(array $createPayload): PaymentIntent
         return $this->stripeClient->paymentIntents->create($createPayload);
     }
 
-    public function retrievePaymentMethod(string $paymentMethodId): PaymentMethod
+    public function updatePaymentMethodBillingDetail(string $paymentMethodId, Donation $donation): PaymentMethod
     {
-        return $this->stripeClient->paymentMethods->retrieve($paymentMethodId);
+        // "A PaymentMethod must be attached a customer to be updated."
+        $this->stripeClient->paymentMethods->attach($paymentMethodId, ['customer' => $donation->getPspCustomerId()]);
```
bdsl, line 48: Have you confirmed that it's OK to attach the same method to the same customer multiple times? Since the function name doesn't indicated that it will attach the method we shouldn't expect the caller to only call it once.

bdsl (commented):
Added comments inline, otherwise all LGTM

src/Client/LiveStripeClient.php
NoelLH: I just tried it in test mode with `pm_card_visa` – seems to be allowed and succeed. Hopefully real payment methods will be the same.

src/Application/Actions/Donations/Confirm.php
NoelLH: https://github.com/thebiggive/matchbot/pull/726/commits/3ab0d9aa83521c10db85209839a48587ba9b1d95

src/Application/Actions/Donations/Confirm.php
NoelLH: https://github.com/thebiggive/matchbot/pull/726/commits/3ab0d9aa83521c10db85209839a48587ba9b1d95

src/Client/LiveStripeClient.php
NoelLH: Makes sense. I just went with a helper on the domain model for now to improve this reasonably quickly, as there are a few PRs in flight atm. https://github.com/thebiggive/matchbot/pull/726/commits/1f659d60a8536373f80029fc3a6875890e71d7d6

bdsl (approved):
LGTM

bdsl:
Although also probably worth adding the test coverage to kill these mutants? It is a relatively simple method but also a quite fundamental one:

https://output.circle-artifacts.com/output/job/22d12111-b79c-4d62-a740-16c87a974410/artifacts/0/reports/infection.html#mutant/Domain/Donation.php
image

NoelLH:
`getDonorFullName()` is mutant free with the newly added test, so will merge.

PR #730, opened 2023-12-11
MAT-330: Hard code fee calculator settings & other refactorings

bdsl:
These settings don't change frequently, and the code will become simpler if we can hard code them instead of passing them in to the calculator.

If we are passing them in then we really should be running unit tests with other possible settings, but that doesn't seem like a good use of time.

See commit messages for other changes - all refactoring


src/Application/Fees/Calculator.php
```diff
@@ -26,11 +27,62 @@ class Calculator
...
+            'NOK' => '1.8',
+            'SEK' => '1.8',
+            'USD' => '0.3',
+            'default' => '0.2',
+        ],
+        'gift_aid_percentage' => '0.75', // 3% of Gift Aid amount.
+        'main_percentage_standard' => '1.5',
+        'main_percentage_amex_or_non_uk_eu' => '3.2',
+        // The rate at which VAT is either being or is about to be charged.
+        'vat_percentage_live' => '20',
```
JayBazuzi, line 46: Rename to make the comment redundant, then delete the redundant comment.
```suggestion
        'rate_at_which_VAT_is_either_being_or_is_about_to_be_charged' => '20',
```

bdsl: Thanks for taking a look @JayBazuzi . This section of code is moved unchanged from settings.php in this PR, and as there isn't a business requirement right now to change how it works (only to change *when* it runs) I don't want to take time to edit inside this code much now. This PR is more about changing where this code sits in relation to the rest of the matchbot system.

bdsl: If we were going to edit it we'd just remove the "about to" thing, since it's a reference to a change that happened in the past.

tests/Application/Actions/Hooks/StripePaymentsUpdateTest.php
```diff
@@ -116,10 +116,6 @@ public function testSuccessfulPayment(): void
 
         $donationRepoProphecy = $this->prophesize(DonationRepository::class);
 
-        $donationRepoProphecy
```
NoelLH, line 119: 🥳

src/Domain/Donation.php
```diff
@@ -1319,4 +1313,56 @@ public function cancel(): void
...
+            $this->getCurrencyCode(),
+            $incursGiftAidFee,
+            $this->getCampaign()->getFeePercentage(),
+        );
+
+        $this->setCharityFee($fees->coreFee);
+        $this->setCharityFeeVat($fees->feeVat);
+    }
+
+    public function collectFromStripeCharge(
```
NoelLH, line 1343: Nice!

NoelLH (approved):
I only had time to focus on the bits close to my previous change & close to the Domain today, but looks really good to me so far!

PR #731, opened 2023-12-13
BG2-2520: Run tests in PHP 8.2 as well as 8.1

bdsl:
This isn't expected to pass until after https://github.com/thebiggive/docker-php/pull/18 is merged and run.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,8 +13,12 @@ jobs:
     # generation-dependent tests to work.
     working_directory: /var/www/html
     resource_class: large
+    parameters:
+      php_version:
+        type: string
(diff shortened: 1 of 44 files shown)


dorota-joanna (approved):
LGTM

NoelLH:
`laminas/laminas-code`, a downstream dependency of `doctrine/migrations`, wants PHP < 8.3 in the current locked version – but its newest version supports 8.3

Since https://github.com/thebiggive/matchbot/pull/731/commits/2f9689c5881c42e9b5b7bf5c794e035f771ce5a7 I think there's a chance it's not possible to support all 3 versions with the same set of package pins.

PR #732, opened 2023-12-13
CLA-35: migration to update giftaid on a donation

dorota-joanna:


bdsl (approved):
LGTM, I checked in the DB and confirmed that the UUID and transactionID in your code do relate to the donation linked in the Jira ticket, and the amount also matches.

bdsl:
And there is currently no difference between matchbot dev version and matchbot main version, so if this is merged into develop there's nothing else to stop us merging that to main as soon as we're ready to apply the change to the prod database.

PR #736, opened 2023-12-15
MAT-335 – test amended special behaviour for Big Give general donations

NoelLH:
i.e. donation fund top-up pseudo-tips

### DIFF
--- a/tests/Application/DonationTestDataTrait.php
+++ b/tests/Application/DonationTestDataTrait.php
@@ -2,12 +2,14 @@
 
 namespace MatchBot\Tests\Application;
 
+use MatchBot\Application\HttpModels\DonationCreate;
 use MatchBot\Domain\Campaign;
 use MatchBot\Domain\Charity;
 use MatchBot\Domain\Donation;
 use MatchBot\Domain\DonationStatus;
 use MatchBot\Domain\PaymentMethodType;
 use MatchBot\Domain\SalesforceWriteProxy;
+use MatchBot\Tests\TestCase;
 use Ramsey\Uuid\Uuid;
 
 trait DonationTestDataTrait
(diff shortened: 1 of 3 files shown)


bdsl:
To remind myself, this is adding tests to show the effect of the changes made in https://github.com/thebiggive/matchbot/pull/700

bdsl:
From mutation testing it looks like almost everything is covered in the relavent code but there's one condition left off, so a `LogicalAnd` mutant survived:

https://output.circle-artifacts.com/output/job/22ae1982-4c0d-4631-9580-e504f4b24f08/artifacts/0/reports/infection.html#mutant/Application/Actions/Donations/Update.php
image

Probably worth adding test coverage for that case.

tests/Application/DonationTestDataTrait.php
```diff
@@ -93,4 +116,23 @@ protected function getAnonymousPendingTestDonation(): Donation
...
     }
+
+    private function getCharity(): Charity
+    {
+        $charity = TestCase::someCharity();
+        $charity->setSalesforceId('123CharityId');
+        $charity->setName('Test charity');
+        $charity->setStripeAccountId('unitTest_stripeAccount_123');
+
+        return $charity;
```
bdsl, line 127: from the names it's not really clear to me what the difference in intent is between this `getCharity` and `TestCase::someCharity` - they both return charities with some arbitrary test data.

If it's just that the data needs to be different maybe clearer to decouple them enitely?

```suggestion
        return new Charity(
            salesforceId: ''123CharityId,
            charityName: "Test charity",
            stripeAccountId: 'unitTest_stripeAccount_123',
            hmrcReferenceNumber: 'H' . self::randomHex(3),
            giftAidOnboardingStatus: 'Onboarded',
            regulator: 'CCEW',
            regulatorNumber: 'Reg-no',
            time: new \DateTime('2023-10-06T18:51:27'),
        );
```

Also because eventually I'd like to lose any setter / mutator methods for things that we don't actually change in prod - like a charity's salesforceId never changes so I'd rather not have a setter for it. The less calls there are to that setter the easier it will be to remove.

bdsl (commented):
Added comments inline, otherwise all LGTM.

tests/Application/DonationTestDataTrait.php
NoelLH: I don't think this needed any variation, just got muddled. Fixed on https://github.com/thebiggive/matchbot/pull/736/commits/8186b1591518f842ea771fbd58e1ec234a6e618e

bdsl:
Relavent page of mutation report for last commit, 141a314 : https://output.circle-artifacts.com/output/job/dce7dc6a-59a6-4c61-acad-7e5f13c2dea2/artifacts/0/reports/infection.html#mutant/Application/Actions/Donations/Update.php

bdsl (approved):
LGTM

PR #737, opened 2023-12-15
MAT-342 – prepare a match funds redistribution script

NoelLH:


src/Application/Commands/RedistributeMatchFunds.php
```diff
@@ -0,0 +1,87 @@
...
+    protected function configure(): void
+    {
+        $this->setDescription('Moves match funding allocations from lower to higher priority funds where possible');
+    }
+
+    protected function doExecute(InputInterface $input, OutputInterface $output): int
+    {
+        // TODO Change the fixed lookback to 2 days, or parameter-ise it, once CC23 is tidied in Prod.
+        $donationsToCheck = $this->donationRepository->findWithMatchingWhichCouldBeReplacedWithHigherPriorityAllocation(
+            new \DateTimeImmutable('-3 weeks')
```
bdsl, line 36: Using ambient context to get current time makes things hard to test - and I see here the test is only specifying that the argument should be of type DateTimeImmutable, not equal to any particular date.

Could we inject either a clock or a timestamp to make it more testable? Something like this, assuming the current is inject as a DateTimeImmutable `$this->now`

```suggestion
            $this->now->sub(new \DateInterval('P3W'))
```

Probably then the current date should come from the PSR11 container but I don't have a strong feeling about that vs it coming from `matchbot-cli.php`

src/Application/Commands/RedistributeMatchFunds.php
```diff
@@ -0,0 +1,87 @@
...
+        );
+
+        $donationsAmended = 0;
+        foreach ($donationsToCheck as $donation) {
+            $highestAllocationOrderUsedForDonation = 0;
+            foreach ($donation->getFundingWithdrawals() as $withdrawal) {
+                $funding = $withdrawal->getCampaignFunding();
+                if (!$funding) {
+                    // Not entirely sure why this is nullable at all; we can't work on withdrawals without a funding.
```
bdsl, line 45: There are no cases in the prod database where it is set to null: `SELECT FundingWithdrawal.campaignFunding_id FROM FundingWithdrawal WHERE campaignFunding_id is null; ` returns empty. 

So can we make it non-nullable instead of adding code to deal with the case of it being null? There's only one call to `$withdrawal = new FundingWithdrawal();` in prod code, and it has a call to `$withdrawal->setCampaignFunding($funding);` just after so it looks like we could easily make it a required constructor param.

bdsl: Although probably worth looking into the reason it's nullable - it's this commit: https://github.com/thebiggive/matchbot/commit/a9d71f4eed239c84dd2082eb2ddb410817fc7050

src/Application/Commands/RedistributeMatchFunds.php
```diff
@@ -0,0 +1,87 @@
...
+    protected function doExecute(InputInterface $input, OutputInterface $output): int
+    {
+        // TODO Change the fixed lookback to 2 days, or parameter-ise it, once CC23 is tidied in Prod.
+        $donationsToCheck = $this->donationRepository->findWithMatchingWhichCouldBeReplacedWithHigherPriorityAllocation(
+            new \DateTimeImmutable('-3 weeks')
+        );
+
+        $donationsAmended = 0;
+        foreach ($donationsToCheck as $donation) {
+            $highestAllocationOrderUsedForDonation = 0;
```
bdsl, line 41: AFAIK there's no rule that says ranks have to be positive, so when we haven't seen any we should start at INT_MIN. Or maybe clearer to start at null and give that special handling inside the loop.

```suggestion
            $highestAllocationOrderUsedForDonation = \PHP_INT_MIN;
```

bdsl: Alternatively `allocationOrder` could be a good candidate to turn into a value object.

src/Application/Commands/RedistributeMatchFunds.php
```diff
@@ -0,0 +1,87 @@
...
+            $fundings = $this->campaignFundingRepository->getAvailableFundings($donation->getCampaign());
+
+            $fundingsAllowForRedistribution = false;
+            foreach ($fundings as $funding) {
+                if ($funding->getAllocationOrder() <= $highestAllocationOrderUsedForDonation) {
+                    continue;
+                }
+
+                $amountToAllocate = min($donation->getAmount(), $funding->getAmountAvailable());
+                if ($amountToAllocate <= 0) {
```
bdsl, line 64: I'm confused - donation amount is never zero or less is it, so isn't this just testing whether the funding amount available is zero or less?

Was originally thinking to just suggest using BC math but then got more confused.

src/Application/Commands/RedistributeMatchFunds.php
```diff
@@ -0,0 +1,87 @@
...
+                $fundingsAllowForRedistribution = true;
+                break;
+            }
+
+            if (!$fundingsAllowForRedistribution) {
+                continue;
+            }
+
+            $this->donationRepository->releaseMatchFunds($donation);
+            $this->donationRepository->allocateMatchFunds($donation);
```
bdsl, line 77: Two questions about this:

1. Is it possible for match funds to be used up by another thread in between these two lines? What happens in that case? Might we need to have a lock that covers both or something to prevent that, or if not do we know how we'll deal with the issue when it happens.

2. Could we combine it into one function call to the donationRepo class? I think that might be clearer, and it can do the two things internally - maybe:

```suggestion
            $this->donationRepository->releaseAndReallocateMatchFunds($donation);
```

src/Domain/Campaign.php
```diff
@@ -30,6 +31,13 @@ class Campaign extends SalesforceReadProxy
      */
     protected Charity $charity;
 
+    /**
+     * @ORM\ManyToMany(targetEntity="CampaignFunding", mappedBy="campaigns")
+     * @psalm-suppress PossiblyUnusedProperty Used in Doctrine ORM mapping
+     * @psalm-suppress PropertyNotSetInConstructor Set by Doctrine on inverse side too
+     */
+    protected Collection $campaignFundings;
```
bdsl, line 39: Are you sure this property needs to be here? What goes wrong if it's deleted, isn't the one on CampaignFunding enough?

If it does need to be here then I think it's better to lose the second psalm annotation and set it to empty in the constructor, so if we do put code in to use it won't crash when we try to use it with an instance that didn't come from the EntityManger.

```
$this->campaignFundings = new ArrayCollection();
```

src/Domain/DonationRepository.php
```diff
@@ -345,6 +345,43 @@ public function findWithExpiredMatching(\DateTimeImmutable $now): array
...
     }
 
+    /**
+     * @return Donation[]   Donations which, when considered in isolation, could have some or all of their match
+     *                      funds swapped with higher priority matching (e.g. swapping out champion funds and
+     *                      swapping in pledges). The caller shouldn't assume that *all* donations may be fully
+     *                      swapped; typically we will choose to swap earlier-collected donations first, and it may
+     *                      be that priority funds are used up before we get to the end of the list.
+     */
+    public function findWithMatchingWhichCouldBeReplacedWithHigherPriorityAllocation(\DateTimeImmutable $cutoff): array
```
bdsl, line 355: Is this a clearer param name:

```suggestion
    public function findWithMatchingWhichCouldBeReplacedWithHigherPriorityAllocation(\DateTimeImmutable $createdSince): array
```

src/Domain/Campaign.php
NoelLH: @bdsl Pretty sure we need it. If I remove it the query gives

`[Semantical Error] line 0, col 171 near 'availableCf WITH': Error: Class MatchBot\Domain\Campaign has no association named campaignFundings`

When we have it it seems to work end to end – though it's hard for me to be 100% certain because my Docker-ised PHP CLI segfaults on most commands at the moment. I understand this is fixed in the PHP 8.2+ images so am trying to avoid spending time on it specifically.

Added constructor initaliser on 65c5f6d8962012b41f603ab0e057d716c5696a63

src/Application/Commands/RedistributeMatchFunds.php
NoelLH: I checked that commit too but I can't tell what we had a problem with before. I think our testing is good enough now that if things are green we can probably assume it's no longer an issue. Changed in 2a2b0988a8daa37b4564257c5f6acb14f7c0598e

src/Domain/DonationRepository.php
```diff
@@ -345,6 +345,43 @@ public function findWithExpiredMatching(\DateTimeImmutable $now): array
...
+                'availableCf',
+                'WITH',
+                'availableCf.amountAvailable > 0 AND availableCf.allocationOrder < donationCf.allocationOrder'
+            )
+            ->where('d.donationStatus IN (:collectedStatuses)')
+            ->andWhere('d.collectedAt > :checkAfter')
+            ->groupBy('d.id')
+            ->setParameter('collectedStatuses', DonationStatus::SUCCESS_STATUSES)
+            ->setParameter('checkAfter', $cutoff)
+        ;
```
bdsl, line 377: This is a relatively complex query. IMHO not worth unit testing but maybe worth integration testing.

Also I think the query builder code is harder to read than it would be if this was done as SQL or DQL code - maybe not worth changing now but IMHO the query builder is very rarely worthwhile, unless the structure of the query has to be dynamically generated. As Ocramius says at https://www.reddit.com/r/PHP/comments/a58yt9/comment/eblvfmd/ (via https://github.com/symfony/demo/issues/917 )

src/Domain/FundingWithdrawal.php
```diff
@@ -10,6 +10,8 @@
  * @ORM\Entity(repositoryClass="FundingWithdrawalRepository")
  * @ORM\HasLifecycleCallbacks
  * @ORM\Table
+ *
+ * @psalm-suppress PropertyNotSetInConstructor Not requiring all props on construct for now.
```
bdsl, line 14: yep, or maybe better add the new Psalm errors to the baseline? We don't require the props defined in old code to be on construct, but IMHO we should require any future props without default values to be set in the constructor.

tests/Application/Actions/Donations/CreateTest.php
```diff
@@ -231,7 +231,10 @@ public function testSuccessWithStripeAccountIDMissingInitiallyButFoundOnRefetch(
         $donation->setCharityFee('0.38'); // Calculator is tested elsewhere.
         $donation->getCampaign()->getCharity()->setStripeAccountId(null);
 
-        $fundingWithdrawalForMatch = new FundingWithdrawal();
+        $campaignFunding = new CampaignFunding();
+        $campaignFunding->setCurrencyCode('GBP');
+        $campaignFunding->setAmountAvailable('8.00');
```
bdsl, line 236: can this also be `self::someCampaignFunding()`?

src/Domain/Campaign.php
```diff
@@ -30,6 +32,12 @@ class Campaign extends SalesforceReadProxy
      */
     protected Charity $charity;
 
+    /**
+     * @ORM\ManyToMany(targetEntity="CampaignFunding", mappedBy="campaigns")
+     * @psalm-suppress PossiblyUnusedProperty Used in Doctrine ORM mapping
+     */
+    protected Collection $campaignFundings;
```
bdsl, line 39: ah I just remembered since PHP 8.1 we can use [new in initialisers](https://stitcher.io/blog/php-81-new-in-initializers). So we can leave the constructor as it was and do

```suggestion
    protected Collection $campaignFundings = new ArrayCollection();
```

bdsl: Oops no that isn't allowed on a property ignore me.

src/Domain/DonationRepository.php
```diff
@@ -600,9 +640,8 @@ private function safelyAllocateFunds(Donation $donation, array $fundings, string
             $amountLeftToMatch = bcsub($amountLeftToMatch, $amountAllocated, 2);
 
             if (bccomp($amountAllocated, '0.00', 2) === 1) {
-                $withdrawal = new FundingWithdrawal();
+                $withdrawal = new FundingWithdrawal($funding);
```
bdsl, line 643: :+1:

tests/Application/Actions/Donations/CreateTest.php
```diff
@@ -231,7 +231,10 @@ public function testSuccessWithStripeAccountIDMissingInitiallyButFoundOnRefetch(
...
+        $fundingWithdrawalForMatch = new FundingWithdrawal($campaignFunding);
         $fundingWithdrawalForMatch->setAmount('8.00'); // Partial match
         $fundingWithdrawalForMatch->setDonation($donation);
```
bdsl, line 239: Looks like this exact sequence of three lines is repeated 5 times. Can we extract to a function so we don't have to edit 5 copies next time?

```
$fundingWithdrawalForMatch = new FundingWithdrawal($campaignFunding);
$fundingWithdrawalForMatch->setAmount('8.00'); // Partial match
$fundingWithdrawalForMatch->setDonation($donation);
```

tests/Application/Commands/RedistibuteMatchFundsTest.php
```diff
@@ -0,0 +1,107 @@
...
+class RedistibuteMatchFundsTest extends TestCase
+{
+    public function testNoEligibleDonations(): void
+    {
+        $donationRepoProphecy = $this->prophesize(DonationRepository::class);
+        $donationRepoProphecy->findWithMatchingWhichCouldBeReplacedWithHigherPriorityAllocation(
+            Argument::type(\DateTimeImmutable::class)
+        )
+            ->willReturn([])
+            ->shouldBeCalledOnce();
```
bdsl, line 31: as mentioned on the SUT, it would be great if we can be more specific with the argument than just the type.

Also as this is a query not a command is there any need for `shouldBeCalledOnce()`?

tests/Application/Commands/RedistibuteMatchFundsTest.php
```diff
@@ -0,0 +1,107 @@
...
+                projectId: 'project-id',
+                psp: 'stripe',
+            ), $this->getMinimalCampaign()),
+        ]);
+        $donationRepoProphecy->releaseMatchFunds(Argument::type(Donation::class))
+            ->shouldBeCalledOnce();
+        $donationRepoProphecy->allocateMatchFunds(Argument::type(Donation::class))
+            ->shouldBeCalledOnce();
+        $donationRepoProphecy->push(Argument::type(Donation::class), false)
+            ->shouldBeCalledOnce();
```
bdsl, line 65: I think we might as well be specific about which donation should be operated on here and in similar tests instead of just relying on the type - Psalm checks types for us, here it'd be good to test specific values. I haven't run this code but I think this should work:

Also gets a bit more concise:

```suggestion

$donation =  Donation::fromApiModel(new DonationCreate(
                currencyCode: 'GBP',
                donationAmount: '10',
                projectId: 'project-id',
                psp: 'stripe',
            );
            $this->getMinimalCampaign())$donationRepoProphecy->findWithMatchingWhichCouldBeReplacedWithHigherPriorityAllocation(Argument::type(\DateTimeImmutable::class))->willReturn([
            $donation,
        ]);
        $donationRepoProphecy->releaseMatchFunds($donation)->shouldBeCalledOnce();
        $donationRepoProphecy->allocateMatchFunds($donation)->shouldBeCalledOnce();
        $donationRepoProphecy->push($donation, false)->shouldBeCalledOnce();
```

tests/Domain/DonationRepositoryMatchFundsAllocationTest.php
```diff
@@ -132,7 +131,7 @@ public function testItAllocates1From1For1(): void
 
 
     /**
-     * @return list<array{0:string, 1: string, 2: string, 3: string, 4: string, 5: string}>
+     * @psalm-return list<array{0:string, 1: string, 2: numeric-string, 3: string, 4: string, 5: string}>
```
bdsl, line 134: No objection, but out of curiosity why change from @return to @psalm-return? Would you prefer if I use the vendor prefix in future?

bdsl (commented):
Review done, added comments inline. Otherwise all looks good.

bdsl:
Also there are still quite a few mutants escaped in the business logic code of the new command - probably worth adding or improving tests to catch them: https://output.circle-artifacts.com/output/job/7cb90b10-429e-4be2-8afa-1e9f3907475d/artifacts/0/reports/infection.html#mutant/Application/Commands/RedistributeMatchFunds.php

src/Domain/FundingWithdrawal.php
```diff
@@ -57,15 +64,7 @@ public function getAmount(): string
...
     }
 
-    /**
-     * @param CampaignFunding $campaignFunding
-     */
-    public function setCampaignFunding(CampaignFunding $campaignFunding): void
-    {
-        $this->campaignFunding = $campaignFunding;
-    }
-
```
bdsl, line 67: :fireworks:

src/Domain/FundingWithdrawal.php
```diff
@@ -29,9 +31,14 @@ class FundingWithdrawal extends Model
 
     /**
      * @ORM\ManyToOne(targetEntity="CampaignFunding", fetch="EAGER")
-     * @var CampaignFunding|null
+     * @var CampaignFunding
      */
-    protected ?CampaignFunding $campaignFunding = null;
+    protected CampaignFunding $campaignFunding;
```
bdsl, line 36:
```suggestion
    private readonly CampaignFunding $campaignFunding;
```

bdsl:
I'm getting an error running this:

```
barney:~/projects/matchbot$ docker-compose exec app ./matchbot matchbot:redistribute-match-funds
matchbot:redistribute-match-funds starting!
[2023-12-16T18:42:36.247153+00:00] matchbot.DEBUG: Successfully acquired the "{resource}" lock. {"resource":{"Symfony\\Component\\Lock\\Key":"matchbot:redistribute-match-funds"}} {"uid":"40b49af","memory_peak_usage":"8 MB"}
[2023-12-16T18:42:36.256991+00:00] matchbot.DEBUG: Expiration defined for "{resource}" lock for "{ttl}" seconds. {"resource":{"Symfony\\Component\\Lock\\Key":"matchbot:redistribute-match-funds"},"ttl":1800.0} {"uid":"40b49af","memory_peak_usage":"8 MB"}

In QueryException.php line 43:
                                                                                                                                               
  [Semantical Error] line 0, col 171 near 'availableCf WITH': Error: Class MatchBot\Domain\Campaign has no association named campaignFundings  
                                                                                                                                               

In QueryException.php line 21:
                                                                                                                                                                                                                               
  SELECT d FROM MatchBot\Domain\Donation d INNER JOIN d.fundingWithdrawals fw INNER JOIN fw.campaignFunding donationCf INNER JOIN d.campaign c INNER JOIN c.campaignFundings availableCf WITH availableCf.amountAvailable > 0  
   AND availableCf.allocationOrder < donationCf.allocationOrder WHERE d.donationStatus IN (:collectedStatuses) AND d.collectedAt > :checkAfter GROUP BY d.id                                                                   
                                                                                                                                                                                                                               

matchbot:redistribute-match-funds

[2023-12-16T18:42:36.283164+00:00] matchbot.DEBUG: Successfully released the "{resource}" lock. {"resource":{"Symfony\\Component\\Lock\\Key":"matchbot:redistribute-match-funds"}} {"uid":"40b49af","memory_peak_usage":"8 MB"}
barney:~/projects/matchbot$ 
```

src/Domain/DonationRepository.php
bdsl: If you want to convert to DQL it should be easy using `echo $qb->getDQL();` and then pasting it in. I tried it and converts to the code shown below (I just added line breaks) but also gives the error as shown at https://github.com/thebiggive/matchbot/pull/737#issuecomment-1858903937 , both before and after conversion. Also not sure what the last line means.


```DQL
            SELECT d FROM MatchBot\Domain\Donation d 
            INNER JOIN d.fundingWithdrawals fw 
            INNER JOIN fw.campaignFunding donationCf 
            INNER JOIN d.campaign c 
            INNER JOIN c.campaignFundings availableCf WITH availableCf.amountAvailable > 0 AND availableCf.allocationOrder < donationCf.allocationOrder
            WHERE d.donationStatus IN (:collectedStatuses) AND
            d.collectedAt > :checkAfter 
            GROUP BY d.idChecked
            0 donations and redistributed matching for 0
```

NoelLH:
> I'm getting an error running this:
> 
> ```
> barney:~/projects/matchbot$ docker-compose exec app ./matchbot matchbot:redistribute-match-funds
> matchbot:redistribute-match-funds starting!
> [2023-12-16T18:42:36.247153+00:00] matchbot.DEBUG: Successfully acquired the "{resource}" lock. {"resource":{"Symfony\\Component\\Lock\\Key":"matchbot:redistribute-match-funds"}} {"uid":"40b49af","memory_peak_usage":"8 MB"}
> [2023-12-16T18:42:36.256991+00:00] matchbot.DEBUG: Expiration defined for "{resource}" lock for "{ttl}" seconds. {"resource":{"Symfony\\Component\\Lock\\Key":"matchbot:redistribute-match-funds"},"ttl":1800.0} {"uid":"40b49af","memory_peak_usage":"8 MB"}
> 
> In QueryException.php line 43:
>                                                                                                                                                
>   [Semantical Error] line 0, col 171 near 'availableCf WITH': Error: Class MatchBot\Domain\Campaign has no association named campaignFundings  
>                                                                                                                                                
> 
> In QueryException.php line 21:
>                                                                                                                                                                                                                                
>   SELECT d FROM MatchBot\Domain\Donation d INNER JOIN d.fundingWithdrawals fw INNER JOIN fw.campaignFunding donationCf INNER JOIN d.campaign c INNER JOIN c.campaignFundings availableCf WITH availableCf.amountAvailable > 0  
>    AND availableCf.allocationOrder < donationCf.allocationOrder WHERE d.donationStatus IN (:collectedStatuses) AND d.collectedAt > :checkAfter GROUP BY d.id                                                                   
>                                                                                                                                                                                                                                
> 
> matchbot:redistribute-match-funds
> 
> [2023-12-16T18:42:36.283164+00:00] matchbot.DEBUG: Successfully released the "{resource}" lock. {"resource":{"Symfony\\Component\\Lock\\Key":"matchbot:redistribute-match-funds"}} {"uid":"40b49af","memory_peak_usage":"8 MB"}
> barney:~/projects/matchbot$ 
> ```

@bdsl This sounds like https://github.com/thebiggive/matchbot/pull/737#discussion_r1428193324

Did you have the new, inversed property on your model? Do you maybe need to clear your Doctrine metadata cache to get the result of that? This has worked for me in "real" dev mode running the command, though frustratingly the segfaults have meant I still can't run integration tests locally seemingly because they are looking at the old metadata.

tests/Domain/DonationRepositoryMatchFundsAllocationTest.php
NoelLH: @bdsl I get an IDE or plugin notice about unknown types when I use `numeric-string` without a prefix, so this seemed like the simplest way to avoid adding dev noise of that sort. I don't have a strong preference for doing it everywhere.

src/Application/Commands/RedistributeMatchFunds.php
NoelLH: I agree longer term but had very limited time left before we said we'd patch CC data. I have made its setter throw on values < 0 for now instead.

src/Application/Commands/RedistributeMatchFunds.php
NoelLH: 1. I think unfortunately yes. We have nothing that locks allocation across funds in this way and I think it could be very disruptive, so there is a trade-off here. I've now added an error log to tell us if it happens. As already planned we will run this at quiet times – we might also want to talk again in the team about the risk vs. benefit before deciding to apply to it to still-open campaigns, since the only advantage is data charities see (which we could always caveat or fudge another way).
2. I don't think this gains us any lock safety, so my preference is to keep the separate-ness and slight risk clear for now.

bdsl:
> Did you have the new, inversed property on your model? Do you maybe need to clear your Doctrine metadata cache to get the result of that?

Hmm possibly.  Not sure exactly what cache is the culprit. My local files are matching the head of this PR according to git, and I've run `docker-compose exec app composer run doctrine:cache:clear` but I still get an error if I run `docker-compose exec app composer run integration-test`, unlike circle CI which shows a failure not an error. 

Will try and see what's going wrong if I have time:

```
~/projects/matchbot$ docker compose exec app composer run doctrine:cache:clear && docker compose exec app composer run integration-test
> doctrine orm:clear-cache:metadata
[Warning] The use of this script is discouraged. See https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/tools.html#doctrine-console for instructions on bootstrapping the console runner.



 // Clearing all Metadata cache entries                                                                                 

                                                                                                                        
 [OK] Successfully deleted cache entries.                                                                               
                                                                                                                        

> doctrine orm:clear-cache:query
[Warning] The use of this script is discouraged. See https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/tools.html#doctrine-console for instructions on bootstrapping the console runner.



 // Clearing all Query cache entries                                                                                    

                                                                                                                        
 [OK] Successfully deleted cache entries.                                                                               
                                                                                                                        

> doctrine orm:clear-cache:result
[Warning] The use of this script is discouraged. See https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/tools.html#doctrine-console for instructions on bootstrapping the console runner.



 // Clearing all Result cache entries                                                                                   

                                                                                                                        
 [OK] Successfully deleted cache entries.                                                                               
                                                                                                                        

> XDEBUG_MODE=coverage phpunit --config=phpunit-integration.xml --order-by=random
PHPUnit 9.6.13 by Sebastian Bergmann and contributors.

Random Seed:   1702896461

..........E.....                                                  16 / 16 (100%)

Time: 00:01.542, Memory: 30.00 MB

There was 1 error:

1) MatchBot\IntegrationTests\RedistributeMatchingCommandTest::testCommandRedistributesMatchingToUsePledges
Doctrine\ORM\Query\QueryException: [Semantical Error] line 0, col 171 near 'availableCf WITH': Error: Class MatchBot\Domain\Campaign has no association named campaignFundings

```

bdsl:
> Will try and see what's going wrong if I have time:

OK I took a bit of a sledgehammer to it, rebuilding the docker container with deleting the var directory and rebuilding the doctrine proxies means now I can run the integration tests and I get the same result as CircleCI does.

tests/Domain/DonationRepositoryMatchFundsAllocationTest.php
bdsl: @NoelLH Fair enough. That might be fixed by installing this plugin - it's bundled with PHPStorm, not sure if it comes bundled with IntelliJ https://plugins.jetbrains.com/plugin/15183-psalm-support

src/Application/Commands/RedistributeMatchFunds.php
bdsl: Adding the error log sounds good. If we start seeing the error in practice we can think about using a lock - I think it should be fairly straightforward to lock creation of new donations during the time this is running, probably using either https://dev.mysql.com/doc/refman/8.0/en/locking-functions.html#function_get-lock or https://symfony.com/doc/current/components/lock.html#available-stores .

src/Domain/Fund.php
```diff
@@ -18,6 +18,9 @@ abstract class Fund extends SalesforceReadProxy
 {
     use TimestampsTrait;
 
+    /** @var 'championFund'|'pledge'|'unknownFund' */
+    public const DISCRIMINATOR_VALUE = 'unknownFund';
```
bdsl, line 22: :+1:

NoelLH:
> > Will try and see what's going wrong if I have time:
> 
> OK I took a bit of a sledgehammer to it, rebuilding the docker container with deleting the var directory and rebuilding the doctrine proxies means now I can run the integration tests and I get the same result as CircleCI does.

I've now included in the Composer script the steps I had to take, in order, to do this reliably without wiping out the whole container. It might be overkill in some cases – we could perhaps split it into 2 versions – but I think it's better to make it reliable by default even if it's slightly noisier for devs.

tests/Application/Commands/RedistibuteMatchFundsTest.php
```diff
@@ -0,0 +1,222 @@
...
+class RedistibuteMatchFundsTest extends TestCase
+{
+    public function testNoEligibleDonations(): void
+    {
+        $now = new \DateTimeImmutable('now');
+
+        $donationRepoProphecy = $this->prophesize(DonationRepository::class);
+        $donationRepoProphecy->findWithMatchingWhichCouldBeReplacedWithHigherPriorityAllocation(
+            $now,
+            $now->sub(new \DateInterval('P6W')),
```
bdsl, line 34: IMHO one of the main reasons to DI the time is to allow us to be specific about what scenario we're testing with regards to dates & times, and avoid repeating code from the SUT in the test. Maybe something like this: https://github.com/thebiggive/matchbot/compare/MAT-342-redistribute-matching...MAT-342-redistribute-matching-use-specific-datetime-in-test?expand=1 ?

I don't know if it's possible to do the same with the integration test - it might be, or that might require wider refactoring.

NoelLH: Thanks, I've merged that branch into this.

Integration tests just use `now` already so I think that short of injecting a clock they're probably as good as they can be. I think there is also an advantage in using the 'real' clock in that case so I'm not sure I would significantly change what is happening even if a clock were being injected.

PR #741, opened 2023-12-22
BG2-2520 – update minimum PHP to 8.2

NoelLH:
Run the real app with that version, and start testing 8.3 again

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
     parameters:
       php_version:
         type: string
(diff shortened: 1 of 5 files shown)


NoelLH:
This can't be merged while we depend upon https://github.com/Lctrs/psalm-psr-container-plugin and it doesn't allow PHP 8.3. I'll make + merge a PR to only update the live version to 8.2 so we can try it on Staging.

NoelLH:
Superseded

PR #748, opened 2024-01-02
BG2-2537 – set email address blank for one donation as requested

NoelLH:


dorota-joanna (approved):
LGTM (the uuid and id are correct)

PR #752, opened 2024-01-03
BG2-2520: Replace Lctrs/psalm-psr-container-plugin with stub file

bdsl:
Lctrs/psalm-psr-container-plugin does not officially support PHP 8.3 yet (see https://github.com/Lctrs/psalm-psr-container-plugin/pull/963 ) and may be the only thing blocking us from using 8.3 imediatly in Matchbot.

This replaces the plugin with a stub file to add Psalm annotations to the ContainerInterface and DI\Container class.

Also moved the Psalm baseline file into a subdirectory to live with the new stub file


dorota-joanna (approved):
LGTM

PR #753, opened 2024-01-05
MAT-342 – consider donations up to 8 weeks back for match funds redistribution

NoelLH:


src/Application/Commands/RedistributeMatchFunds.php
```diff
@@ -38,7 +38,7 @@ protected function doExecute(InputInterface $input, OutputInterface $output): in
             // Since very long campaigns usually only have one funding type, it's currently unlikely
             // that the combination of minimum & maximum dates will stop funds being redistributed when
             // we'd like them to.
-            donationsCollectedAfter: $this->now->sub(new \DateInterval('P6W')),
+            donationsCollectedAfter: $this->now->sub(new \DateInterval('P8W')),
```
bdsl, line 41: test needs updating also @NoelLH

NoelLH: Yep, done now

PR #754, opened 2024-01-05
BG2-2520: Replace PHP 8.2 with 8.3

bdsl:
Since this is deployed as a container with a PHP runtime included it doesn't make sense to try to support multiple PHP versions at the same time as I was doing before the holiday.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
     parameters:
       php_version:
         type: string
-        default: dev-8.2
+        default: dev-8.3
     docker:
       - image: thebiggive/php:<< parameters.php_version >>
         auth:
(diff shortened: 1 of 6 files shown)


NoelLH (approved):
I think we just need to change the branch rules back to make GitHub happy. Content of the change looks good to me (although I think keeping matrices in for some apps is fine too, for quick or early testing against new PHPs).

bdsl:
Thanks @NoelLH . I think the branch rules are not version controlled so I can merge this first and then change the rules. But one thing I'm confused by is why circle-ci is call the thing `test-1` instead of just `test`. I couldn't find anything about a number for it in the config. `mutation_converage` doesn't have a number.

NoelLH:
> Thanks @NoelLH . I think the branch rules are not version controlled so I can merge this first and then change the rules. But one thing I'm confused by is why circle-ci is call the thing `test-1` instead of just `test`. I couldn't find anything about a number for it in the config. `mutation_converage` doesn't have a number.

I think it probably numbers them when more than one job shares a name globally in the config, even if in different workflows

PR #762, opened 2024-01-22
DON-954: Fix processing of web hoooks for non-card payments

bdsl:
This may have been broken recently, but on my local I was getting an exception for non-card (i.e. donation funds) payments:

[2024-01-22T15:24:59.419989+00:00] matchbot.ERROR: Exception: Stripe Notice: Undefined property of Stripe\StripeObject instance: card (etc)

Tested the new version and it works both with cards and non cards - we still capture the full details of the card.


dorota-joanna (approved):
LGTM

PR #764, opened 2024-01-23
BG2-2548: Fix stripe update script

bdsl:
The previous version ran in prod but didn't actually do anything useful I realise now the `updateAttributes` function was just for updating the attributes in the object in memory on the PHP side - it doesn't actually send anything to Stripe. `updatePerson` is what is recommended at
https://stripe.com/docs/connect/update-verified-information?locale=en-GB#change-the-account-representative


dorota-joanna (approved):
LGTM

PR #769, opened 2024-02-05
MAT-337: Refactor - simplify Matching Adapter, replacing "transaction" concept with explicit internal calls to save to DB.

bdsl:
Please see list of commit messages.


dorota-joanna (approved):
LGTM

PR #770, opened 2024-02-06
BG2-2550: Resumbit gift aid for one campaign to HMRC

bdsl:


src/Migrations/Version20240206145629.php
```diff
@@ -0,0 +1,43 @@
...
+
+        $this->addSql(<<<SQL
+            UPDATE Donation
+            SET
+                tbgGiftAidRequestQueuedAt = NULL,
+                tbgGiftAidRequestFailedAt = NULL,
+                tbgGiftAidRequestConfirmedCompleteAt = NULL,
+                tbgGiftAidRequestCorrelationId = NULL,
+                tbgGiftAidResponseDetail = NULL,
+                salesforcePushStatus = 'pending-update'
```
bdsl, line 27: We haven't updated salesforcePushStatus in similar tasks before, but I think we should now that we're going to put `tbgGiftAidRequestConfirmedCompleteAt` in Salesforce.

dorota-joanna (approved):
LGTM

PR #773, opened 2024-02-19
BG2-2554: Disallow attempting to claim gift aid without providing a home address

bdsl:
With this change it becomes impossible to complete a donation with gift aid ticked and the home address given as only whitespace. The user sees the following. I'll make a second PR in frontend to make the enforcement of this friendlier and happen at the point of progressing through the GA step instead of at the end of the form.

As part of this I have refactored to pass several update fields to the donation model at once, instead of passing them one at a time, so that the model can enforce the invariant. It should make it easy to enforce any similar invariants we may want in future.

image

### DIFF
--- a/tests/Application/Actions/Donations/GetTest.php
+++ b/tests/Application/Actions/Donations/GetTest.php
@@ -153,7 +153,7 @@ class GetTest extends TestCase
         $this->assertNotEmpty($payloadArray['createdTime']);
         $this->assertEquals('1 Main St, London N1 1AA', $payloadArray['billingPostalAddress']);
         $this->assertTrue($payloadArray['giftAid']);
-        $this->assertTrue($payloadArray['optInCharityEmail']); // optincharityemail should be true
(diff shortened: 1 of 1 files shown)


tests/Application/Actions/Donations/GetTest.php
```diff
@@ -152,7 +153,7 @@ public function testSuccess(): void
         $this->assertNotEmpty($payloadArray['createdTime']);
         $this->assertEquals('1 Main St, London N1 1AA', $payloadArray['billingPostalAddress']);
         $this->assertTrue($payloadArray['giftAid']);
-        $this->assertTrue($payloadArray['optInCharityEmail']);
+        $this->assertTrue($payloadArray['optInCharityEmail']); // optincharityemail should be true
```
bdsl, line 156:
```suggestion
        $this->assertTrue($payloadArray['optInCharityEmail']);
```

dorota-joanna (approved):
LGTM

PR #774, opened 2024-02-20
BG2-2545 (tangentially related): Disallow updates to any non-pending …

bdsl:
There's no reason for the browser to send an update for a donation if it's already not pending.

Currently if they tried it would be blocked by an exception generated by the Stripe library, but I don't like the fact that we're relying on stripe to block it.


dorota-joanna (commented):
LGTM

PR #784, opened 2024-03-05
CLA-37 – reconcile donations from retried payouts, setting them 'Paid'

NoelLH:
This should ensure that for all future payouts, donations' status is updated after an earlier payout they were in failed but the latest payout's gone through.


bdsl (approved):
LGTM - @NoelLH will you just check the commits I pushed in before merging?

bdsl (approved):

PR #788, opened 2024-03-12
CLA-37 – add payout status check & temporary command for processing old edge case payouts

NoelLH:
* [check the Stripe API agrees payouts are "paid" before we update their donations' statuses](https://github.com/thebiggive/matchbot/pull/788/commits/330c0bc0fae395e0f4ce0f06b5166a285abcb98d)
[330c0bc](https://github.com/thebiggive/matchbot/pull/788/commits/330c0bc0fae395e0f4ce0f06b5166a285abcb98d)
* [add a temporary on-demand command `matchbot:reconcile-payouts` for historic payout edge cases](https://github.com/thebiggive/matchbot/pull/788/commits/606e6e1c9a82fbd444ebdbce7099551e327e2731)
[606e6e1](https://github.com/thebiggive/matchbot/pull/788/commits/606e6e1c9a82fbd444ebdbce7099551e327e2731)

### DIFF
--- a/tests/Application/Commands/ReconcilePastPayoutsTest.php
+++ b/tests/Application/Commands/ReconcilePastPayoutsTest.php
@@ -1,0 +1,75 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MatchBot\Tests\Application\Commands;
+
+use MatchBot\Application\Commands\ReconcilePastPayouts;
+use MatchBot\Application\Messenger\StripePayout;
+use MatchBot\Tests\Application\DonationTestDataTrait;
+use MatchBot\Tests\TestCase;
+use Prophecy\Prophecy\ObjectProphecy;
+use Psr\Log\NullLogger;
+use Stripe\Event;
+use Symfony\Component\Console\Tester\CommandTester;
+use Symfony\Component\Lock\LockFactory;
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\RoutableMessageBus;
+use Symfony\Component\Messenger\Stamp\BusNameStamp;
+use Symfony\Component\Messenger\Stamp\TransportMessageIdStamp;
+
+class ReconcilePastPayoutsTest extends TestCase
+{
+    use DonationTestDataTrait;
+
+    public function testRun(): void
+    {
+        // Just mock out without checks for this temporary command. The output lines check is enough to be fairly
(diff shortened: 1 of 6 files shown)


tests/Application/Messenger/Handler/StripePayoutHandlerTest.php
```diff
@@ -278,18 +278,75 @@ public function testSuccessfulUpdateForRetriedPayout(): void
...
+
+        $stripeBalanceTransactionProphecy = $this->prophesize(BalanceTransactionService::class);
+        $stripeBalanceTransactionProphecy->all(
+            [
+                'limit' => 100,
+                'payout' => self::DEFAULT_PAYOUT_ID,
+            ],
+            ['stripe_account' => self::CONNECTED_ACCOUNT_ID],
+        )
+            ->shouldNotBeCalled();
```
bdsl, line 303: Can say we don't want it called with any arguments, simpler than just saying don't call it with these specific ones

```suggestion
        $stripeBalanceTransactionProphecy->all(Argument::cetera())
            ->shouldNotBeCalled();
```

bdsl (approved):
I think testNoOpWhenPayoutFailed could be made a bit simpler to read by making some things less specific and also maybe adding arrange/act/assert or given/when/then comments. 

Also for testing logging something like https://github.com/MetaSyntactical/inmemory-logger/blob/2.2.x/src/InMemoryLogger.php (which is pretty trivial and we could easily implement our own version or copy/paste it instead of depending on) might make tests simpler instead of using prophecy for logging assertions.

tests/Application/Commands/ReconcilePastPayoutsTest.php
```diff
@@ -0,0 +1,75 @@
...
+use Symfony\Component\Messenger\RoutableMessageBus;
+use Symfony\Component\Messenger\Stamp\BusNameStamp;
+use Symfony\Component\Messenger\Stamp\TransportMessageIdStamp;
+
+class ReconcilePastPayoutsTest extends TestCase
+{
+    use DonationTestDataTrait;
+
+    public function testRun(): void
+    {
```
bdsl, line 26: this test is fine and we might as well keep it for now and delete with ReconcilePastPayouts when that's been run. But really ReconcilePastPayouts seems simple enough to not need testing, especially if we did a search and replace in the source code to change the CSV string into a hard-coded PHP array.

PR #790, opened 2024-03-12
MAT-279: Don't push very new donations to salesforce via PushDonation…

bdsl:
…s command


src/Application/Commands/PushDonations.php
```diff
@@ -8,11 +8,14 @@
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
+use function Symfony\Component\Clock\now;
+
```
NoelLH, line 12:
```suggestion
```

NoelLH (approved):

PR #796, opened 2024-03-13
CLA-37 – handle multiply-retried payouts when they succeed

NoelLH:


src/Application/Messenger/Handler/StripePayoutHandler.php
```diff
@@ -22,6 +22,10 @@
  */
 class StripePayoutHandler implements MessageHandlerInterface
 {
+    private const MAX_RETRY_DEPTH = 10;
+    /** @var string[] */
+    private static array $processedPayoutIds = [];
```
bdsl, line 27: Why is this static? Isn't there only ever going to be up to one instance of this class per process anyway (outside tests), in which case an instance property would do the same job. We never instantiate StripePayoutHandler ourselves, we let the container do it which afaik will not make more than one.

NoelLH: True, no good reason. Changed.

bdsl (dismissed):
LGTM although the complexity of this handler is getting higher, at some point it might be nice to try a non-recursive solution, either with an explicit loop or by placing work back on the message queue.

PR #799, opened 2024-03-14
MAT-356: Require fee cover amount to be zero for all donations

bdsl:


bdsl:
Before and after if the frontend sends a short project ID 08a876c9d301816ad18780444ff5e24f661ffb81 :

image

image

PR #806, opened 2024-03-19
main <- develop

bdsl:


NoelLH (commented):
This all looks good, but as discussed we should probably either revert the Doctrine property name change temporarily or make sure to release when traffic's very low

bdsl:
Thanks @NoelLH . I've reverted that property name change temporarily - thinking about it I don't like the idea of deploy when traffic is low if that also means out of office hours and therefore when there's few people around to deal with any possible issues.

PR #809, opened 2024-03-19
CLA-37 – tuning and log additions for payout reconciliation

NoelLH:


bdsl (approved):
Looks good - not quiet sure if use of the autoPagingIterator just makes our code simpler (which it does seem to do nicely) or is expected to change the behaviour. 

While we're here would it be worth building clickable links to the accounts and payouts in the stripe dashboard and including them in any logged error messages? Would avoid confusions like https://thebiggive.slack.com/archives/CDAF6QBFX/p1710758475530349?thread_ts=1710751599.430149&cid=CDAF6QBFX . I could make a PR to this PR or something to do that.

NoelLH:
> Looks good - not quiet sure if use of the autoPagingIterator just makes our code simpler (which it does seem to do nicely) or is expected to change the behaviour.

I couldn't identify a specific issue this would fix, but it removes one possible issue while simplifying.

> While we're here would it be worth building clickable links to the accounts and payouts in the stripe dashboard and including them in any logged error messages? Would avoid confusions like https://thebiggive.slack.com/archives/CDAF6QBFX/p1710758475530349?thread_ts=1710751599.430149&cid=CDAF6QBFX . I could make a PR to this PR or something to do that.

I think these are fast to put together once you've clicked around Stripe a bit. It could be good to add on a future PR but I wouldn't want to hold up the main fix.

PR #814, opened 2024-03-19
CLA-37 – patch 3 donations for one out of sync 2021 campaign

NoelLH:
..which doesn't have the transfer IDs to automate it


bdsl (approved):
LGTM

PR #824, opened 2024-04-04
Don't generate alarm if client attempts to confirm donation without p…

bdsl:
…roviding CVC

We think this is a result of bot activity, not anything we need to be alarmed by.

### DIFF
--- a/src/Application/Actions/Donations/Confirm.php
+++ b/src/Application/Actions/Donations/Confirm.php
@@ -11,7 +11,6 @@ use MatchBot\Client\NotFoundException;
 use MatchBot\Client\Stripe;
 use MatchBot\Domain\Donation;
 use MatchBot\Domain\DonationRepository;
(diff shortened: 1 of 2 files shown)


src/Application/Actions/Donations/Confirm.php
```diff
@@ -115,16 +114,25 @@ protected function action(Request $request, Response $response, array $args): Re
...
             );
         } catch (ApiErrorException $exception) {
-            $this->logger->error(sprintf(
-                'Stripe %s on Confirm updatePaymentMethodBillingDetail for donation %s (%s): %s',
-                get_class($exception),
-                $donation->getUuid(),
-                $paymentIntentId,
-                $exception->getMessage(),
-            ));
-
```
bdsl, line 125: removing this manual error log - it's not needed, and causes double errors, since throw an exception on line 128

dorota-joanna (approved):
LGTM

PR #825, opened 2024-04-04
Simplify calculator

bdsl:


dorota-joanna (approved):
LGTM

bdsl:
Mutation score is only 46% but pretty much all the escaped mutants are from our non-encapsulated bcmath usage: https://output.circle-artifacts.com/output/job/51c39121-7a79-4de5-bfe2-e4ef676cdb75/artifacts/0/reports/mutation-testing-all-files/infection.html#mutant/Application/Fees/Calculator.php

NoelLH (commented):
@bdsl I'm not sure about the wisdom of removing existing logic for the VAT rate changing in an election year – though I guess we should explicitly test both old and new rate behaviour separately if keeping the logic.

Otherwise the refactors LGTM.

bdsl:
@NoelLH The old code was accessing the system clock via ambient context, so didn't invite testing the logic. If the next government announces a VAT rate change on a future date then I think that will be straightforward for us to implement, and having code still active about a change in 2021 doesn't make it any easier.

If that happens I expect I would want to make the datetime an argument to the `Calculator::calculate` function, and consider whether the time the script runs is actually the most relavent date or if it could be something else, e.g. the creation time of the Donation - I'm not 100% sure now and don't need to be.

bdsl:
@NoelLH would you want to block this with the removal of the 2021 VAT change logic?

NoelLH:
> @NoelLH would you want to block this with the removal of the 2021 VAT change logic?

@bdsl Nope, I think you're right given the previous implementation not being very testable. We can handle it better in new code if it happens.

NoelLH (approved):

PR #826, opened 2024-04-04
MAT-220: removing trailing numbers from email addresses

dorota-joanna:


bdsl (approved):

bdsl:
created pr as a pair with @dorota-joanna

PR #827, opened 2024-04-05
MAT-289 – remove a source of Stripe test mode error log noise introduced in Jan

NoelLH:
I think this can be reviewed without reopening the original issue / new QA as it only affects non-live log levels.


bdsl (approved):
LGTM

PR #830, opened 2024-04-09
BG2-2598 – push donations to Salesforce in bigger batches and a more logical order

NoelLH:


src/Domain/SalesforceWriteProxyRepository.php
```diff
@@ -107,7 +107,7 @@ public function push(SalesforceWriteProxy $proxy, bool $isNew): bool
      * @param int $limit    Maximum of each type of pending object to process
      * @return int  Number of objects pushed
      */
-    public function pushSalesforcePending(\DateTimeImmutable $now, int $limit = 200): int
+    public function pushSalesforcePending(\DateTimeImmutable $now, int $limit = 400): int
```
bdsl, line 110: Not a blocker, none of the callers to this actually pass $limit, so for simplicity it might be worth removing the param and making it a class constant, or even just an inline literal.

bdsl: e.g. something like this?

```php
     *
     * @return int  Number of objects pushed
     */
    public function pushSalesforcePending(\DateTimeImmutable $now): int
    {
        // We don't want to push donations that were created or modified in the last 5 minutes,
        // to avoid collisions with other pushes.
        $fiveMinutesAgo = $now->modify('-5 minutes');

        /** @var SalesforceWriteProxy[] $proxiesToCreate */
        $proxiesToCreate = $this->findBy(
            criteria: ['salesforcePushStatus' => SalesforceWriteProxy::PUSH_STATUS_PENDING_CREATE],
            orderBy: ['id' => 'ASC'],
            limit: 200,
        );
```

bdsl (approved):
LGTM

src/Domain/SalesforceWriteProxyRepository.php
NoelLH: @bdsl Makes sense. It's used twice so have gone with a class constant.

src/Domain/SalesforceWriteProxyRepository.php
```diff
@@ -14,6 +14,9 @@
  */
 abstract class SalesforceWriteProxyRepository extends SalesforceProxyRepository
 {
+    /** Maximum of each type of pending object to process */
+    private const int MAX_PER_BULK_PUSH = 400;
```
bdsl: phpcs doesn't know that consts can have types. So far I've just avoided putting types on consts. If they're private then the type is fully inferable anyway.

NoelLH: Thanks, reached the same conclusion

bdsl (approved):

PR #831, opened 2024-04-10
BG2-2595: Resubmit gift aid claims

bdsl:


src/Migrations/Version20240410102808.php
```diff
@@ -0,0 +1,45 @@
...
+                tbgGiftAidRequestCorrelationId = NULL,
+                tbgGiftAidResponseDetail = NULL,
+                salesforcePushStatus = 'pending-update'
+            WHERE
+                campaign_id = $campaignIdForBG2_2595 AND
+                giftAid = 1 AND
+                tbgShouldProcessGiftAid = 1 AND
+                tbgGiftAidRequestCorrelationId IS NOT NULL AND
+                donationStatus = 'Paid'
+            LIMIT 73;
```
bdsl, line 36: noticed the 73 here was copied from a previous migration, isn't actually right in this case.

bdsl: Running ```
SELECT count(*) from Donation WHERE campaign_id = 6178 AND giftAid = 1 AND tbgShouldProcessGiftAid = 1 AND tbgGiftAidRequestCorrelationId IS NOT NULL AND donationStatus = 'Paid'; 
```

in prod returns 57

Comment (bdsl) [inline src/Migrations/Version20240410102808.php:36]:
```suggestion
            LIMIT 57;
```

NoelLH (approved):

PR #836, opened 2024-04-19
Mat 332 real time storage understanding

bdsl:
See also spike commit accidentally pushed to develop (and then reverted):  https://github.com/thebiggive/matchbot/commit/af89e236e3e12e456186bb5d53d56da045a7ffc8

### DIFF
--- a/tests/Application/Matching/AdapterTest.php
+++ b/tests/Application/Matching/AdapterTest.php
@@ -46,6 +46,7 @@ class AdapterTest extends TestCase
     public function testItReturnsAmountAvailableFromAFundingNotInStorage(): void
     {
         $funding = new CampaignFunding();
+        $funding->setId(1);
         $funding->setAmountAvailable('12.53');
 
         $amountAvaialble = $this->sut->getAmountAvailable($funding);
@@ -56,6 +57,7 @@ class AdapterTest extends TestCase
     public function testItAddsAmountForFunding(): void
     {
         $funding = new CampaignFunding();
+        $funding->setId(1);
         $funding->setAmountAvailable('50');
         $this->sut->addAmount($funding, '12.53');
         $this->entityManagerProphecy->persist($funding)->shouldBeCalled();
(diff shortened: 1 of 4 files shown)


tests/Application/Matching/AdapterTest.php
```diff
@@ -111,39 +115,43 @@ public function testItBailsOutAndReleasesFundsIfRetryingDoesntWorkDueToConcurren
...
+                "Fund 53 balance sub-zero after 6 attempts. Releasing final -14000 'cents'",
+                $exception->getMessage()
+            );
+        }
+
+        // fund started with 50, we successfully removed 30, and the callback removed 40 six times. We cleared up after
+        // our unsuccessful attempts to withdraw more, not cleared up after the callback as would be the responsibility
+        // of the other process.
+        \assert(50 - 30 - 40 * 6 === -220);
+        $this->assertSame('-220.00', $this->sut->getAmountAvailable($funding));
```
bdsl, line 148: the most important thing is not the individual reservations and releases, but that it ads up to make the total correct at the end. Hence adding this assertion.

tests/Application/Matching/AdapterTest.php
```diff
@@ -111,39 +115,43 @@ public function testItBailsOutAndReleasesFundsIfRetryingDoesntWorkDueToConcurren
         // let's assume another thread is causing the funds to reduce by 30 pounds just
         // after each time we increase it by 30 pounds.
```
NoelLH, line 116:
```suggestion
        // let's assume another thread is causing the funds to reduce by 40 pounds just
        // after each time we increase it by 30 pounds – via `subtractAmountWithoutSavingToDB()`
        // recovery calling `incrBy()` with overspent amount.
```

I think half this comment is talking about the amended value and the second half is about the subtracted £30 getting put back. But I'm not I fully understood it – so also proposing extra detail to confirm whether I have.

bdsl: I think your understanding is right, this is a good change to the comment. It was 30 but I increased it to 40 to see how that effected the result.

NoelLH (dismissed):

PR #839, opened 2024-04-22
MAT-358: Fix Deploying change to Doctrine Entity field name generates…

bdsl:
… errors

We now store the commit ID into a PHP file, and use that as part of the cache key for the Doctrine metadata in Redis. Therefore after any deploy of a new version of Matchbot the metadata cache will be effectively empty.

Writing the commit ID into a PHP file rather than a plain text file so that its content can be cached by opcache

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -163,6 +163,12 @@
             "Composer\\Config::disableProcessTimeout",
             "php matchbot-cli.php messenger:consume -vv --time-limit=7080"
         ],
+        "save-build-commit-id": [
+            "echo '<?php \nreturn \"'`git rev-parse HEAD`'\";\n' > .build-commit-id.php"
+        ],
+        "post-install-cmd": [
+            "composer save-build-commit-id"
+        ],
         "start": "php -S localhost:8080 -t public",
         "test": "XDEBUG_MODE=coverage phpunit --order-by=random --log-junit reports/unit-test.xml",
         "integration-test": [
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -163,6 +163,12 @@
             "Composer\\Config::disableProcessTimeout",
             "php matchbot-cli.php messenger:consume -vv --time-limit=7080"
         ],
+        "save-build-commit-id": [
+            "echo '<?php \nreturn \"'`git rev-parse HEAD`'\";\n' > .build-commit-id.php"
+        ],
+        "post-install-cmd": [
+            "composer save-build-commit-id"
+        ],
```
bdsl, line 171: tested by sshin'g into circleCI's build container:

```
root@6b92a6681b73:/var/www/html# cat .build-commit-id.php   
<?php 
return "0531dd6fb7981a4fd92220e4955f4fe534f81def";

```

So a file like that should now be present in every docker image we build.

NoelLH (approved):
Nice, seems like a good approach

Not sure if we might want to split it up to allow releasing even before Thurs, or just assume we will delay the next Prod release?

NoelLH:
> Nice, seems like a good approach
> 
> Not sure if we might want to split it up to allow releasing even before Thurs, or just assume we will delay the next Prod release?

As discussed, the `@todo` dropped is mostly about the risk that this PR fixes, so my concern there didn't really make sense. Releasing this will mean a cache miss so if nothing urgent comes up we might wait until after GMF closes, but it's not crucial to wait.

PR #843, opened 2024-04-26
MAT-360: Log Donation ID and return 400 in case we get an attempt to …

bdsl:
…confirm a donation with no specified payment method ID

### DIFF
--- a/src/Application/Actions/Donations/Confirm.php
+++ b/src/Application/Actions/Donations/Confirm.php
@@ -93,10 +93,10 @@ class Confirm extends Action
         }
 
         if (trim($paymentMethodId) === '') {
-            $donationID = $donation->getId();
+            $donationUUID = $donation->getId();
             $this->logger->error(
                 <<<EOF
-Donation Confirmation attempted with blank payment method id "$paymentMethodId" for Donation #$donationID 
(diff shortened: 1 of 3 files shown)


src/Application/Actions/Donations/Confirm.php
```diff
@@ -92,6 +92,17 @@ protected function action(Request $request, Response $response, array $args): Re
             throw new NotFoundException();
         }
 
+        if (trim($paymentMethodId) === '') {
+            $donationID = $donation->getId();
+            $this->logger->error(
+                <<<EOF
+Donation Confirmation attempted with blank payment method id "$paymentMethodId" for Donation #$donationID
```
NoelLH, line 99: UUID is surfaced in more places so maybe more helpful to log instead / as well?

NoelLH (commented):
Small suggestion on which ID we log, but looks good

src/Application/Actions/Donations/Confirm.php
bdsl: Makes sense, I will change it.

bdsl: Done in 6900cdf

PR #844, opened 2024-04-29
Only generate alarm if push to SF is 5 mins ore more delayed

bdsl:


NoelLH (commented):
It feels a little risky to me to make initial push failures quieter in Production while still keeping quite a long timeout. I think there have been occasions where this has been our first indication of a not yet reported Salesforce performance degradation that went on to cause bigger problems.

bdsl:
> It feels a little risky to me to make initial push failures quieter in Production while still keeping quite a long timeout. I think there have been occasions where this has been our first indication of a not yet reported Salesforce performance degradation that went on to cause bigger problems.

OK - let's close for now.

PR #845, opened 2024-05-01
MAT-362: Increase lookback time for checking for charity updates from Salesforce

bdsl:


src/Domain/CampaignRepository.php
```diff
@@ -28,25 +28,27 @@ class CampaignRepository extends SalesforceReadProxyRepository
      */
     public function findRecentLiveAndPendingGiftAidApproval(): array
     {
-        $oneWeekAgo = (new DateTime('now'))->sub(new \DateInterval('P7D'));
-        $fourMonthsAgo = (new DateTime('now'))->sub(new \DateInterval('P4M'));
```
bdsl, line 32: we had the one week ago info repeated 5 times in this file in various forms. I took four of them out instead of updating all 5.

integrationTests/CampaignRepositoryTest.php
```diff
@@ -49,7 +49,7 @@ public function testItFindsATwoWeekOldCampaignForACharityAwaitingGiftAidApproval
         $this->assertSame('Charity Name', $firstCampaign->getCharity()->getName());
     }
 
-    public function testItFindsNoTwentyWeekOldCampaignEvenIfCharityAwaitingGiftAidApproval(): void
+    public function testItFinds10MonthOldCampaignEvenIfCharityAwaitingGiftAidApproval(): void
```
NoelLH, line 52:
```suggestion
    public function testItFindsNo10MonthOldCampaignEvenIfCharityAwaitingGiftAidApproval(): void
```

NoelLH (commented):
I think we've reversed the meaning of a test name without reversing its logic. Otherwise LGTM

bdsl:
Thanks @NoelLH . I applied your suggestion and merged.

PR #847, opened 2024-05-02
Mat 361 matched amounts adjustments

bdsl:


src/Migrations/Version20240502101845.php
```diff
@@ -0,0 +1,65 @@
...
+            UPDATE CampaignFunding SET amount = amount - 2500, amountAvailable = amountAvailable - 1890 WHERE id = 29101 LIMIT 1;
+
+            -- reduce amount from 2_500 to 1890;
+            UPDATE FundingWithdrawal set amount = amount - 610 where id = 717813;
+            
+            -- reduce amount from 20_000 to 10_000, reduce amountAvailable from 9952 to 0
+            UPDATE CampaignFunding SET amount = amount - 10000, amountAvailable = amountAvailable - 9952 WHERE id = 29116 LIMIT 1;
+
+            -- reduce amount from 10_000 to zero
+            UPDATE FundingWithdrawal set amount = amount - 9952 where id = 709467 LIMIT 1;
```
bdsl, line 34: hmm now I'm thinking I went wrong somewhere because the 9952 is repeated in the previous line but the 610 further up wasn't repeated.

bdsl: I think 41feab0 fixes it.

NoelLH (approved):
Principle looks good as discussed on our call earlier.

Just did some last ID & value checks in more detail:
* CampaignFunding IDs are plausible (linked to the right Fund) and amount reductions look correct
* Campaign_CampaignFunding associations to the CampaignFundings link them to the expected campaigns and charities, and each CampaignFunding has only one such link
* donation SF IDs to re-push are correct

PR #848, opened 2024-05-02
Mat 361 bg2 2621 remove fund amount

bdsl:
There were no calls to Fund::getAmount except in test classes, showing that Matchbot was not using this data.

### DIFF
--- a/tests/Domain/FundRepositoryTest.php
+++ b/tests/Domain/FundRepositoryTest.php
@@ -49,7 +49,6 @@ class FundRepositoryTest extends TestCase
         $repo->updateFromSf($fund, false); // Don't auto-save as non-DB-backed tests can't persist
 
         $this->assertEquals('API Fund Name', $fund->getName());
-        $this->assertEquals('123.45', $fund->getAmount());
     }
 
     public function testPullForCampaignAllNew(): void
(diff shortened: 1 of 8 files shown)


tests/Domain/FundRepositoryTest.php
```diff
@@ -133,13 +124,6 @@ public function testPullForCampaignExistingFundButNewToCampaign(): void
             ->persist(Argument::type(CampaignFunding::class))
             ->shouldBeCalledTimes(2);
 
-        // This is not mututally exclusive with the above call expectations. It's a quick way to double check
-        // that both persists are setting their respective object's amount to £500, even when the pre-existing
-        // Fund we simulate had a £400 total balance before.
-        $entityManagerProphecy
-            ->persist(Argument::which('getAmount', '500'))
```
NoelLH, line 140: Took me a moment, but looks good. We were using our now deleted getter here as a convenient helper in test code to run against the Fund argument that live code passes to `persist()`. Ceasing to set amount in live code means there is nothing to get in test code.

NoelLH (approved):

tests/Domain/FundRepositoryTest.php
bdsl: Yes. And deleting the getter means we can't ask prophecy to call the getter. Took me a while too, I wasn't familiar with Agument::which

PR #852, opened 2024-05-10
Replace codecov with html code coverage artifacts to be generated ins…

bdsl:
…ide CircleCI

### DIFF
--- a/phpunit-integration.xml
+++ b/phpunit-integration.xml
@@ -9,6 +9,7 @@
     </exclude>
     <report>
       <clover outputFile="coverage-integration.xml" />
+      <html outputDirectory="reports/integration-coverage"/>
     </report>
   </coverage>
   <testsuites>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,6 @@ orbs:
   aws-ecr: circleci/aws-ecr@9.0.2
   aws-ecs: circleci/aws-ecs@3.2.0
   aws-cli: circleci/aws-cli@4.1.3
-  codecov: codecov/codecov@3.2.4
   jira: circleci/jira@1.3.1
   slack: circleci/slack@4.12.1
 
@@ -72,10 +71,10 @@ jobs:
       - store_artifacts:
           path: reports
 
-      - codecov/upload:
-          file: 'coverage.xml'
-      - codecov/upload:
-          file: 'coverage-integration.xml'
+#      - codecov/upload:
+#          file: 'coverage.xml'
+#      - codecov/upload:
+#          file: 'coverage-integration.xml'
 
   mutation_coverage:
     resource_class: large
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -25,7 +25,8 @@
       <directory suffix=".php">./src/Migrations</directory>
     </exclude>
     <report>
-      <clover outputFile="coverage.xml" />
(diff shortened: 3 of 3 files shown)


phpunit-integration.xml
```diff
@@ -9,6 +9,7 @@
     </exclude>
     <report>
       <clover outputFile="coverage-integration.xml" />
+      <html outputDirectory="reports/integration-coverage"/>
```
NoelLH, line 12: Is there a way to get combined coverage figures for the two test types? It would be great to have numbers that are comparable to the joint summary Codecov gives.

bdsl: That's what I was tying to do with codecov, but I ran into problem and gave up. Probably we could have PHPUnit config that runs all the tests at once to get combined coverage though.

bdsl:
I've now got a test run working (most of the time) that does unit and integration tests combined - and so reporting both traditional coverage and also mutation coverage based on the combined test suite:

image

NoelLH:
This seems like a nice thing to have as an artefact.

I think the main thing I would miss is being able to see overall and granular coverage changes on demand for a PR, at a glance. Not as status checks but just comparatively somewhere in an app.

Could we consider doing this to remove checks instead of removing all Codecov? https://github.com/thebiggive/matchbot/pull/854 @bdsl 

Very happy with the addition of CI artefact reports staying too.

bdsl:
> This seems like a nice thing to have as an artefact.
> 
> I think the main thing I would miss is being able to see overall and granular coverage changes on demand for a PR, at a glance. Not as status checks but just comparatively somewhere in an app.
> 
> Could we consider doing this to remove checks instead of removing all Codecov? #854 @bdsl
> 
> Very happy with the addition of CI artefact reports staying too.

Yes I don't see why not - do you want to merge your PR first, then I can fix conflicts and merge this one to to keep the codecov in addition to the extra artifacts.

bdsl:
> > This seems like a nice thing to have as an artefact.
> > I think the main thing I would miss is being able to see overall and granular coverage changes on demand for a PR, at a glance. Not as status checks but just comparatively somewhere in an app.
> > Could we consider doing this to remove checks instead of removing all Codecov? #854 @bdsl
> > Very happy with the addition of CI artefact reports staying too.
> 
> Yes I don't see why not - do you want to merge your PR first, then I can fix conflicts and merge this one to to keep the codecov in addition to the extra artifacts.

Thanks @NoelLH . I pushed a change to this PR to restore uploading to codecov.

But it seems the second test run seems very flaky is shown below - specifically the Redis connection fails at least half the time. Since it's not actually used as any sort of deployment gate we could just put `|| true` at the end of every flaky command there but having coverage artifacts missing half the time is going to be probably worse than not having them at all.

The PR you merged is probably enough to do what we talked about in the retro, so I think unless you can see any cheap way to fix the flakiness here I'll just close this PR without merging. I should really have started with that, I think I hadn't actually noticed that we had a codecov.yml file.


https://app.circleci.com/pipelines/github/thebiggive/matchbot/2965/workflows/463a580a-36e9-4667-ab73-02645d39d5d5/jobs/8235
image

NoelLH:
Thanks @bdsl, can't immediately think of why that would be happening / how to improve it. Bit of a shame not to get the bonus artefacts but I agree we shouldn't keep it while it's unstable.

PR #853, opened 2024-05-14
MAT-365 – adjust messaging when a donation create attempt is for a Stripe account without capabilities

NoelLH:
* Surface detail directly in Slack #stripe for whole team
* Return a 409 to reduce alarm noise for devs

### DIFF
--- a/src/Application/Actions/Donations/Create.php
+++ b/src/Application/Actions/Donations/Create.php
@@ -18,27 +18,30 @@ use MatchBot\Application\Auth\PersonWithPasswordAuthMiddleware;
 use MatchBot\Application\HttpModels\DonationCreate;
 use MatchBot\Application\HttpModels\DonationCreatedResponse;
 use MatchBot\Application\Matching\Adapter;
+use MatchBot\Application\Notifier\StripeChatterInterface;
 use MatchBot\Application\Persistence\RetrySafeEntityManager;
 use MatchBot\Client\Stripe;
 use MatchBot\Domain\CampaignRepository;
 use MatchBot\Domain\Charity;
-use MatchBot\Domain\Donation;
 use MatchBot\Domain\DonationRepository;
-use Monolog\Logger;
+use Psr\Container\ContainerInterface;
 use Psr\Http\Message\ResponseInterface as Response;
 use Psr\Http\Message\ServerRequestInterface as Request;
 use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
 use Random\Randomizer;
 use Stripe\Exception\ApiErrorException;
 use Symfony\Component\Clock\ClockInterface;
+use Symfony\Component\Notifier\ChatterInterface;
+use Symfony\Component\Notifier\Message\ChatMessage;
 use Symfony\Component\Serializer\Exception\UnexpectedValueException;
 use Symfony\Component\Serializer\SerializerInterface;
 
 class Create extends Action
 {
     private const MAX_RETRY_COUNT = 4;
+    private ChatterInterface $chatter;
 
-    #[Pure]
     public function __construct(
         private DonationRepository $donationRepository,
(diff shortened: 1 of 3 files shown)


src/Application/Actions/Donations/Create.php
```diff
@@ -238,24 +250,51 @@ function () use ($donation) {
...
-                $error = new ActionError(ActionError::SERVER_ERROR, 'Could not make Stripe Payment Intent (B)');
-                return $this->respond($response, new ActionPayload(500, null, $error));
+                );
+
+                $level = $accountLacksCapabilities ? LogLevel::WARNING : LogLevel::ERROR;
+                $this->logger->log($level, $failureMessage);
+
+                if ($accountLacksCapabilities) {
+                    /** @var string $env */
+                    $env = getenv('APP_ENV');
```
bdsl, line 275: Assertions are generally preferable to docblocks

```
                    $env = getenv('APP_ENV');
                    \assert(is_string($env));
```

bdsl, line 275: Assertions are generally preferable to docblocks

```suggestion
                    $env = getenv('APP_ENV');
                    \assert(is_string($env));
```

tests/Application/Actions/Donations/CreateTest.php
```diff
@@ -988,6 +927,66 @@ public function testSuccessWithMinimalData(): void
         $this->assertEquals('123CampaignId12345', $payloadArray['donation']['projectId']);
     }
 
+    public function testCharityLackingCapabilities(): void
```
bdsl, line 930: From a quick look this test seems quite long and hard to follow - I feel like we should very rarely if ever want a test function that doesn't fit on the screen, and it might be overspecifying the test subject and making hard work for us in future if we want to change it. Happy to have a look at adjusting with you tomorrow if you like.

bdsl (commented):
Generally LGTM, added some comments inline.

tests/Application/Actions/Donations/CreateTest.php
```diff
@@ -1072,4 +1071,39 @@ private static function someCampaignFunding(): CampaignFunding
 
         return $campaignFunding;
     }
+
+    private static function somePaymentIntentArgs(): array
```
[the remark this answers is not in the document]
NoelLH: Went with a class var and set in `setup()` since const values and var defaults cannot be expressions

tests/Application/Actions/Donations/CreateTest.php
NoelLH: I'm not sure it is feasible for this test to fit on one screen without extracting things specific to it, which IMO would make it more time consuming to understand. But I've now extracted out a bunch of common DI boilerplate I found was the same in a majority of tests, making a lot of them shorter.

tests/Application/Actions/Donations/CreateTest.php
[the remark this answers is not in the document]
bdsl: Ah I hadn't spotted that it contained a call to `getenv('APP_ENV')`. Not sure about static class var since that makes it look like one test case may influence the next run to run, although it wouldn't in this case since getenv will return the same every time.

tests/Application/Actions/Donations/CreateTest.php
bdsl: Thanks - it does look a lot clearer now, I think the AAA comments help quite a lot as well as the extractions you've done. I still feel like it should be feasible to make it simpler, reducing the use of the automocker and maybe either calling the controller under test directly instead of going through the app, or extracting the logic to a different class and testing that instead of testing the controller.

bdsl:
Thanks @NoelLH - as mentioned inline I think this is a lot easier to read than it was yesterday, I'd suggest we merge it as-is and then I'll try to make a follow up PR to see if I can get the test even easier to work with.

PR #860, opened 2024-05-21
Remove GA for non-elgible donations

bdsl:
(see slack thread in #gift-aid from May 2nd)

### DIFF
--- a/src/Migrations/Version20240521122101.php
+++ b/src/Migrations/Version20240521122101.php
@@ -18,7 +18,7 @@ final class Version20240521122101 extends AbstractMigration
     {
         $this->addSql(<<<SQL
             UPDATE Donation set giftAid = 0, tipGiftAid = 0, salesforcePushStatus = 'pending-update'
-            WHERE Donation.salesforcePushStatus IN ('a0669000020MEiaAAG', 'a0669000020M9saAAC', 'a0669000020LtFrAAK')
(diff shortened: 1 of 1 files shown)


src/Migrations/Version20240521122101.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public function getDescription(): string
+    {
+        return 'Remove GA setting from inelligble donations';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql(<<<SQL
+            UPDATE Donation set giftAid = 0, salesforcePushStatus = 'pending-update'
```
NoelLH, line 20:
```suggestion
            UPDATE Donation set giftAid = 0, tipGiftAid = 0, salesforcePushStatus = 'pending-update'
```

src/Migrations/Version20240521122101.php
```diff
@@ -0,0 +1,31 @@
...
+            WHERE Donation.salesforcePushStatus IN ('a0669000020MEiaAAG', 'a0669000020M9saAAC', 'a0669000020LtFrAAK')
```
NoelLH, line 21:
```suggestion
            WHERE Donation.salesforceId IN ('a0669000020MEiaAAG', 'a0669000020M9saAAC', 'a0669000020LtFrAAK')
```

NoelLH (commented):
I think we need to ensure we do not claim Gift Aid on tips too, and there's a typo in the IDs clause. Otherwise makes sense, checked donations are the ones I expected.

PR #864, opened 2024-05-22
Refactor: Extract several private functions from \MatchBot\Domain\Don…

bdsl:
…ationService::createDonation

Not really ticket related, but I was idle and this does seem to make it easier to get an overview of what create donation does. All extractions done with phpstorm.

The only intended behaviour change is that if the donation is not a stripe donation we now throw a an error as we don't know how to process anything else, rather than doing nothing.


bdsl:
Closing as not directly related to any current ticket. May re-open if it seems useful for future work.

PR #866, opened 2024-05-30
MAT-363: Dispatch event to symfony messenger when donation status upd…

bdsl:
…ated

We will use this to push updates to salesforce async

### DIFF
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -364,8 +364,7 @@ return function (ContainerBuilder $containerBuilder) {
             $busContainer->set('claimbot.donation.result', $bus);
             $busContainer->set(\Stripe\Event::PAYOUT_PAID, $bus);
 
-            $routableMessageBus = new RoutableMessageBus($busContainer, $bus);
-            return $routableMessageBus;
+            return new RoutableMessageBus($busContainer, $bus);
         },
 
         SerializerInterface::class => static function (ContainerInterface $c): SerializerInterface {
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -360,6 +360,11 @@ return function (ContainerBuilder $containerBuilder) {
             $busContainer = new Container();
             $bus = $c->get(MessageBusInterface::class);
 
+            /**
+             * Every message defaults to our only bus, so we think these are technically redundant for
+             * now. This also means the list isn't exhaustive – we *know* we don't need
+             * {@see DonationStateUpdated} here. Removing `claimbot` items would require some more testing.
+             */
             $busContainer->set('claimbot.donation.claim', $bus);
             $busContainer->set('claimbot.donation.result', $bus);
             $busContainer->set(\Stripe\Event::PAYOUT_PAID, $bus);
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -13,6 +13,7 @@ use MatchBot\Application\Auth;
 use MatchBot\Application\Auth\IdentityToken;
 use MatchBot\Application\Matching;
 use MatchBot\Application\Messenger\DonationStateUpdated;
+use MatchBot\Application\Messenger\Handler\DonationStateUpdatedHandler;
 use MatchBot\Application\Messenger\Handler\GiftAidResultHandler;
 use MatchBot\Application\Messenger\Handler\StripePayoutHandler;
 use MatchBot\Application\Messenger\StripePayout;
@@ -249,7 +250,7 @@ return function (ContainerBuilder $containerBuilder) {
                 [
                     Messages\Donation::class => [$c->get(GiftAidResultHandler::class)],
                     StripePayout::class => [$c->get(StripePayoutHandler::class)],
-                //       DonationStateUpdated::class => [$c->get(DonationStateUpdatedHandler::class)],
+                    DonationStateUpdated::class => [$c->get(DonationStateUpdatedHandler::class)],
                 ],
             ));
             $handleMiddleware->setLogger($logger);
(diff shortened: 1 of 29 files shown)


app/dependencies.php
```diff
@@ -247,6 +249,7 @@ static function (ContainerInterface $c): Matching\Adapter {
                 [
                     Messages\Donation::class => [$c->get(GiftAidResultHandler::class)],
                     StripePayout::class => [$c->get(StripePayoutHandler::class)],
+                //       DonationStateUpdated::class => [$c->get(DonationStateUpdatedHandler::class)],
```
dorota-joanna, line 252:
```suggestion
                // DonationStateUpdated handler isn't written yet
                //       DonationStateUpdated::class => [$c->get(DonationStateUpdatedHandler::class)],
```

app/dependencies.php
```diff
@@ -384,6 +390,7 @@ static function (ContainerInterface $c): Matching\Adapter {
             $transportFactory = new TransportFactory([
                 new AmazonSqsTransportFactory(),
                 new RedisTransportFactory(),
+               // new Symfony\Component\Messenger\Transport\InMemory\InMemoryTransportFactory(), // only used
```
dorota-joanna, line 393:
```suggestion
```

matchbot-cli.php
```diff
@@ -4,6 +4,7 @@
 declare(strict_types=1);
 
 $psr11App = require __DIR__ . '/bootstrap.php';
+//\assert($psr11App instanceof DI\Container);
```
dorota-joanna, line 7:
```suggestion
```

src/Application/Commands/PushDonations.php
```diff
@@ -8,7 +8,7 @@
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
-class PushDonations extends LockingCommand
+class PushDonations extends LockingCommand // replace this perhaps
```
dorota-joanna, line 11:
```suggestion
class PushDonations extends LockingCommand
```

src/Application/Messenger/DonationStateUpdated.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace MatchBot\Application\Messenger;
+
+use MatchBot\Application\Assertion;
+use MatchBot\Domain\Donation;
+use Ramsey\Uuid\Uuid;
+
+readonly class DonationStateUpdated
+{
+    public function __construct(
```
bdsl, line 11:
```suggestion
    private function __construct(
```

src/Domain/SalesforceWriteProxyRepository.php
```diff
@@ -122,6 +122,13 @@ public function pushSalesforcePending(\DateTimeImmutable $now): int
             self::MAX_PER_BULK_PUSH,
         );
 
+        if ($proxiesToCreate !== []) {
+            $count = count($proxiesToCreate);
+            $this->logger->error(
+                "Found $count pending items to push to SF, suggests push via Symfony Messenger failed"
```
NoelLH, line 128: I think this might be a bit noisier than we want – at busy times, it might be fairly normal for the regularly scheduled push to have some messages in this state.

src/Application/Actions/Donations/Confirm.php
```diff
@@ -16,9 +17,14 @@
...
 use Ramsey\Uuid\Uuid;
 use Slim\Exception\HttpBadRequestException;
+use Stripe\Event;
 use Stripe\Exception\ApiErrorException;
 use Stripe\Exception\CardException;
 use Stripe\Exception\InvalidRequestException;
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\RoutableMessageBus;
+use Symfony\Component\Messenger\Stamp\BusNameStamp;
+use Symfony\Component\Messenger\Stamp\TransportMessageIdStamp;
```
NoelLH, line 27:
```suggestion
```

NoelLH (dismissed):
Looks really good! Just small things inline

Do you think we would keep the push command around but maybe make it less frequent (e.g. daily overnight) if we implement resends on nack and they seem to work well?

src/Domain/SalesforceWriteProxyRepository.php
bdsl: Shall we keep it like this for now and be ready to change it if it becomes noisy? I'd definitely like to have extra visibility of what this is doing immediately after this goes to prod for the first time. Maybe we can / should make it less noisy by making this only activate when the backlog gets longer, e.g. replace fiveMinutesAgo with 20 minutes ago or an hour ago.

tests/Application/Commands/RetrospectivelyMatchTest.php
```diff
@@ -5,39 +5,47 @@
...
 use Prophecy\Argument;
+use Prophecy\Prophecy\ObjectProphecy;
 use Psr\Log\NullLogger;
 use Symfony\Component\Console\Tester\CommandTester;
 use Symfony\Component\Lock\LockFactory;
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\RoutableMessageBus;
 use Symfony\Component\Notifier\ChatterInterface;
 
+use const true;
```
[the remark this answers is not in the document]
bdsl: Nope

```suggestion
```

tests/TestCase.php
```diff
@@ -189,6 +191,20 @@ public static function someCampaign(?string $stripeAccountId = null): Campaign
         return $campaign;
     }
 
+    public static function someDonation(): Donation
+    {
+        return Donation::fromApiModel(new DonationCreate(
+            currencyCode: 'GBB',
```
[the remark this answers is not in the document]
bdsl: Thanks, pushed in 5cc0d5d. There were no test assertions to update, I added an assertion to prod code.

src/Domain/SalesforceWriteProxyRepository.php
NoelLH: Yep, thought about it a bit more and if a donation awaiting a push is this old I think there should probably be log noise from previous push attempts anyway. So one more message could help and is unlikely to hurt anything.

PR #867, opened 2024-05-31
Type fixes

bdsl:
As well as fixing the issues Psalm considers most server, this increases our type coverage from 98.8430% to 98.9690%.


dorota-joanna (approved):
LGTM

PR #873, opened 2024-06-05
MAT-370 – patch fund

NoelLH:


NoelLH:
Got issues confused, closing in favour of https://github.com/thebiggive/matchbot/pull/874 so at least some of the work links to Jira more helpfully

PR #874, opened 2024-06-05
MAT-369 – patch fund

NoelLH:


src/Application/Commands/HandleOutOfSyncFunds.php
```diff
@@ -105,6 +116,26 @@ protected function doExecute(InputInterface $input, OutputInterface $output): in
                 $numFundingsOvermatched++;
                 $overmatchAmount = bcsub($fundingWithdrawalTotal, $campaignFundingAllocated, 2);
                 $output->writeln("Funding {$funding->getId()} is over-matched by $overmatchAmount. $details");
+
+                if ($mode === '2024-06-05-patch' && in_array($funding->getId(), $problemIds, true)) {
+                    // Over match expected to be £300 but safest to drive it from live data and also assert.
+                    \assert(bccomp($overmatchAmount, '300.00', 2) === 0);
```
bdsl, line 122: for production safety it's better to use `\MatchBot\Application\Assertion`, rather than `assert`. `assert` is intended just for dev usage and can be turned off in production.

NoelLH: Done in 85837c5a972dddaf4e197fb7f5db64e4019bb923

src/Application/Commands/HandleOutOfSyncFunds.php
```diff
@@ -74,8 +77,17 @@ protected function doExecute(InputInterface $input, OutputInterface $output): in
         $numFundingsCorrect = 0;
         $numFundingsOvermatched = 0;
         $numFundingsUndermatched = 0;
+
+        $problemIds = [];
         /** @var CampaignFunding[] $fundings */
-        $fundings = $this->campaignFundingRepository->findAll();
+        if ($mode === '2024-06-05-patch') {
+            $problemIds = [29822];
```
bdsl, line 84: doesn't need to be changed now but $problemIds is always the same when it's used, it could be moved outside of conditionals or even made a class constant, to avoid having to set it to empty just for the sake of Psalm. Could also be renamed to something like `$mat369_problemFundingIds`. I don't think it's worth actually making either of these changes now since this is just temporary code, just noting here to get shared understanding of what we're doing.

src/Application/Commands/HandleOutOfSyncFunds.php
```diff
@@ -105,6 +117,26 @@ protected function doExecute(InputInterface $input, OutputInterface $output): in
                 $numFundingsOvermatched++;
                 $overmatchAmount = bcsub($fundingWithdrawalTotal, $campaignFundingAllocated, 2);
                 $output->writeln("Funding {$funding->getId()} is over-matched by $overmatchAmount. $details");
+
+                if ($mode === '2024-06-05-patch' && in_array($funding->getId(), $problemIds, true)) {
+                    // Over match expected to be £300 but safest to drive it from live data and also assert.
```
bdsl, line 122: :+1:

src/Application/Commands/HandleOutOfSyncFunds.php
NoelLH: Yeah I was just going for the shortest path from the last time we did a somewhat similar change – since as you say we will hopefully delete it again by tomorrow.

bdsl (approved):
LGTM

src/Application/Commands/HandleOutOfSyncFunds.php
```diff
@@ -74,8 +77,17 @@ protected function doExecute(InputInterface $input, OutputInterface $output): in
...
         $numFundingsUndermatched = 0;
+
+        $problemIds = [];
         /** @var CampaignFunding[] $fundings */
-        $fundings = $this->campaignFundingRepository->findAll();
+        if ($mode === '2024-06-05-patch') {
+            $problemIds = [29822];
+            $output->writeln('Running in 2024-06-05-patch mode');
+            // https://stackoverflow.com/a/52427915/2803757
+            $fundings = $this->campaignFundingRepository->findBy(['id' => $problemIds]);
```
bdsl, line 87: @NoelLH are you sure findBy works with an array of ids rather than a single ID?

We could simplify this line to
`$this->campaignFundingRepository->find(29822);`

bdsl: Looks like this is good. Hard to find actual docs on what criteria can be for findBy, but the implementation isn't that hard to follow, it does test if the $problemIds is an array and generate an SQL `IN` query if so:

image

NoelLH: Yep, I think I vaguely remember checking this last time – and moreover we saw it work live then. But definitely good to double check, thanks!

PR #876, opened 2024-06-05
Clarify fee logic and docs

NoelLH:
Per discussions yesterday

### DIFF
--- a/src/Application/Fees/Calculator.php
+++ b/src/Application/Fees/Calculator.php
@@ -120,6 +120,7 @@ class Calculator
 
             $currencyCode = strtoupper($this->currencyCode); // Just in case (Stripe use lowercase internally).
             // Currency code has been compulsory for some time.
+            /** @psalm-suppress ImpureMethodCall */
             Assertion::keyExists(self::FEES_FIXED, $currencyCode);
             $feeAmountFixed = self::FEES_FIXED[$currencyCode];
 
--- a/src/Application/Fees/Calculator.php
+++ b/src/Application/Fees/Calculator.php
@@ -21,18 +21,20 @@ class Calculator
 
     private const string STRIPE_FEE_MAIN_PERCENTAGE_STANDARD = '1.5';
 
+    /** @var string[]   Whole currency unit (e.g. pounds) fee charged *by us* for Stripe credit/debit
+     *                  card donations. These values were chosen based on a Stripe support email about
+     *                  their own core fees in mid 2021 BUT they don't necessarily reflect what Stripe
+     *                  charge *us* due to special contract arrangements.
+     */
     private const array STRIPE_FEES_FIXED = [
-        // Based on Stripe support email 9/4/21.
         'CHF' => '0.3',
         'DKK' => '1.8',
         'EUR' => '0.25',
-        'GBP' => '0.2', // Baseline fee in pounds
+        'GBP' => '0.2', // Baseline fee in pounds for recharge; not necessarily exactly what Stripe charged BG.
         'NOK' => '1.8',
         'SEK' => '1.8',
         'USD' => '0.3',
     ];
-
-    private const string STRIPE_FEES_DEFAULT = '0.2';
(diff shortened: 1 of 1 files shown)


src/Application/Fees/Calculator.php
```diff
@@ -118,11 +120,9 @@ private function getCoreFee(): string
             // a fee on Gift Aid. May vary by card type & country.
 
             $currencyCode = strtoupper($this->currencyCode); // Just in case (Stripe use lowercase internally).
-            if (array_key_exists($currencyCode, self::STRIPE_FEES_FIXED)) {
-                $feeAmountFixed = self::STRIPE_FEES_FIXED[$currencyCode];
-            } else {
-                $feeAmountFixed = self::STRIPE_FEES_DEFAULT;
-            }
+            // Currency code has been compulsory for some time.
+            \assert(array_key_exists($currencyCode, self::STRIPE_FEES_FIXED));
```
NoelLH, line 124: I'd like to do
```suggestion
            Assertion::keyExists(self::STRIPE_FEES_FIXED, $currencyCode);
```

But that seems to give me Psalm warnings of https://psalm.dev/docs/running_psalm/issues/ImpureMethodCall/ – even if I take `#[Pure]` out of this class entirely.

bdsl: @NoelLH I think it's because of the `@psalm-immutable` on the class, psalm expects all methods to be pure. Maybe `\Assert\Assertion::keyExists` should be marked as pure, I can't remember whether throwing is considered pure or not.

But in this case I think I'd just do:

```suggestion
            /** @psalm-suppress ImpureMethodCall */
            Assertion::keyExists(self::STRIPE_FEES_FIXED, $currencyCode);
```

We know that keyExists is pure enough for our purposes.

src/Application/Fees/Calculator.php
```diff
@@ -21,19 +21,21 @@ class Calculator
 
     private const string STRIPE_FEE_MAIN_PERCENTAGE_STANDARD = '1.5';
 
+    /** @var string[]   Whole currency unit (e.g. pounds) fee charged *by us* for Stripe credit/debit
```
bdsl, line 24:
```suggestion
    /** @var string[]   Major currency unit (e.g. pounds) fee charged *by us* for Stripe credit/debit
```

Major / minor seems to be the standard jargon for things like pounds and pence.

src/Application/Fees/Calculator.php
```diff
@@ -21,19 +21,21 @@ class Calculator
...
+     *                  card donations. These values were chosen based on a Stripe support email about
+     *                  their own core fees in mid 2021 BUT they don't necessarily reflect what Stripe
+     *                  charge *us* due to special contract arrangements.
+     */
     private const array STRIPE_FEES_FIXED = [
```
bdsl, line 29: Not a blocker as the other changes are good, but maybe worth making things clearer still by renaming this? e.g. from `STRIPE_FEES_FIXED` to just `FEES_FIXED` to further avoid confusion between fee charged to us and fee charged by us. Probably goes for the two other `STRIPE` constants - the point of the calculator is all about deciding what we charge, not what we pay.

NoelLH: Makes sense, done in https://github.com/thebiggive/matchbot/pull/876/commits/2ee4436a529d434dafb919704d2fb716a8eccf9e

bdsl (approved):
LGTM. Wondering if it's also worth trying to get this whole class as easy to understand as we can at this point, which IMHO would mean getting rid of the instance properties and making all methods static.

bdsl:
Also sort of want to point how much harder I think these sort of clarifications and understanding would be in the old version of the calculator that refereed to settings outside the source code for fees: https://github.com/thebiggive/matchbot/blob/7b5275a9f1a0cf2a4ecf5dffb5487ab7ab7771e9/src/Application/Fees/Calculator.php

PR #878, opened 2024-06-05
Remove (old, incomplete) support for creating a donation associated with no Stripe customer

NoelLH:


bdsl:
Can we also update the signature and docblock of `\MatchBot\Domain\DonationService::createDonation` ? That still implies that customer id is optional. And then worth taking it back to https://github.com/thebiggive/matchbot/blob/96355705fdd13761f045c771527022528849aaab/src/Application/Actions/Donations/Create.php#L50 and requiring the id to be not null when we take it off the customer attribute.

Otherwise the code is still confusing about whether this is required or not.

bdsl:
Nice simplification! I made a PR to this PR: https://github.com/thebiggive/matchbot/pull/879

If that is merged all LGTM.

bdsl (approved):

PR #880, opened 2024-06-05
MAT-370 – explicitly avoid the result cache when a campaign was initially missing

NoelLH:
Also replace deprecated use of `EntityManager::create`, and maybe fix some non-critical Doctrine options not applying as intended.

Just an idea for now. I think that if we think this is worthwhile it should maybe happen in concert with the equivalent changes to existing fund checks in `FundRepository::pullForCampaign()`.


NoelLH:
This PR makes it compulsory to set `APP_ENV` to 'test' in CI, because while we dummy it for phpunit that doesn't work for actually running migrations ahead of integration tests. Added that to our CircleCI project.

src/Domain/CampaignRepository.php
```diff
@@ -14,6 +14,23 @@
  */
 class CampaignRepository extends SalesforceReadProxyRepository
 {
+    public function getCampaignBySalesforceId(string $salesforceId, bool $withResultCache): ?Campaign
```
bdsl, line 17: probably for simplicity we could set 
```suggestion
    public function getCampaignBySalesforceId(string $salesforceId, bool $withResultCache = true): ?Campaign
```

and only pass the argument when it's false. Otherwise it's not obvious unless you know Doctrine better than I do that using the result cache is the default, and we only disable it an edge case semi-unhappy path. 

Same applies to the param of `\MatchBot\Domain\DonationRepository::buildFromApiRequest` which is only called twice from prod code, with false just after the `UniqueConstraintViolationException` is caught.

NoelLH:
We merged the part of this that was definitely good on https://github.com/thebiggive/matchbot/pull/884

For the rest, I am now convinced there was no result cache in play anyway so it's redundant. Will close.

PR #881, opened 2024-06-05
MAT-370 – fix `DonationService::runWithPossibleRetry()` silently swallowing errors

NoelLH:
Prevent problems from donation create actions continuing after persistent errors. And add a test for `Donations\Create` specifically to verify that errors are surfaced in API responses and that exactly the maximum # of attempts are made at persistence before giving up.

### DIFF
--- a/src/Application/Actions/Donations/Create.php
+++ b/src/Application/Actions/Donations/Create.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace MatchBot\Application\Actions\Donations;
 
+use Doctrine\DBAL\Exception\ServerException as DBALServerException;
+use Doctrine\ORM\Exception\ORMException;
 use MatchBot\Application\Actions\Action;
 use MatchBot\Application\Actions\ActionError;
 use MatchBot\Application\Actions\ActionPayload;
@@ -137,6 +139,19 @@ class Create extends Action
                     new ActionError(
                         ActionError::SERVER_ERROR,
                         'Could not make Stripe Payment Intent (B)'
+                    ),
+                ),
+            );
+        } catch (ORMException | DBALServerException $ex) {
+            // '(D)' errors are DB persistence issues, typically ones that still exist after some retries.
+            return $this->respond(
+                $response,
+                new ActionPayload(
+                    500,
+                    null,
+                    new ActionError(
+                        ActionError::SERVER_ERROR,
+                        'Could not make Stripe Payment Intent (D)'
                     ),
                 ),
             );
--- a/src/Domain/DonationService.php
+++ b/src/Domain/DonationService.php
@@ -25,7 +25,7 @@ use Symfony\Component\Notifier\Message\ChatMessage;
 
 readonly class DonationService
 {
-    private const MAX_RETRY_COUNT = 4;
+    private const MAX_RETRY_COUNT = 3;
 
     public function __construct(
         private DonationRepository $donationRepository,
@@ -80,7 +80,7 @@ readonly class DonationService
             throw new CampaignNotOpen("Campaign {$donation->getCampaign()->getSalesforceId()} is not open");
         }
 
-        // Must persist before Stripe work to have ID available.
+        // Must persist before Stripe work to have ID available. Outer fn throws if all attempts fail.
         $this->runWithPossibleRetry(function () use ($donation) {
             $this->entityManager->persistWithoutRetries($donation);
             $this->entityManager->flush();
@@ -260,6 +260,7 @@ readonly class DonationService
      * then we still have any required related new entities in the Unit of Work.
      * @param \Closure $retryable The action to be executed and then retried if necassary
      * @param string $actionName The name of the action, used in logs.
+     * @throws ORMException|DBALServerException if they're occurring when max retry count reached.
      */
     private function runWithPossibleRetry(\Closure $retryable, string $actionName): void
     {
@@ -282,6 +283,18 @@ readonly class DonationService
                 $seconds = (new Randomizer())->getFloat(0.1, 1.1);
                 \assert(is_float($seconds)); // See https://github.com/vimeo/psalm/issues/10830
                 $this->clock->sleep($seconds);
+
+                if ($retryCount === self::MAX_RETRY_COUNT) {
+                    $this->logger->error(
+                        sprintf(
+                            $actionName . ' error: %s. Giving up after %d retries.',
+                            $exception->getMessage(),
+                            self::MAX_RETRY_COUNT,
+                        )
+                    );
+
+                    throw $exception;
+                }
             }
         }
     }
--- a/tests/Application/Actions/Donations/CreateTest.php
+++ b/tests/Application/Actions/Donations/CreateTest.php
@@ -877,36 +877,23 @@ class CreateTest extends TestCase
     {
         $donation = $this->getTestDonation(true, true, true);
 
-        $app = $this->getAppInstance();
+        $app = $this->getAppWithCommonPersistenceDeps(
+            donationPersisted: false,
+            donationPushed: false,
+            donationMatched: false,
+            donation: $donation,
+        );
         $container = $app->getContainer();
         \assert($container instanceof Container);
-        $donationRepoProphecy = $this->prophesize(DonationRepository::class);
-
-        // Use a custom Prophecy Promise to vary the simulated behaviour.
-        $donationRepoProphecy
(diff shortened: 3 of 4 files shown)


src/Domain/DonationService.php
```diff
@@ -25,7 +25,7 @@
 
 readonly class DonationService
 {
-    private const MAX_RETRY_COUNT = 4;
+    private const MAX_RETRY_COUNT = 3;
```
NoelLH, line 28: Given some failure modes have slightly increased negative impacts by retrying a lot of times, and I can't remember seeing anything in donation setup succeeding after a 3rd time or later, I think 3 is a good number for now. Not set on the change though!

src/Application/Actions/Donations/Create.php
```diff
@@ -140,6 +142,19 @@ protected function action(Request $request, Response $response, array $args): Re
                     ),
                 ),
             );
+        } catch (ORMException | DBALServerException $ex) {
```
bdsl, line 145: don't we need to make an error log / alarm if we catch these exceptions?

src/Domain/DonationService.php
```diff
@@ -282,6 +283,18 @@ private function runWithPossibleRetry(\Closure $retryable, string $actionName):
...
+                    $this->logger->error(
+                        sprintf(
+                            $actionName . ' error: %s. Giving up after %d retries.',
+                            $exception->getMessage(),
+                            self::MAX_RETRY_COUNT,
+                        )
+                    );
+
+                    throw $exception;
+                }
```
bdsl, line 297: this LGTM - equivalent to wrapping the entire runWithPossibleRetry in another try/catch just to log the final error in case all tries fail, making it clear that it failed after the retries.

src/Application/Actions/Donations/Create.php
NoelLH: There should be two alarms so far:
1. the new "... Giving up ..." log in `DonationService`
2. the return status 500 alarm

It won't have a full trace but I think maybe for this type of exception the message is enough for it to be quite clear. Open to suggestions on tweaking it though.

tests/Application/Actions/Donations/CreateTest.php
```diff
@@ -871,6 +873,46 @@ public function testSuccessWithMinimalData(): void
         $this->assertEquals('123CampaignId12345', $payloadArray['donation']['projectId']);
     }
 
+    public function testErrorWhenAllDbPersistCallsFail(): void
```
bdsl, line 876: the code you added to the Create controller is very simple and IMHO doesn't need testing. But the if condition in the DonationService probably does. I feel like we'd probably be able to write a significantly simpler and clearer test if it was a test for the DonationService, rather than a test for the Create controller. 

This test also doesn't mutation-cover the call to `$this->logger->error`, which I think is the main point of this PR. @NoelLH what do you think about testing the DonationService instead?

bdsl (approved):
LGTM but please also see comments inline.

tests/Application/Actions/Donations/CreateTest.php
bdsl: https://output.circle-artifacts.com/output/job/58cf5124-bb3d-46de-8f7d-97c369b00810/artifacts/0/reports/mutation-testing-all-files/infection.html#mutant/Domain/DonationService.php 

image

NoelLH: Good point. I would definitely prefer to do both, I think it's good to also cover the end to end behaviour for such an important endpoint. Agree that the greater value is in testing `DonationService`.

src/Application/Actions/Donations/Create.php
bdsl: ah ok, I didn't realise returning 500 automatically generated an alarm. That should be fine then. I don't quite like relying on an alarm having been logged earlier in a different class when we catch an exception.

tests/Application/Actions/Donations/CreateTest.php
bdsl: @NoelLH I think covering the end to end happy path behaviour is important, I don't really see the value in testing at this level for the error case.

tests/Domain/DonationServiceTest.php
```diff
@@ -88,6 +70,71 @@ public function testIdentifiesCharityLackingCapabilities(): void
...
-        $this->sut->createDonation($donationCreate, $customerId);
+        $this->sut->createDonation($donationCreate, self::CUSTOMER_ID);
+    }
+
+    public function testInitialPersistRunsOutOfRetries(): void
+    {
+        $logger = $this->prophesize(LoggerInterface::class);
+        $logger->info('Donation Create persist before stripe work error: . Retrying 1 of 3.')->shouldBeCalledOnce();
+        $logger->info(Argument::type('string'))->shouldBeCalled();
+        $logger->error('Donation Create persist before stripe work error: . Giving up after 3 retries.')
```
NoelLH, line 81: Potentially a not-worthwhile amount of work to make the exception more realistic and still of this type?

Trying to do it the normal Prophecy way gives
> Can not add prophecy for a method `Double\Doctrine\DBAL\Exception\UniqueConstraintViolationException\P37::getMessage()`
as it is a final method.

bdsl: Yeah it's awkward because the exception constructor is internal and its final. Maybe that's a hint we should be catching something more general - not sure. But if you don't mind the psalm errors this way gives us a test that checks we're passing the message through:

```diff
--- a/tests/Domain/DonationServiceTest.php
+++ b/tests/Domain/DonationServiceTest.php
@@ -2,6 +2,7 @@
 
 namespace MatchBot\Tests\Domain;
 
+use Doctrine\DBAL\Driver\PDO\Exception;
 use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 use MatchBot\Application\HttpModels\DonationCreate;
 use MatchBot\Application\Matching\Adapter;
@@ -76,9 +77,9 @@ class DonationServiceTest extends TestCase
     public function testInitialPersistRunsOutOfRetries(): void
     {
         $logger = $this->prophesize(LoggerInterface::class);
-        $logger->info('Donation Create persist before stripe work error: . Retrying 1 of 3.')->shouldBeCalledOnce();
+        $logger->info('Donation Create persist before stripe work error: An exception occurred in the driver: EXCEPTION_MESSAGE. Retrying 1 of 3.')->shouldBeCalledOnce();
         $logger->info(Argument::type('string'))->shouldBeCalled();
-        $logger->error('Donation Create persist before stripe work error: . Giving up after 3 retries.')
+        $logger->error('Donation Create persist before stripe work error: An exception occurred in the driver: EXCEPTION_MESSAGE. Giving up after 3 retries.')
             ->shouldBeCalledOnce();
 
         $this->sut = $this->getDonationService(withAlwaysCrashingEntityManager: true, logger: $logger->reveal());
@@ -101,8 +102,9 @@ class DonationServiceTest extends TestCase
     ): DonationService {
         $emProphecy = $this->prophesize(RetrySafeEntityManager::class);
         if ($withAlwaysCrashingEntityManager) {
-            $exception = $this->prophesize(UniqueConstraintViolationException::class);
-            $emProphecy->persistWithoutRetries(Argument::type(Donation::class))->willThrow($exception->reveal());
+            $emProphecy->persistWithoutRetries(Argument::type(Donation::class))->willThrow(
+                new UniqueConstraintViolationException(new Exception('EXCEPTION_MESSAGE'), null)
+            );
         }
 
```

I need to eat a bit, lets have a call later and get this merged together?

Also for the logger I feel like an in-memory logger would make the test quite a bit simpler to read than mocking the calls to it? Can easily write our own or use one off the shelf like  https://github.com/MetaSyntactical/inmemory-logger/blob/2.2.x/src/InMemoryLogger.php  (I don't remember where I found that one now, no idea if it's the best)

tests/Application/Actions/Donations/CreateTest.php
bdsl: Maybe - I think these sort of tests are not all that cheap to own, they tend to slow down changes quite a lot.

tests/Domain/DonationServiceTest.php
NoelLH: I'll be offline before the end of your current session I think @bdsl but suppressing a psalm error sounds OK in this case. Integrating a different test logger seems like it would make sense to do as a later enhancement to me.

bdsl (approved):

PR #882, opened 2024-06-05
MAT-370 – ensure EntityManager's open right after campaign data preparation

NoelLH:


src/Domain/DonationService.php
```diff
@@ -80,6 +80,14 @@ public function createDonation(DonationCreate $donationData, string $pspCustomer
             throw new CampaignNotOpen("Campaign {$donation->getCampaign()->getSalesforceId()} is not open");
         }
 
+        // A closed EM can happen if the above tried to insert a campaign or fund, hit a duplicate error because
+        // another thread did it already, then successfully got the new copy. There's been no subsequent
+        // database persistence that needed an open manager, so none replaced the broken one. In that
+        // edge case, we need to handle that before `persistWithoutRetries()` has a chance of working.
+        if (!$this->entityManager->isOpen()) {
+            $this->entityManager->resetManager();
```
bdsl, line 88: I think reset manager is going to mean we lose track of any entities, that we might have made changes to and would otherwise expect to be persisted as part of the unit of work. Can we know from inside DonationService that that isn't a problem?

Maybe this is an argument against having created the donation service. We can see right now that there's nothing before this in `\MatchBot\Application\Actions\Donations\Create::action` that would need to be persisted, but it isn't safe to assume that that will stay that way or that that will always be the only caller to the service method. 

I think if the EM is closed it's probably safer to either just let the script crash, or if we have to reset the manager from the controller where we have a bit more of an overview.

bdsl: Also thinking about if the issue is two requests trying to create the same campaign - is there any reason we can't do an upsert instead of insert? 

I think the ORM can't do an upsert  https://github.com/doctrine/orm/issues/5178 but we should be able to upsert it with SQL code fairly easily.

bdsl: Or I'm wondering if we could catch the unique constraint exception inside \MatchBot\Domain\CampaignRepository::pullNewFromSf

If that's thrown it implies the campaign is already in our DB so we can find it and return there.

NoelLH: I think that the recovery from the unique constraint crash already works (as far as getting a Campaign in memory), but leaves the EM unusable. So I think catching in a different method would leave us with the same amount of cleaning up to think about?

Upserting without the ORM is definitely interesting. In principle yes, but as we have some ORM-managed cascades to funds etc. I think it might introduce more risk than it fixes at this point.

In this particular case I think that a closed EM suggests that we have a Campaign in memory, and that we didn't just persist it but rather queried an existing one. So while it's true that this wouldn't be safe to do in other situations it doesn't feel intrinsically that risky to me. Maybe we can add more assertions to satisfy ourselves that we are in that exact situation before we do an EM reset? I think moving the catch to the controller would be OK but not necessarily better than this is we only know about one very narrow case where it's necessary.

I'd much rather not bail out entirely if we know we have workable methods to recover and avoid a 500.

bdsl: @NoelLH OK - shall we just just add a note to the dockblock of createDonation to warn that it has a possibility to reset the EM and flush the EM while losing any entities that were previously tracked in the UoW?

bdsl (approved):

src/Domain/DonationService.php
NoelLH: Yes, that makes sense @bdsl – will add and then merge.

PR #892, opened 2024-06-13
MAT-363 – re-apply work to queue Salesforce pushes, log more, and lock Donations proactively

NoelLH:
Reverts thebiggive/matchbot#891 and adds a fix.

It's possible that we will see some level of errors from locks not being acquired and will need to add retries, but I think we can try it without first.


dorota-joanna (approved):
LGTM

PR #893, opened 2024-06-13
MAT-363 – attempted extra Donation locking

NoelLH:
WIP. Like https://github.com/thebiggive/matchbot/pull/892 plus locks that seemed prudent, but in one case also break integration tests.


bdsl:
@NoelLH is this still relevant?

NoelLH:
> @NoelLH is this still relevant?

Maybe. I need to review the risk of not locking in the extra places.

NoelLH:
Superseded by https://github.com/thebiggive/matchbot/pull/1365

PR #916, opened 2024-06-21
MAT-363 – patch SF metadata together; fix test broken in last `develop` push

NoelLH:


src/Domain/DonationRepository.php
```diff
@@ -43,12 +42,11 @@ public function setMatchingAdapter(Matching\Adapter $adapter): void
         $this->matchingAdapter = $adapter;
     }
 
-    public function doCreate(SalesforceWriteProxy $proxy): bool
+    public function doCreate(SalesforceWriteProxy $proxy): ?string
     {
         $donation = $proxy;
         try {
             $salesforceDonationId = $this->getClient()->create($donation);
-            $donation->setSalesforceId($salesforceDonationId);
```
bdsl, line 51: is deleting this to avoid making the entity appear dirty to the EM? I think that makes sense with the other changes.

PR #922, opened 2024-06-25
MAT-363 – Salesforce push queueing with much less database state mgmt

NoelLH:


NoelLH:
WIP – couple of unit test issues and a circular dependency problem in integration tests

NoelLH:
Tests and DI sorted for the most part, but some caveats block merging until some final work is done I think. I'll try to comment on those points.

integrationTests/DonationRepositoryTest.php
```diff
@@ -249,7 +255,11 @@ public function donationAgesDataProvider(): array
         return [
             [0, false],
             [299, false],
-            [300, true],
+            // TODO Even with a prophesised response, this last data set gives
```
NoelLH, line 258: Not clear why this scenario is failing when uncommented – the prophesised bus should cover both calling and not calling dispatch

src/Application/Messenger/DonationUpdated.php
```diff
@@ -0,0 +1,31 @@
...
+use MatchBot\Domain\Donation;
+
+class DonationUpdated extends AbstractStateChanged
+{
+    private function __construct(
+        public string $uuid,
+        public ?string $salesforceId,
+        public array $json,
+    ) {
+        // TODO Things like Stripe hooks could come back before pushing to Salesforce is complete, so we need to be
```
NoelLH, line 15: Realised late due to integration tests that asserting this isn't blank is probably a design flaw, because Stripe hooks might come in before SF ID is set. I guess one option is to deliberately error and HTTP 500 on those so that Stripe resend but that feels unnecessarily brittle. We might need to think about bring back a recovery mechanism where we do talk to the database if we get an update job with no SF ID. Or possibly pause for a few seconds and re-queue it?

tests/Domain/DonationRepositoryTest.php
```diff
@@ -42,145 +46,40 @@ class DonationRepositoryTest extends TestCase
...
-        $donation->setSalesforcePushStatus(SalesforceWriteProxy::PUSH_STATUS_PENDING_UPDATE);
 
-        $success = $this->getRepo($donationClientProphecy)->push($donation, false);
-
-        // We let push() handle both steps for older-than-30s donations, without waiting for a new process.
-        $this->assertTrue($success);
-        $this->assertEquals('complete', $donation->getSalesforcePushStatus());
-    }
-
-    public function testExistingPushWithMissingProxyIdButPendingUpdateStatusNew(): void
```
NoelLH, line 143: I deleted this one (I think, diff has gone a bit weird) because I don't think the same forgiving create then update logic happens now. But this might also be a problem, per the TODO near where I wanted to assert that the SF ID is set.

src/Application/Messenger/DonationUpdated.php
bdsl: I'm also not sure if you expected the id to always be not null why you would use `Assertion::notNull($this->salesforceId);` rather than typing it as `string` instead of `?string`

bdsl:
Closing for now as we've paused this work, but will add link on the ticket so we can re-open.

integrationTests/DonationRepositoryTest.php
```diff
@@ -214,7 +216,9 @@ public function testPushSalesForcePendingPushesNewDonationButNotVeryNewDonation(
 
         $sut = $this->getService(DonationRepository::class);
         $donationClientProphecy = $this->prophesize(\MatchBot\Client\Donation::class);
-        /** @var Donation $donationInDB */
+
+        $busProphecy = $this->prophesize(RoutableMessageBus::class);
+        $dummyEnvelope = new Envelope(new \stdClass()); // Final so can't prophesise.
```
bdsl, line 221: yes, `\Symfony\Component\Messenger\Envelope` is final but as its a value object I think there's no reason we'd want to prophesise it.

Also $dummyEnvelope is only used once, probably simpler to inline it?

integrationTests/DonationRepositoryTest.php
```diff
@@ -230,22 +234,16 @@ public function testPushSalesForcePendingPushesNewDonationButNotVeryNewDonation(
         \assert($simulatedNow instanceof \DateTimeImmutable);
 
         // assert
-        $method = $donationClientProphecy->createOrUpdate(Argument::type(Donation::class));
+        $busDispatchMethod = $busProphecy->dispatch(Argument::type(Envelope::class))
```
bdsl, line 237: probably worth making a more specific assertion of what should be dispatched, instead of just any envelop with any possible contents. Might need to be done via a callback, since in an integration test context it might not be realistic to create an equal reference envelope to compare against.

matchbot-cli.php
```diff
@@ -71,25 +71,33 @@
...
-        $now,
-        $psr11App->get(DonationRepository::class),
-        $psr11App->get(LoggerInterface::class),
+        campaignFundingRepository: $psr11App->get(CampaignFundingRepository::class),
+        entityManager: $psr11App->get(EntityManagerInterface::class),
+        now: $now,
+        donationRepository: $psr11App->get(DonationRepository::class),
+        logger: $psr11App->get(LoggerInterface::class),
+        bus: $psr11App->get(RoutableMessageBus::class),
     ),
```
bdsl, line 80: instead of making this and similar entries longer, it might be better to replace the whole thing and let PHP-DI instantiate the commands? It can often do it all by reflection on the constructor params:

github won't let me do a suggestion on a merged PR quite reasonably, but e.g.

```php
    $psr11App->get(RedistributeMatchFunds::class),
```

src/Application/Actions/Donations/Confirm.php
```diff
@@ -44,6 +45,7 @@ public function __construct(
         private DonationRepository $donationRepository,
         private Stripe $stripe,
         private EntityManagerInterface $entityManager,
+        private RoutableMessageBus $bus,
```
bdsl, line 48: this only needs a `\Symfony\Component\Messenger\MessageBusInterface` doesn't it? Confirm doesn't care that it's routable.

src/Application/Messenger/AbstractStateChanged.php
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * For now, this is named with the idea that `SalesforceWriteProxyRepository` may soon use it for both donations
+ * and e.g. recurring donation mandates. If we decide against that we could rename it to `AbstractDonationChanged`.
+ */
+abstract class AbstractStateChanged
+{
+    /**
+     * @psalm-suppress PossiblyUnusedProperty Plan is to use `$json` with mandates
+     */
+    public array $json;
```
bdsl, line 16: $json is not a very meaningful name. How about `snapshot`? It's a snapshot of the state of the entity at the time the message was generated.

src/Domain/DonationRepository.php
```diff
@@ -37,75 +45,50 @@ class DonationRepository extends SalesforceWriteProxyRepository
...
     /** @var Donation[] Tracks donations to persist outside the time-critical transaction / lock window */
     private array $queuedForPersist;
+
     public function setMatchingAdapter(Matching\Adapter $adapter): void
     {
         $this->matchingAdapter = $adapter;
     }
 
-    public function doCreate(SalesforceWriteProxy $proxy): bool
+    public function doCreate(AbstractStateChanged $changeMessage): bool
```
bdsl, line 54: why is a donation repository claiming to be able to handle creating an abstract state changed? Shouldn't it only handle the `DonationUpserted` ?

bdsl: I see it's because this is called by `\MatchBot\Domain\SalesforceWriteProxyRepository::push`. I think the fact that we're using one type here just to let the SA pass and then a narrower type in the assertion on the next line shows we're doing something a bit odd. It does seem like we're having quite a bit of extra work with the `SalesforceWriteProxyRepository` class that only has one child, but if we want it to keep working as it does without doing weird stuff then I think the way would be to remove the DonationUpserted class and replace it with `StateChanged<Donation>` . That allows \MatchBot\Domain\SalesforceWriteProxyRepository::push to have a param type of `StateChanged<T>`. Alternatively we could add a third template variable to SalesforceWriteProxyRepository, e.g. `@template S of AbstractStateChanged` and then the param type would just be `S`.

tests/Application/Actions/Donations/UpdateHandlesLockExceptionTest.php
```diff
@@ -83,17 +76,11 @@ public function testRetriesOnUpdateToCancelledLockException(): void
 
         $this->setExpectationsForPersistAfterRetry($donationId, $donation, DonationStatus::Cancelled);
 
-        $this->donationRepositoryProphecy->push($donation, false)->shouldBeCalled()->willReturn(true);
+        // TODO expect something on bus instead
+//        $this->donationRepositoryProphecy->push($upsertMessage, false)->shouldBeCalled();
```
bdsl, line 80: is this todo still relevant?  Do we need it on the ticket or something if so?

tests/Application/Messenger/Handler/DonationUpsertedHandlerTest.php
```diff
@@ -0,0 +1,58 @@
...
+    public function testItPushesOneDonationToSf(): void
+    {
+        $this->donationRepositoryProphecy->push(Argument::type(DonationUpserted::class), false)->shouldBeCalledOnce();
+
+        $sut = new DonationUpsertedHandler(
+            $this->donationRepositoryProphecy->reveal(),
+            new NullLogger(),
+        );
+
+        $sut->__invoke(self::someUpsertedMessage());
```
bdsl, line 36: Can we assert on the particular message we push, not just the type?

e.g.

```php
        $message = self::someUpsertedMessage();
        $this->donationRepositoryProphecy->push($message, false)->shouldBeCalledOnce();

        $sut = new DonationUpsertedHandler(
            $this->donationRepositoryProphecy->reveal(),
            new NullLogger(),
        );

        $sut->__invoke($message);
```

It doesn't really matter whether we make the same change or not in the next test.

src/Domain/DonationRepository.php
```diff
@@ -667,24 +649,69 @@ public function removeAllFundingWithdrawalsForDonation(Donation $donation): void
...
+    /**
+     * Re-queues proxy objects to Salesforce en masse.
+     *
+     * By using FIFO queues and deduplicating on UUID if there are multiple consumers, we should make it unlikely
+     * that Salesforce hits Donation record lock contention issues.
+     *
+     * @param RoutableMessageBus $bus   We had to not DI this to avoid a circular dependency.
+     * @return int  Number of objects pushed
+     */
+    public function pushSalesforcePending(\DateTimeImmutable $now, RoutableMessageBus $bus): int
```
bdsl, line 661: I think I remember couldn't inject the bus via the constructor because that would make a circular dependency, but I think we could probably inject it via a setter as we do for some other dependencies. Maybe simpler than injecting just to this method? 

Also we could widen the type to `MessageBusInterface`.

src/Domain/DonationRepository.php
bdsl: There's also a much easier option - change the parameter types in `SalesforceWriteProxyRepository` from `AbstractStateChanged` to `DonationUpserted`. Of course that means those methods can't be used with anything other than a donation, but it doesn't currently need to be, and we could deal with that when we come to it.

matchbot-cli.php
NoelLH: I don't think this is possible until `$now` is replaced with something DI-able? As the changes were already getting big I was trying to avoid refactors of that sort.

src/Domain/DonationRepository.php
NoelLH: I think because it's a Doctrine repo, DI in the constructor which has to follow a very specific signature isn't really possible? I'm not really sure about the accuracy of my circular deps comment, I think possibly that was incorrectly migrated from when the logic lived somewhere else.

Agree about widening the type, done in c0d6d320d245fad4e77560e4cc4ff68e6d4fc19a. I'm not sure about adding a method to set the service if exactly one thing uses it, seems a bit like adding complexity earlier than it's helpful to me.

src/Application/Actions/Donations/Confirm.php
NoelLH: Done for this class in c0d6d320d245fad4e77560e4cc4ff68e6d4fc19a

integrationTests/DonationRepositoryTest.php
NoelLH: Done in c0d6d320d245fad4e77560e4cc4ff68e6d4fc19a

src/Domain/DonationRepository.php
NoelLH: @bdsl IMO, the assertion is an OK trade-off for now until we decide (per the `AbstractStateChanged` class comment) whether we think the model is useful to expand mandates. My inclination would be to keep the workaround right now and then either use a more suitable type as you suggest if the pushing is similar enough that the abstraction is useful; or delete it if not. We'll be working on that reasonably soon I hope.

tests/Application/Messenger/Handler/DonationUpsertedHandlerTest.php
NoelLH: Done in e7d63af6c540d6c7b4bc2a9273a3bbbd50e74c66

integrationTests/DonationRepositoryTest.php
NoelLH: Done in 212c74235d9ea1a531081a44888c64e24cedc757

PR #924, opened 2024-07-03
Mat 364 card testing prevention

bdsl:
See Symfony Rate Limiter docs here: https://symfony.com/doc/current/rate_limiter.html . They do somewhat assume you're using Symfony framework though.

I'm planning to make a PR to frontend display a spinner if the rate limit is hit, but I think this is fine without it - you'll just get our generic something went wrong / please try again message when trying to move on from the first step in the stepper if you've exhausted the rate limit.

### DIFF
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -61,6 +61,8 @@ use Symfony\Component\Messenger\Transport\TransportInterface;
 use Symfony\Component\Notifier\Bridge\Slack\SlackTransport;
 use Symfony\Component\Notifier\Chatter;
 use Symfony\Component\Notifier\ChatterInterface;
+use Symfony\Component\RateLimiter\RateLimiterFactory;
+use Symfony\Component\RateLimiter\Storage\CacheStorage;
 use Symfony\Component\Serializer\Encoder\JsonEncoder;
 use Symfony\Component\Serializer\Normalizer\BackedEnumNormalizer;
 use Symfony\Component\Serializer\Normalizer\ObjectNormalizer;
@@ -68,6 +70,7 @@ use Symfony\Component\Serializer\Serializer;
 use Symfony\Component\Serializer\SerializerInterface;
 
 return function (ContainerBuilder $containerBuilder) {
+    $donationCreateRateLimiterKey = 'donation-creation-rate-limiter-factory';
     $containerBuilder->addDefinitions([
         Auth\DonationPublicAuthMiddleware::class =>
             function (ContainerInterface $c): Auth\DonationPublicAuthMiddleware {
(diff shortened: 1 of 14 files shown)


src/Application/Persistence/RetrySafeEntityManager.php
```diff
@@ -167,14 +167,14 @@ public function refresh($object, ?int $lockMode = null): void
     public function flush($entity = null): void
     {
         try {
-            $this->entityManager->flush($entity);
+            $this->entityManager->flush();
```
NoelLH, line 170: Seems counterintuitive to keep the method sig aligned with `EntityManagerDecorator`, including the `$entity`, but then do nothing with it. If there's some advantage might be good to explain the method doc block?

app/dependencies.php
```diff
@@ -234,6 +237,24 @@ public function __invoke(array $record)
             return $logger;
         },
 
+        $donationCreateRateLimiterKey => function (ContainerInterface $c): RateLimiterFactory {
+            return new RateLimiterFactory(
+                config: [
+                    'id' => 'create-donation',
+                    'policy' => 'token_bucket',
```
NoelLH, line 244: I've looked at the docs but am still confused as to what the default 'limiter' that's in effect here will be. Is it based on the direct client IP? If so that will be a problem because we're behind a load balancer.

It might also be worth thinking about the relative advantages of using a new limited as we have solved this problem once already via middleware like at https://github.com/thebiggive/matchbot/blob/824c343503f9862c1964a356b56a2348fa5cd95d/app/routes.php#L39-L40

NoelLH (commented):
Couple of things inline, let me know if you'd like to discuss @bdsl?

src/Application/Persistence/RetrySafeEntityManager.php
bdsl: This was already effectively doing nothing with the entity I think, since `\Doctrine\Persistence\ObjectManager::flush` doesn't take any parameters. Psalm reported it as an error. 

But I see thanks to your comment that the decorator interface does take the param and so does the concrete EntityManager from Doctrine. If passed that triggers the deprecation message with a link to https://github.com/doctrine/orm/issues/8459 . So I guess passing the entity doesn't do nothing as I thought, its just not recommended. 

We never pass anything when calling `flush`. Not sure why Doctrine removed the param from the interface before removing it from the decorator and the concrete class. I think I'll put it back here for now.

app/dependencies.php
bdsl: No it won't be based on client IP, it will be based on the key passed to `\Symfony\Component\RateLimiter\RateLimiterFactory::create`  in `createDonation` is the psp customer ID.

src/Application/Persistence/RetrySafeEntityManager.php
bdsl: Done in d53b31dbec4acf01fd

src/Application/Persistence/RetrySafeEntityManager.php
```diff
@@ -167,14 +167,18 @@ public function refresh($object, ?int $lockMode = null): void
...
+             */
             $this->entityManager->flush($entity);
         } catch (EntityManagerClosed $closedException) {
             $this->logger->error(
                 'EM closed. RetrySafeEntityManager::flush() trying with a new instance,' .
                 $closedException->__tostring()
             );
             $this->resetManager();
-            $this->entityManager->flush($entity);
+            $this->entityManager->flush();
```
NoelLH, line 181:
```suggestion
            $this->entityManager->flush($entity);
```
Finishes the revert for the same reasons as line 174

app/dependencies.php
NoelLH: Thanks and sorry, missed the key line! I've added a table in 8bdbdec672b5cc3899c04ddc2f4cd01dad42c3ff with a high level overview to hopefully make it easier to understand the approaches used through the app.

NoelLH (approved):

app/dependencies.php
bdsl: Welcome, no need to apologies.

PR #926, opened 2024-07-05
Bg2 2632 stripe version upgrade

bdsl:


bdsl:
Apologies I should have closed this when I started again on #927

PR #927, opened 2024-07-08
BG2-2632: Hard-code stripe api version

bdsl:
There is no reason for this to vary between environments that have the same version of matchbot code - the stripe API version used in prod should match the one used for testing, so IMHO this should be hard-coded.

Simplifes updating to a more recent version.


dorota-joanna (approved):
LGTM

app/dependencies.php
```diff
@@ -420,7 +420,8 @@ static function (ContainerInterface $c): Matching\Adapter {
         StripeClient::class => static function (ContainerInterface $c): StripeClient {
             return new StripeClient([
                 'api_key' => $c->get('settings')['stripe']['apiKey'],
-                'stripe_version' => $c->get('settings')['stripe']['apiVersion'],
+                // Stripe version determined by defaults built into stripe-php library - upgrade the library to get
+                // a new version. See \Stripe\Util\ApiVersion::CURRENT
```
NoelLH, line 424: Think I agree it makes more sense to couple the version to the code, but I'm not sure about just using the library default. I think this will make it easier to upgrade API without noticing / properly testing, and harder to quickly compare the API use to other apps, vs. setting the version explicitly. I think it's also better that we are able to upgrade to a major SDK version, e.g. to receive a security patch, without committing to testing our entire integration against a new API version on the same timeline.

It's also unlikely we will be able to consistently release changes to all apps that 'talk to' Stripe exactly simultaneously. The env var might be slightly more arguable if we think about this – we will still never get the upgrade out at _exactly_ the same time but we could use it to get a bit closer if we wanted. The question then is whether it's more dangerous to run an untested version in isolation or to knowingly use different versions across the stack simultaneously. IMO it probably depends on what breaking changes we've spotted, so this _might_ be an argument for having the option to redeploy the same app code with a different API version if we preempt problems or observe them after release and need a short notice hotfix to just one of the apps.

There's tension between the options because the env var has the same problem as using the library default in making it harder to just search the code to check what we're using. Overall I think my preferred option is to hardcode and not use the library's default or constant, because the benefits of decoupling dependency upgrades from API ones and quickly seeing what is active both seem more likely to be a real world win than 'hot swapping' MatchBot's API version.

composer.json
```diff
@@ -47,7 +47,7 @@
         "ramsey/uuid-doctrine": "^2.0.0",
         "slim/psr7": "^1.3",
         "slim/slim": "^4.7",
-        "stripe/stripe-php": "^9.1.0",
+        "stripe/stripe-php": "^v15.1.0",
```
NoelLH, line 50: Are the version anchoring semantics the same when `v` is included? I'd probably still avoid it unless there's a reason it's needed as I would probably miss this when e.g. searching a codebase for particular version specs.

src/Application/Actions/Donations/Update.php
```diff
@@ -374,7 +375,18 @@ private function addData(
                     /** @var string|null $nextActionType */
                     $nextActionType = null;
                     if ($confirmedIntent->status === PaymentIntent::STATUS_REQUIRES_ACTION) {
-                        $nextActionType = (string) $confirmedIntent->next_action?->type;
+                        /** @var ?ErrorObject $nextAction */
+                        $nextAction = $confirmedIntent->next_action;
+
+                        if ($nextAction && ! $nextAction instanceof ErrorObject) {
```
NoelLH, line 381: How do we know this is an `ErrorObject`?

If it is, it seems like our logic deviates from the docs:
* `ErrorObject::$type` is annotated "One of `api_error`, `card_error`, `idempotency_error`, or `invalid_request_error`."
* Just after this change, we are using the type on the left side of `$nextActionType === 'display_bank_transfer_instructions'`

src/Application/Actions/Hooks/StripePaymentsUpdate.php
```diff
@@ -17,6 +17,7 @@
 use MatchBot\Domain\DonorAccountRepository;
 use MatchBot\Domain\Money;
 use MatchBot\Domain\StripeCustomerId;
+use phpDocumentor\Reflection\Type;
```
NoelLH, line 20:
```suggestion
```

Think this is an accidental import?

app/dependencies.php
bdsl: From reading the docs it seems like letting the library choose the default is the main recommanded way - https://github.com/stripe/stripe-php/wiki/Migration-guide-for-v12 has "Upgrade your integration to be compatible " as the recommended option, and explicitly specifying an older API version as only something they think you'd want to do to "avoid or postpone upgrading API Versions". They don't suggest at any point that we might want to explicitly specify the current API version.

I don't see much issue with using different API versions across the stack, the normal way of things is that each app should use the API version that it was developed & tested against. I guess the only problem comes when concepts from the API leak into the way the apps communicate between each other - possibly that could be an issue, but its not really any different to one app talking to a later version of itself via stuff saved in a database etc.

And we can easily change the one line of app code that sets the version without changing anything else - for me at least that would be quicker and more reliable than changing an env var since its a routine task. I'd be happy to add something like this to make it a tiny bit easier to edit the version in code:

```
                'stripe_version' => \Stripe\Util\ApiVersion::CURRENT, //  default, set explicitly here for clarity.
```

The changelog for each version of the PHP library has a link to the upgrading docs for the API - e.g. https://github.com/stripe/stripe-php/blob/master/CHANGELOG.md#1500---2024-06-24 links to https://docs.stripe.com/upgrades#2024-06-20 . I feel like it's very good to know that we're using a versions of the the PHP library and API version that were designed to work together.

src/Domain/DonationService.php
```diff
@@ -164,6 +164,8 @@ function () use ($donation) {
                 'amount' => $donation->getAmountFractionalIncTip(),
                 'currency' => strtolower($donation->getCurrencyCode()),
                 'description' => $donation->getDescription(),
+                'capture_method' => 'automatic', // 'automatic' was default in previous API versions,
+                                                // default is now 'automatic_async'
```
NoelLH, line 168: Yep, so far seems to make sense to avoid automatic_async. https://docs.stripe.com/payments/payment-intents/asynchronous-capture

It sounds like in return for a modest speed up in a call that is already quite fast, async mode would require us to start doing checks later in the process, e.g. perhaps making sure donors cannot press the final donate button too early if fees haven't yet been calculated. And opening up the possibility of `charge.succeeded` not always having properties we need to check, meaning we have to start thinking about later charge events too, seems like a very significant downside.

NoelLH (commented):
Mostly looks good – added a few things inline besides the standalone discussion on how/whether we specify version

composer.json
bdsl: I think composer ignores the `v` prefix. I hadn't thought about this, happy to remove. Done in af0e4681

app/dependencies.php
NoelLH: Mm – definitely pro's and con's to each way. If I was designing their API my advice would be the same – but I think the differences in our case all derive from the multiple apps working with the same objects. Simply searching our codebases to quickly check if they're all aligned, and if not by how many versions, is really I think the biggest win.

I do still think decoupling security patches from having to do time consuming testing is also worth thinking about. I'm not sure it's realistic to think we will always remember that a patch comes with this implicit API change, especially if e.g. Dependabot proposed a major version security PR to us. I've also not seen anything suggesting that SDKs aren't designed to support older API versions.

src/Application/Actions/Donations/Update.php
bdsl: I wasn't 100% sure that it would be an ErrorObject, which is why I made it log an error but not necessarily stop if its a different type. I guess it probably isn't an ErrorObject then. Do you have an idea what else it would be? If we just remove the @var annotation we get this Psalm error:

```

ERROR: UndefinedMagicPropertyFetch
at /home/barney/projects/matchbot/src/Application/Actions/Donations/Update.php:389:52
Magic instance property Stripe\StripeObject::$type is not defined (see https://psalm.dev/218)
                        $nextActionType = (string) $nextAction?->type;
```

src/Application/Actions/Hooks/StripePaymentsUpdate.php
bdsl: Yes, thanks.

app/dependencies.php
bdsl: As discussed did a combination of hard-coding and relying on library default in fa7077452 by asserting that the two are equal.

src/Application/Actions/Donations/Update.php
bdsl: Replaced with a Psalm suppress in 7b3a04b6383f8

app/dependencies.php
bdsl: Hopefully security patches would generally be minor / patch versions of the PHP SDK and not force us to use a new HTTP API version.

bdsl:
@NoelLH this is ready for re-review. Also similar one at https://github.com/thebiggive/identity/pull/137

NoelLH (approved):

PR #928, opened 2024-07-09
BG2-2620: Allow salesforce to mark a charity as needing update

bdsl:
For now matchbot will just record this mark in the db, and not do anything about it. Next step is doing something about it. Possibly I should have implemented them in the opposite order for more effective CI.

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -67,7 +67,7 @@ return function (App $app) {
     // Requests from Salesforce
 
     $app->post(
-        '/charities/{salesforceId:[a-zA-Z0-9]{18}}/update-required',
+        '/hooks/charities/{salesforceId:[a-zA-Z0-9]{18}}/update-required',
         MarkUpdateRequiredFromSF::class
     )
         ->add(SalesforceAuthMiddleware::class);
--- a/app/routes.php
+++ b/app/routes.php
@@ -68,8 +68,8 @@ return function (App $app) {
 
     $app->post(
         '/charities/{salesforceId:[a-zA-Z0-9]{18}}/update-required',
-        MarkUpdateRequiredFromSF::class
-    )->add(SalesforceAuthMiddleware::class);
+        MarkUpdateRequiredFromSF::class)
+        ->add(SalesforceAuthMiddleware::class);
 
     $app->options('/{routes:.+}', function ($request, $response, $args) {
         return $response;
--- a/app/routes.php
+++ b/app/routes.php
@@ -3,6 +3,7 @@
 declare(strict_types=1);
 
 use Los\RateLimit\RateLimitMiddleware;
+use MatchBot\Application\Actions\Charities\MarkUpdateRequiredFromSF;
 use MatchBot\Application\Actions\DeletePaymentMethod;
 use MatchBot\Application\Actions\UpdatePaymentMethod;
 use MatchBot\Application\Actions\Donations;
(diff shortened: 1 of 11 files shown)


src/Application/Actions/Charities/MarkUpdateRequiredFromSF.php
```diff
@@ -0,0 +1,54 @@
...
+use MatchBot\Domain\CharityRepository;
+use MatchBot\Domain\DomainException\DomainRecordNotFoundException;
+use Psr\Http\Message\ResponseInterface as Response;
+use Psr\Http\Message\ServerRequestInterface as Request;
+use Psr\Log\LoggerInterface;
+
+/**
+ * @psalm-suppress UnusedClass - to be added to routes.php, need to think about authorization requirements.
+ */
+class MarkUpdateRequiredFromSF extends Action
```
bdsl, line 19: this class feels so simple that it might not be worth automated testing. I can make a test if anyone wants though, or this might end up just being a used in a test (and therefore tested incidentally) for the process to actually pull the updates from SF after they get marked as required.

app/routes.php
```diff
@@ -62,6 +64,13 @@ function (RouteCollectorProxy $paymentMethodsGroup) {
     $app->post('/hooks/stripe', Hooks\StripePaymentsUpdate::class);
     $app->post('/hooks/stripe-connect', Hooks\StripePayoutUpdate::class);
 
+    // Requests from Salesforce
+
+    $app->post(
+        '/charities/{salesforceId:[a-zA-Z0-9]{18}}/update-required',
```
NoelLH, line 70:
```suggestion
        '/hooks/charities/{salesforceId:[a-zA-Z0-9]{18}}/update-required',
```

Might be worth having this prefix consistent to help spot everything inbound at a glance?

src/Application/Actions/Charities/MarkUpdateRequiredFromSF.php
```diff
@@ -0,0 +1,54 @@
...
+use MatchBot\Application\Assertion;
+use MatchBot\Domain\CharityRepository;
+use MatchBot\Domain\DomainException\DomainRecordNotFoundException;
+use Psr\Http\Message\ResponseInterface as Response;
+use Psr\Http\Message\ServerRequestInterface as Request;
+use Psr\Log\LoggerInterface;
+
+/**
+ * @psalm-suppress UnusedClass - to be added to routes.php, need to think about authorization requirements.
+ */
```
NoelLH, line 18:
```suggestion
```

NoelLH (commented):
Looks good so far. Minor things inline.

I guess we'll test the action later when it does more?

app/routes.php
bdsl: Not sure if this is really a webhook, I think a defining feature of a webhook is that the address is user-configurable on the sending application side. But it doesn't really matter much and in SF the line between user and developer is pretty blurry. Done in cb28ca10

bdsl:
> I guess we'll test the action later when it does more?

Yes, possibly as part of a larger scale test that also covers the code for pulling the updates from SF.  Or possibly not at all - as it has no conditionals of any sort except for simple error handling I'm not sure it needs any tests.

NoelLH (approved):

bdsl:
> IMO it's still prudent for any Action we are keeping permanently to have at least a cursory 'happy path' unit or integration test. We've definitely had cases of apparently simple ones falling over due to config changes etc. etc., and the cost of maintaining a test for something simple is also low.

OK - I will add some form of test to cover it before moving the ticket to ready for QA.

PR #929, opened 2024-07-10
BG2-2620: Add command for updating charities from SF when required

bdsl:
I think this should complete the work in matchbot for this ticket. Work still required in Salesforce and infrastructure.

### DIFF
--- a/integrationTests/PullCharityUpdatedBasedOnSfHookTest.php
+++ b/integrationTests/PullCharityUpdatedBasedOnSfHookTest.php
@@ -25,10 +25,12 @@ class PullCharityUpdatedBasedOnSfHookTest extends \MatchBot\IntegrationTests\Int
         // arrange
         $em = $this->getService(EntityManager::class);
 
-        $charity = TestCase::someCharity();
+        $campaign = TestCase::someCampaign();
+        $charity = $campaign->getCharity();
         $sfId = $charity->getSalesforceId();
         \assert(is_string($sfId));
 
+        $em->persist($campaign);
         $em->persist($charity);
         $em->flush();
 
@@ -37,6 +39,7 @@ class PullCharityUpdatedBasedOnSfHookTest extends \MatchBot\IntegrationTests\Int
         $fundClientRepository->getForCampaign(Argument::type('string'))->willReturn([]);
 
         $campaignClientProphecy = $this->prophesize(Client\Campaign::class);
+
         $campaignClientProphecy->getById(Argument::any())->willReturn([
             'currencyCode' => 'GBP',
             'endDate' => '2020-01-01',
--- a/integrationTests/PullCharityUpdatedBasedOnSfHookTest.php
+++ b/integrationTests/PullCharityUpdatedBasedOnSfHookTest.php
@@ -4,12 +4,19 @@ namespace MatchBot\IntegrationTests;
 
 use Doctrine\ORM\EntityManager;
 use Laminas\Diactoros\ServerRequest;
+use MatchBot\Application\Commands\UpdateCampaigns;
+use MatchBot\Client;
 use MatchBot\Domain\Charity;
 use MatchBot\Domain\CharityRepository;
 use MatchBot\Domain\Salesforce18Id;
+use MatchBot\Tests\Application\Commands\AlwaysAvailableLockStore;
 use MatchBot\Tests\TestCase;
+use Prophecy\Argument;
+use Psr\Log\NullLogger;
 use Slim\Psr7\Request;
 use Slim\Psr7\Uri;
+use Symfony\Component\Console\Tester\CommandTester;
+use Symfony\Component\Lock\LockFactory;
 
 class PullCharityUpdatedBasedOnSfHookTest extends \MatchBot\IntegrationTests\IntegrationTest
 {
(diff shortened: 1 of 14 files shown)


src/Domain/CampaignRepository.php
```diff
@@ -28,26 +28,26 @@ class CampaignRepository extends SalesforceReadProxyRepository
...
                 charity.tbgClaimingGiftAid = 1 AND
                 charity.tbgApprovedToClaimGiftAid = 0 AND
                 c.endDate >= :extendedLookbackDate
-EOT
             )
-            ->orderBy('c.createdAt', 'ASC')
-            ->setParameter('shortLookBackDate', $shortLookBackDate)
-            ->setParameter('extendedLookbackDate', $extendedLookBackDate);
+            ORDER BY c.createdAt ASC
+            DQL
```
bdsl, line 41: Changed this from query builder to DQL for easier editing, then edited, then thought better and put back to how it was before. I think the change to DQL is worth keeping though.

src/Migrations/Version20240712140753.php
bdsl, line 1: looking back at the previous migrations I see why this is needed - the first migration to add this column wrongly set it it to not null. The second migration made the default null as it should be, so that should be good for any new charities, but didn't affect the values on the existing charities.

src/Application/Commands/UpdateCharities.php
```diff
@@ -0,0 +1,78 @@
...
+
+use Doctrine\Common\Cache\CacheProvider;
+use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
+use GuzzleHttp\Exception\TransferException;
+use MatchBot\Client\NotFoundException;
+use MatchBot\Domain\Campaign;
+use MatchBot\Domain\CampaignRepository;
+use MatchBot\Domain\CharityRepository;
+use MatchBot\Domain\DomainException\DomainCurrencyMustNotChangeException;
```
NoelLH, line 15: Think there's some extraneous imports here – at least this one

src/Application/Commands/UpdateCharities.php
```diff
@@ -0,0 +1,78 @@
...
+        parent::__construct();
+    }
+
+    protected function doExecute(InputInterface $input, OutputInterface $output): int
+    {
+
+        $charitiesNeedingUpdate = $this->charityRepository->findAllNeedingUpdateFromSf();
+
+        foreach ($charitiesNeedingUpdate as $charity) {
+            $campaign = $this->campaignRepository->findOneBy(['charity' => $charity->getId()]);
```
NoelLH, line 54: Probably worth explicitly choosing which one? If we are going to stick with one probably the latest – and perhaps it is worth doing multiple up to e.g. 18 months so we also have a consistent set of expectations for how far back campaign data itself might be patched.

NoelLH (commented):
LGTM so far – a few things inline

integrationTests/PullCharityUpdatedBasedOnSfHookTest.php
```diff
@@ -0,0 +1,117 @@
...
+                'New Charity Name',
+                $charity->getStripeAccountId() ?? throw new \Exception('Missing Stripe ID')
+            )
+        );
+
+        $this->getContainer()->set(Client\Campaign::class, $campaignClientProphecy->reveal());
+
+        // act
+        $this->simulateRequestFromSFTo('/hooks/charities/' . $sfId . '/update-required');
+        $this->invokeCommand(UpdateCharities::class);
```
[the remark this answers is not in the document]
bdsl: I'm not keen to do it synchronously as that feels quite inefficient - would mean blocking up one SF thread while it waits for matchbot to callback and wait for another SF thread. Maybe not a problem with SF but I remember that style can easily double or worse costs if done with lambdas.

Also would there not be a human waiting for a response? If we edit an account whether through portal or back office don't we have to wait for the trigger to return to get a confirmation that saving it is complete?

Via SQS sounds ideal, I just didn't want to get into that after all the trouble we had with https://thebiggive.atlassian.net/browse/MAT-363 . But could change to use SQS if/when MAT-363 is done or near done and gives us an example to follow.

src/Application/Commands/UpdateCharities.php
bdsl: Thanks, done in b3630d45 . Doing multiple, going back 18 months as you suggested. Also doing a database join to get rid of the N+1 query problem. Although there may still be one extra query for each charity triggered by calling get functions on the charity ORM proxy object, but performance isn't critical here.

src/Application/Commands/UpdateCharities.php
bdsl: Thanks, fixed in 7e92778e .

integrationTests/PullCharityUpdatedBasedOnSfHookTest.php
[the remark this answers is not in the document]
NoelLH: Callouts from SF happen in one of the kinds of async thread generally – I think maybe always. Certainly we can choose not to make BG team wait for the API reply.

NoelLH (approved):

PR #932, opened 2024-07-16
Mat 368 funding allocation data patch ii

bdsl:


src/Migrations/Version20240716142043.php
```diff
@@ -0,0 +1,57 @@
...
+            'a056900002SEVhXAAX' => ['a0969000021yW3AAAU', 'Postcode Green Trust'],
+            'a056900002SEVkqAAH' => ['a0969000021yW3AAAU', 'Postcode Green Trust'],
+
+            'a056900002SEVoiAAH' => ['a0969000021yW3AAAU', 'Postcode Green Trust'],
+            'a056900002SEVpMAAX' => ['a0969000021yW3AAAU', 'Postcode Green Trust'],
+            'a056900002SEVvtAAH' => ['a0969000021yW3AAAU', 'Postcode Green Trust'],
+
+            'a056900002SEW18AAH' => ['a0969000021y3rLAAQ', 'The Reed Foundation'],
+            'a056900002SEW8YAAX' => ['a0969000021yW3AAAU', 'Postcode Green Trust'],
+            'a056900002SMAFHAA5' => ['a0969000021yW3AAAU', 'Postcode Green Trust'],
```
bdsl, line 30: pasting in the above IDs from Jira is potentially error-prone, as was entering them into Jira in the first place, so worth double checking to catch errors in either step.

PR #936, opened 2024-07-17
MAT-363 – replace Salesforce pushing with combined UUID endpoint approach

NoelLH:


src/Domain/Donation.php
```diff
@@ -107,7 +107,7 @@ class Donation extends SalesforceWriteProxy
      * @see Donation::$currencyCode
      */
     #[ORM\Column(type: 'decimal', precision: 18, scale: 2)]
-    protected readonly string $amount;
+    protected string $amount; // todo remember how we made it readonly without breaking tests
```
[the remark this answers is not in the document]
NoelLH: Think I thought of a more sensible option in a couple of ways that sidesteps this problem: a98d19bdd5afa710b2333e3371ceb6eb9a31058d

src/Domain/Donation.php
```diff
@@ -442,35 +442,11 @@ public function replaceNullPaymentMethodTypeWithCard(): void
...
-            $data['matchedAmount'],
-            $data['cardBrand'],
-            $data['cardCountry'],
-        );
-
-        return $data;
-    }
-
-    public function toApiModel(): array
+    public function toApiModel(bool $forSalesforce = false): array
```
bdsl, line 445: I'm not keen on the flag argument, I think it'd be clearer to have two functions, e.g. `toFEApiModel` and `toSFApiModel`. The latter is easy to implement on top of the former.  Easier to see `toSFApiModel` is used that way.

bdsl: If you really want a flag arg then maybe make it an enum instead of a bool, and at least then it'd be easy to find usages of the `ApiConsumer::salesforce` case.

src/Client/Donation.php
```diff
@@ -123,23 +70,30 @@ public function put(DonationModel $donation): bool
...
                 $ex->getResponse() ? $ex->getResponse()->getBody() : 'N/A',
             ));
 
-            return false;
+            throw new BadRequestException('Donation not created');
         }
 
-        return ($response->getStatusCode() === 200);
-    }
+        if ($response->getStatusCode() > 201 || $response->getStatusCode() < 200) {
```
bdsl, line 84: this might be clearer?

```suggestion
        if (! in_array($response->getStatusCode(), [200, 201], true)) {

```

src/Client/Donation.php
```diff
@@ -123,23 +70,30 @@ public function put(DonationModel $donation): bool
...
+            throw new BadRequestException('Donation not upserted');
+        }
 
-    private function hash(string $body): string
-    {
-        return hash_hmac('sha256', trim($body), getenv('WEBHOOK_DONATION_SECRET'));
+        /**
+         * @var array{'salesforceId': string} $donationCreatedResponse
+         */
+        $donationCreatedResponse = json_decode((string) $response->getBody(), true);
```
bdsl, line 92: casting removes type safety, since you can cast anything (including null) to anything. Suppresses both static and runtime errors. Instead of casting to string can we do this?

```suggestion
        $donationCreatedResponse = json_decode($response->getBody()->getContents(), true);
```

src/Domain/Donation.php
```diff
@@ -442,35 +442,11 @@ public function replaceNullPaymentMethodTypeWithCard(): void
...
     {
         $data = [
+            'amountMatchedByChampionFunds' => (float) $this->getConfirmedChampionWithdrawalTotal(),
```
bdsl, line 448: Seems a shame to convert to float when we've been so careful to always treat as strings up to now, and it will be converted to Decimal in salesforce. But should be fine and not much we can do. Made me find this PHP issue - would be good if it was fixed in PHP https://bugs.php.net/bug.php?id=80484

src/Domain/DonationRepository.php
```diff
@@ -662,4 +666,48 @@ public function removeAllFundingWithdrawalsForDonation(Donation $donation): void
             }
         });
     }
+
+    private function setSalesforceIdIfNeeded(Donation $donation, string $salesforceId): void
```
bdsl, line 670: Worth using the Salesforce18Id VO that we have here and in the related functions? 

```suggestion
    private function setSalesforceIdIfNeeded(Donation $donation, string $salesforceId): void
```

src/Domain/Donation.php
bdsl: But I think the non-flag way was more or less what we had before, so would be good to understand the reasoning for the change.

src/Domain/DonationRepository.php
```diff
@@ -662,4 +666,48 @@ public function removeAllFundingWithdrawalsForDonation(Donation $donation): void
...
+    {
+        if ($donation->getSalesforceId() !== null) {
```
bdsl, line 672: throw if the donation's existing sf ID is not null and different to the one we're trying to set?

src/Domain/DonationRepository.php
bdsl: Done in pairing

bdsl (dismissed):

bdsl (dismissed):

integrationTests/DonationRepositoryTest.php
```diff
@@ -247,8 +255,8 @@ public function donationAgesDataProvider(): array
     {
         // age in seconds, should be pushed to sf
         return [
-            [0, false],
-            [299, false],
+//            [0, false],
+//            [299, false],
```
bdsl, line 259:
```suggestion
            [0, false],
            [299, false],
```

PR #937, opened 2024-07-17
MAT-368 – patch fund data by amending just CampaignFunding.fund_id

NoelLH:


src/Migrations/Version20240717173149.php
```diff
@@ -0,0 +1,52 @@
...
+            $this->addSql(<<<SQL
+                UPDATE CampaignFunding
+                INNER JOIN `Campaign_CampaignFunding` ON `Campaign_CampaignFunding`.`campaignfunding_id` = `CampaignFunding`.`id`
+                INNER JOIN `Campaign` ON `Campaign`.`id` = `Campaign_CampaignFunding`.`campaign_id`
+                SET CampaignFunding.fund_id = $newFundNumericId
+                WHERE Campaign.salesforceId = '$campaignSFID'
+            SQL
+            );
+        }
+    }
```
bdsl, line 46: just to make sure I confirmed that doctrine does wrap all the SQL queries in a transaction. It does by default, if we ever want to opt out of that we have to add this to a migration file:

```
    public function isTransactional(): false
    {
        return false;
    }
```

bdsl (approved):

PR #941, opened 2024-07-22
DON-983: Allow listing all donation for logged in donor

bdsl:


dorota-joanna (approved):
LGTM and integration tests passing

PR #943, opened 2024-07-24
MAT-363 – DB persist stability improvements; review fixes; test improvements

NoelLH:
* Addressed what I could from https://github.com/thebiggive/matchbot/pull/922 review
* Combined needlessly two step Salesforce field updates and simplified surrounding logic further
* Introduced a retry for saving the Salesforce fields and increased log severity level if it fails
* Moved some tests that asserted things about `push()` calls to the equivalent bus dispatches

### DIFF
--- a/integrationTests/DonationRepositoryTest.php
+++ b/integrationTests/DonationRepositoryTest.php
@@ -217,7 +217,6 @@ class DonationRepositoryTest extends IntegrationTest
         $donationClientProphecy = $this->prophesize(\MatchBot\Client\Donation::class);
 
         $busProphecy = $this->prophesize(RoutableMessageBus::class);
-        $dummyEnvelope = new Envelope(new \stdClass()); // Final so can't prophesise.
 
         $pendingCreate = \MatchBot\Domain\SalesforceWriteProxy::PUSH_STATUS_PENDING_CREATE;
         $connection->executeStatement(<<<SQL
@@ -234,7 +233,7 @@ class DonationRepositoryTest extends IntegrationTest
 
         // assert
         $busDispatchMethod = $busProphecy->dispatch(Argument::type(Envelope::class))
-            ->willReturn($dummyEnvelope);
+            ->willReturn(new Envelope(new \stdClass()));
         if ($shouldPush) {
             $busDispatchMethod->shouldBeCalledOnce();
         } else {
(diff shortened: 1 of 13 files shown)


src/Domain/Donation.php
```diff
@@ -1415,7 +1415,9 @@ public function update(
         $this->setTipGiftAid($tipGiftAid);
         $this->setTbgShouldProcessGiftAid($this->getCampaign()->getCharity()->isTbgClaimingGiftAid());
         $this->setDonorHomePostcode($donorHomePostcode);
-        $this->setDonorName($donorName);
+        if ($donorName) {
```
NoelLH, line 1418: One of the tests in `UpdateHandlesLockExceptionTest` would bail out without a message being dispatched, due to the lack of a first / any donor name, without this condition being included. I think there's probably other test changes that could have addressed the problem, but it seems prudent not to let one version of the property wipe out the other version and potentially already-persisted data in any case.

integrationTests/DonationRepositoryTest.php
```diff
@@ -234,7 +234,15 @@ public function testPushSalesForcePendingPushesNewDonationButNotVeryNewDonation(
 
         // assert
         $busDispatchMethod = $busProphecy->dispatch(Argument::type(Envelope::class))
-            ->willReturn($dummyEnvelope);
+            ->will(function (array $args) use ($donationUUID) {
+                /** @var Envelope $envelope */
+                $envelope = $args[0];
+                $message = $envelope->getMessage();
+                \assert($message instanceof DonationUpserted);
+                \assert($message->uuid === $donationUUID);
```
bdsl, line 242: All very minor points but:
- probably better to put the type on the param and let the local be inferred from that instead of typing the local:
- and for assertions that are the point of the test (rather than assertions that are just needed incidentally) a bit better to use the PHPUnit assertions instead of built in `assert`. I realise we can't use $this in these closures because its rebound, but these are static functions anyway so we don't need to.

```suggestion
            ->will(/**
             * @param array{0: Envelope} $args
             */ function (array $args) use ($donationUUID) {
                $envelope = $args[0];
                $message = $envelope->getMessage();
                TestCase::assertInstanceOf(DonationUpserted::class, $message);
                TestCase::assertSame($donationUUID, $message->uuid);
```

bdsl (dismissed):
There's one non-blocking suggestion inline, all LGTM.

PR #946, opened 2024-07-30
Fix Psalm issues from baseline

bdsl:
Just to make the code marginly easier to work with. Reduces count of Psalm errors from 537 to 467, a 13% reduction.

### DIFF
--- a/src/Application/Matching/Adapter.php
+++ b/src/Application/Matching/Adapter.php
@@ -71,7 +71,7 @@ class Adapter
          * @psalm-suppress PossiblyInvalidArrayAccess
          * @psalm-suppress PossiblyFalseReference - we know incrBy will retrun an array in multi mode
          */
-        [$_, $fundBalanceFractional] = $this->storage->multi()
+        [$_initResponse, $fundBalanceFractional] = $this->storage->multi()
             // Init if and only if new to Redis or expired (after 24 hours), using database value.
             ->set(
                 $this->buildKey($funding),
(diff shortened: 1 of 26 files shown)


src/Application/Matching/Adapter.php
```diff
@@ -71,7 +71,7 @@ private function addAmountWithoutSavingFundingsToDB(CampaignFunding $funding, st
          * @psalm-suppress PossiblyInvalidArrayAccess
          * @psalm-suppress PossiblyFalseReference - we know incrBy will retrun an array in multi mode
          */
-        [$initResponse, $fundBalanceFractional] = $this->storage->multi()
+        [$_, $fundBalanceFractional] = $this->storage->multi()
```
NoelLH, line 74: Possibly more helpful for future editors to use underscore but with a more descriptive name still?

NoelLH (dismissed):
Non-blocking comment inline re. unused underscore vars (which I think applies in 2 places in the same file). Otherwise all LGTM

src/Application/Matching/Adapter.php
bdsl:
```suggestion
        [$_initResponse, $fundBalanceFractional] = $this->storage->multi()
```

src/Application/Matching/Adapter.php
```diff
@@ -101,8 +101,9 @@ public function subtractAmountWithoutSavingToDB(CampaignFunding $funding, string
 
         /**
          * @psalm-suppress PossiblyFalseReference - in mulit mode decrBy will not return false.
+         * @psalm-suppress PossiblyInvalidArrayAccess - in this case we know exec returns array
          */
-        [$initResponse, $fundBalanceFractional] = $this->storage->multi()
+        [$_, $fundBalanceFractional] = $this->storage->multi()
```
bdsl, line 106:
```suggestion
        [$_initResponse, $fundBalanceFractional] = $this->storage->multi()
```

PR #948, opened 2024-08-06
DON-1025: Add stubbed out RegularGiving/GetAllForUser route

bdsl:
The main thing here is the start of working out what a regular giving mandate looks like as a JSON model to be served to front end for the donor to view. I'm sure we'll refine it as we go but worth spending time working out what we think we'll need now. Probably easiest to view and discuss it in the test file `ListRegularGivingMandatesTest.php`.

For now the data is all hard coded dummy data. Once we've decided what we want it to look like I'll go ahead and implement the mandate as an entity to load with Doctrine.

### DIFF
--- a/integrationTests/ListRegularGivingMandatesTest.php
+++ b/integrationTests/ListRegularGivingMandatesTest.php
@@ -15,37 +15,58 @@ use Ramsey\Uuid\Uuid;
  */
 class ListRegularGivingMandatesTest extends IntegrationTest
 {
-    public function setUp(): void
-    {
-        parent::setUp();
-    }
-
     /**
      * @psalm-suppress MixedArrayAccess - hard to avoid in integration testing like this.
      */
-    public function testItReturnsAListWithCollectedDonation(): void
+    public function testItListsRegularGivingMandate(): void
     {
-//        $this->db()->executeStatement(
-//            "",
-//            []
-//        );
-
-        $this->getService(EntityManagerInterface::class)->clear();
-
         $personId = PersonId::of(Uuid::uuid4()->toString());
 
-        // act
         $allMandatesBody = (string) $this->requestFromController($personId)->getBody();
 
-        // assert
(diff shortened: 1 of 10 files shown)


integrationTests/ListRegularGivingMandatesTest.php
```diff
@@ -0,0 +1,87 @@
...
+                    'type' => 'monthly',
+                    'dayOfMonth' => 31, // i.e. donation to be taken on last day of each calendar month
+                    'activeFrom' => '2024-08-06T00:00:00+00:00',
+                ],
+                // I'm not sure if we'll details of the donor like their name and home address
+                // recorded as part of the mandate. I think probably not - we have that on the person
+                // record and on each individual donation.
+                'charityName' => 'Some Charity',
+                'createdTime' => '2024-08-06T00:00:00+00:00',
+                'giftAid' => true, // not 100% sure yet if GA will be an option on regular giving, assuming it will.
```
bdsl, line 56:
```suggestion
                'giftAid' => true,
```

bdsl:
Result of running the auto generated migration. I'm not sure why it doesn't seem to include the unique constraint for combination of donor id and campaign ID that I included in the entity class.

image

integrationTests/ListRegularGivingMandatesTest.php
```diff
@@ -0,0 +1,79 @@
...
+                'charityName' => 'Some Charity',
+                'giftAid' => true,
+                'status' => 'active',
+                'tipAmount' => [
+                    // todo before calling ticket done: confirm if we are taking tips like this for regular giving
+                    'amountInPence' => 100,
+                    'currency' => 'GBP',
+                ],
+                'createdTime' => '2024-08-06T00:00:00+00:00',
+                'updatedTime' => '2024-08-06T00:00:00+00:00',
```
bdsl, line 58: we also want a field for payment method as noted on Jira, but I'll probably leave that for a future PR now.

dorota-joanna (approved):
LGTM

PR #953, opened 2024-08-07
BG2-2620: Don't generate alarm on attempt to update charity with expi…

bdsl:
…red campaigns

It's probably better to allow the campaigns to be updated to reflect the expired status, but we don't have to do that today.

### DIFF
--- a/src/Application/Commands/UpdateCampaigns.php
+++ b/src/Application/Commands/UpdateCampaigns.php
@@ -61,7 +61,7 @@ EOT
 
             try {
                 $this->pull($campaign, $output);
-            } catch (NotFoundException $exception) {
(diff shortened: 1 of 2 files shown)


src/Application/Commands/UpdateCampaigns.php
```diff
@@ -61,7 +61,7 @@ protected function doExecute(InputInterface $input, OutputInterface $output): in
 
             try {
                 $this->pull($campaign, $output);
-            } catch (NotFoundException $exception) {
+            } catch (NotFoundException|CampaignNotReady $exception) {
```
bdsl, line 64: should have been done earlier today when I introduced the CampaignNotReady exception class

dorota-joanna (approved):
LGTM

PR #960, opened 2024-08-12
DON-1025: Only show active regular giving mandates in list for donor

bdsl:


dorota-joanna (approved):
LGTM and mandate test passes

PR #963, opened 2024-08-14
Better document what `Fund` and `CampaignFunding` models do

NoelLH:


src/Domain/CampaignFunding.php
```diff
@@ -48,6 +58,10 @@ class CampaignFunding extends Model
      * The amount of this funding allocation not already claimed. If you plan to allocate funds, always read this
      * with a PESSIMISTIC_WRITE lock and modify it in the same transaction you create a FundingWithdrawal.
      *
+     * This is a less-realtime, eventually consistent copy of the information that we store in Redis
+     * as the primary source of truth, for better performance and to avoid locking issues around fund
+     * allocation.
```
bdsl, line 63: Think this was slightly ambiguous about which is the primary source of truth. 
```suggestion
     * This is a less-realtime, eventually consistent copy of the information that we store in Redis. 
     * Redis is the primary source of truth, for better performance and to avoid locking issues around fund allocation.
```

bdsl (dismissed):
Thanks! Added one small suggestion. LGTM

PR #965, opened 2024-08-14
DON-1025 / DON-998: Restrict the range of regular giving payment amounts

bdsl:


src/Domain/Money.php
```diff
@@ -63,4 +63,27 @@ public function jsonSerialize(): mixed
...
+        if ($this->currency !== $that->currency) {
+            throw new \UnexpectedValueException("Cannot compare amounts with different currencies");
+        }
+
+        return $this->amountInPence > $that->amountInPence;
+    }
+
+    public function __toString()
+    {
+        return $this->amountInPence / 100 . $this->currency->isoCode();
```
NoelLH, line 87:
```suggestion
        return $this->currency->isoCode() . ' ' . ($this->amountInPence / 100);
```
I think this is more typical way to stringify when using the ISO codes?

Would need corresponding test tweaks.

NoelLH (approved):
One suggestion inline, not a blocker. LGTM

src/Domain/Money.php
bdsl: Yes I think you're right. I would avoid __toString for generating any user facing output, but will change to make any logs its used in a bit easier to read, then merge.

PR #970, opened 2024-08-19
Don-993: Record total paid by donor on donation record

bdsl:
At present the calculation of the total paid by donor (core amount + any tip + any fee cover) is done several times and is expressed in four separate lines of code - one here and three in frontend. This PR makes matchbot record the result as it comes back from Stripe so that it can replace two of the lines in frontend. We may or may not also want to replace the final one, which is for displaying the expected amount *before* the payment is taken.

See https://github.com/thebiggive/donate-frontend/pull/1617#discussion_r1720106522


NoelLH (approved):
Pushed tiny doc block change we paired on. All LGTM!

PR #972, opened 2024-08-21
DON-983: Remove support for feeCover / feePercentage

bdsl:
These were introduced to the code in 2021 but have never had any value other than zero in production, and are not likely to be used in the near future.

We discussed removing feeCover. As it's the presence of feePercentage that would have enabled feeCover as an option in frontend I am assuming we want to remove that as well.


src/Migrations/Version20240821103038.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
bdsl, line 1: probably this can actually lead to downtime since the db migration will be run before the new code is used to serve web requests. So ideally we could remove this from the PR and do it later, but I think matchbot deploys are quick enough that we don't really need to worry as long as we're not deploying to prod at a busy time.

dorota-joanna (approved):
LGTM

src/Migrations/Version20240821103038.php
NoelLH: Yeah, this is probably around a couple of minutes for web tasks, which roll out gradually in a blue-green deployment live. And only that short if the old task starts failing health checks and be taken out of service if the new one is passing. I think out of sync proxies will make that happen due to `/ping` 500-ing but not 100% certain.

It might be that consumer tasks which don't have a web endpoint and comparable health check may need a force redeploy, to avoid them potentially being broken for their whole normal lifetime.

PR #976, opened 2024-08-23
BG2-2722: Remove DI-workaround

bdsl:
Testing to see what happens if we remove this


tests/TestCase.php
```diff
@@ -8,6 +8,7 @@
 use Exception;
 use MatchBot\Application\HttpModels\DonationCreate;
 use MatchBot\Application\Messenger\DonationUpserted;
+use MatchBot\Application\Messenger\Handler\CharityUpdatedHandler;
```
NoelLH, line 11: I don't think any of the new imports here are needed any more?

NoelLH (approved):
Extraneous imports notwithstanding all LGTM

PR #977, opened 2024-08-23
Don-1016: Allow recording pre-authorization date on donations and create command to collect pre-authorized donations

bdsl:
The new command class is only barely there but safe to merge as nothing uses it yet. The main work here is moving code from \MatchBot\Application\Actions\Donations\Confirm to \MatchBot\Domain\DonationService::confirm to make it re-usable, and also adding the ability to record on a Donation record that it's preauthorized to be collected on a future date.


src/Application/Actions/Donations/Confirm.php
```diff
@@ -116,17 +118,16 @@ protected function action(Request $request, Response $response, array $args): Re
             throw new HttpBadRequestException($request, $message);
         }
 
-        $paymentIntentId = $donation->getTransactionId();
         try {
-            $paymentMethod = $this->stripe->updatePaymentMethodBillingDetail($paymentMethodId, $donation);
```
bdsl, line 121: this is a slight performance degradation - not using the paymentMethod returned from `updatePaymentMethodBillingDetail` means we have to retrieve it from Stripe later.  But applying command query separation simplifies moving code around.

src/Domain/Donation.php
```diff
@@ -306,6 +307,17 @@ class Donation extends SalesforceWriteProxy
...
+    /**
+     * We only have permission to collect a preAuthorized donation on or after the given date. Intented to be used
+     * with regular giving.
+     *
+     * @psalm-suppress UnusedProperty (will use soon)
+     *
+     * @see DonationStatus::PreAuthorized
+     */
+    #[ORM\Column(nullable: true, type: 'LocalDate')]
+    private ?LocalDate $preAuthorizationDate = null;
```
bdsl, line 319: we had some discussion yesterday about whether this is better as a LocalDate (type from [Brick\DateTime library](https://github.com/brick/date-time), "an isolated date such as 2014-08-31") or a \DateTimeImmutable as we've used before.

I sort of like the LocalDate because it's just as much information as we need and no more, and what we store can exactly match what we display to the donor, e.g. "we will collect your payment on 2014-08-31". A DateTimeImmutable also has a time of day, which I don't think the donor will be interested in, and a timezone, which they also won't be interested in and which wouldn't be stored in the database.

It is part of a wider question of whether we might want to adopt the library, and use more of its classes. I also use the Instant class in this PR, which "Represents a point in time, with a nanosecond precision."

NoelLH: @bdsl I'm not sure it's relevant whether a donor cares about the time, more whether _we_ benefit from clearly including the time at which collection becomes appropriate in the data. We can summarise it to donors however we choose either way.

Similarly to our discussion about dates in Salesforce, IMO it only _looks like_ the time is redundant because comparing to midnight is implicit. But that's ambiguous: we are introducing potential misunderstanding about whether we mean midnight UTC or midnight UK. By using a datetime from the start we:
* remove that ambiguity, and
* avoid introducing a new approach to datetimes vs. the rest of the app

I'm not against using the library where it has other value to us, but I don't see the benefit of excluding time from the data here.

src/Domain/DonationService.php
```diff
@@ -323,4 +327,67 @@ private function runWithPossibleRetry(\Closure $retryable, string $actionName):
...
+    /**
+     * Finalized a donation, instructing stripe to attempt to take payment.
+     */
+    public function confirm(
+        Donation $donation,
+        string $paymentMethodId
+    ): \Stripe\PaymentIntent {
+        $this->updateDonationFees($paymentMethodId, $donation);
+
+        // looks like sometimes $paymentIntentId and $paymentMethodId are for different customers.
```
NoelLH, line 340: I know this comment is migrated from a Confirm Action edit a few months ago, but I'm struggling to understand exactly why it would be the case or how it relates to the next line. Do we still need it, and if so can we clarify?

NoelLH (commented):
Not sure about the Date type only per discussion outside this review.

Everything else LGTM, very minor Q on a code comment mystery inline.

src/Domain/DonationService.php
bdsl: good question, I'm not sure. Let me try and find some context from when it was originally written.

src/Domain/Donation.php
bdsl: OK. Firstly I think this probably doesn't matter a huge amount to be honest - the logic is not complicated so we can work it either way, and even if we do mess up and collect money early it's most likely going to just mean collecting one hour early at 11pm when we should have collected at midnight. We might choose not to run the processing at night time anyway. So @NoelLH if you want to make a call on what to use I can go with your decision for now.

But I don't really think storing date only is ambiguous - or if it is we could just fix that by documenting that it refers to London timezone. If there is any real ambiguity then it seems like we would need to display the TZ so that we have a shared understanding of the agreement with the donor.

And when doctrine stores a datetime in the db it doesn't store the timezone info anyway (see https://www.doctrine-project.org/projects/doctrine-dbal/en/4.1/reference/types.html#datetime ) . So persisting that to the db and then pulling it out is throwing away information.

It's a very edge case but in principle if the UK TZ rules changed between the point the donation was prepared and the collection date we'd want to collect according to the new rules, not the one in place previously, which to me means we want to save something based on the UK timezone not the UTC timezone. 

I guess we could also remove all ambiguity by storing a `LocalDate` alongside a `TimeZoneRegion` - but it seems unnecessary to store something that will never vary. (although I did store currency in the mandate table, I think more because of influence of it being variable in previous models than because of any actual prospect of it being variable) (the LocalDate with TimeZoneRegion could also be a `ZonedDateTime`, but [that isn't something MySQL supports saving](https://github.com/brick/date-time-doctrine/issues/1))

bdsl: I read this a while ago and thought it was good. LocalDate is like a partial version Skeet's Option "*Option 3: preserve local time, using UTC as derived data to be recomputed*" here:

[Storing UTC is not a silver bullet - Jon Skeet's coding blog](https://codeblog.jonskeet.uk/2019/03/27/storing-utc-is-not-a-silver-bullet/)

src/Domain/DonationService.php
bdsl: So this is from issue MAT-339 which we never quite got to the bottom of. Seems to have been some isolated cases and we don't know if they were related to genuine donation attempts. Shall we just delete the comment?

NoelLH: Yeah I think so. We're not knowingly _supporting_ cases where different Customers own the two objects, right?

bdsl, line 340: Delete comment related to MAT-339 
```suggestion
```

bdsl: We're not. I've deleted it.

src/Domain/Donation.php
NoelLH: It's interesting that Doctrine DBAL frames it as not storing TZ information. I have always seen its relationship with MySQL mapping as being a UTC representation, as that is, I believe, what happens automatically when you use its ORM.

For sure, it's possible to make it unambiguous in docs – I just think it is less likely to go wrong if it is intuitive and unambiguous in code. This includes choosing some time in the early morning rather than midnight, which I agree we may well want to do. There are advantages both ways round (fixing the time sooner vs. later) but overall IMO it's cleaner to set the time as the point of creation and then not have to think about it later in the process.

I don't think either approach requires 'conversion' to print a date for the donor really – and both require simple formatting. We want to show them something nicer than _2020-01-01_ regardless of how we stored it.

Overall I would still vote for a datetime immutable I think – I'm not really seeing a gain from moving the complexity to time reading, which we'll do more of.

bdsl: I think it's a conversion because if we're talking about e.g. '2024-08-24' then with the DateTimeImmutable approach we we would be storing it in UTC as '2024-08-23T23:00:00', so would need a conversion to make the 23 into a 24 for display. 

I'm not sure voting makes sense here though - I can change it to DateTimeImutable. Will have to rethink the signature of the function I just started writing since we want to display the output on the regular giving page, currently

```
public function firstPaymentDayAfter(LocalDateTime $currentDateTime): LocalDate
```

bdsl: I think the same considerations about which approach to use may apply to other places we could use the library, so I removed it in 809d44f616cb5c45b69db56c3e8ccbe79d19f73 - although I do think it makes the code to implement `firstPaymentDayAfter` a quite a bit less intuitive. 

There is also the option of using a library like `cakephp/chronos` that is much closer to the native PHP system and also provides convenient methods and properties , e.g. `->day` instead of `->format('d')`, although disappointingly not any wither methods like `->withDay(4)`.

NoelLH (approved):
All LGTM

Would just suggest we re-title the PR to clarify the scope of what's changed, especially to help when scanning the list from Jira

bdsl:
@NoelLH Thanks, I'll fix the merge conflict after the meeting and merge.

PR #980, opened 2024-08-23
Restore DB columns that have been dropped from staging but not yet dr…

bdsl:
…opped from prod

Starting to regret dropping these in advance of deploying the code that doens't need them to prod. Wanting to follow the more conservative path of first deploying the code that doesn't require these columns and only after that dropping the columns.

When we're ready to drop the columns we can do another migration a try-catch to drop them just in environments where they're not dropped yet.

Editing old migrations before they hit prod rather than just adding a new one to run after them because I want to avoid dropping the columns in prod for now, not drop them and then restore them.

### DIFF
--- a/integrationTests/DonationPersistenceTest.php
+++ b/integrationTests/DonationPersistenceTest.php
@@ -87,6 +87,7 @@ class DonationPersistenceTest extends IntegrationTest
             'charityFeeVat' => '0.00',
             'originalPspFee' => '0.00',
             'currencyCode' => 'GBP',
+            'feeCoverAmount' => '0.00',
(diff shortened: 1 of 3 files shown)


bdsl:
We run doctrine diff in CI so it shows us that the DB schema is now out of sync with the entity class structure at https://app.circleci.com/pipelines/github/thebiggive/matchbot/3531/workflows/63559a0a-c95d-4277-b3f2-1c66b0b41555/jobs/10079 but it does not fail the build (as I thought it might).

NoelLH (approved):

dorota-joanna (commented):
LGTM

PR #981, opened 2024-08-27
DON-1025: Expose mandate next payment date to frontend

bdsl:
Diff in test shows the change in output here:

```diff
              "schedule": {
                    "type": "monthly",
                    "dayOfMonth": 12,
                    "activeFrom": null
+                   "expectedNextPaymentDate": "2024-09-12"
              },
```

### DIFF
--- a/src/Domain/RegularGivingService.php
+++ b/src/Domain/RegularGivingService.php
@@ -2,21 +2,25 @@
 
 namespace MatchBot\Domain;
 
-class RegularGivingService
+readonly class RegularGivingService
 {
     /** @psalm-suppress PossiblyUnusedMethod - used by DI container  */
-    public function __construct(private RegularGivingMandateRepository $regularGivingMandateRepository)
-    {
+    public function __construct(
+        private RegularGivingMandateRepository $regularGivingMandateRepository,
+        private \DateTimeImmutable $now
+    ) {
     }
 
     public function allActiveForDonorAsApiModel(PersonId $donor): array
     {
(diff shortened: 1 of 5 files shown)


src/Domain/RegularGivingService.php
```diff
@@ -2,21 +2,25 @@
...
+        private RegularGivingMandateRepository $regularGivingMandateRepository,
+        private \DateTimeImmutable $now
+    ) {
     }
 
     public function allActiveForDonorAsApiModel(PersonId $donor): array
     {
         $mandatesWithCharities = $this->regularGivingMandateRepository->allActiveForDonorWithCharities($donor);
 
+        $currentUKTime = $this->now->setTimezone(new \DateTimeZone("Europe/London"));
```
bdsl, line 18: not sure if this is the best place to do the conversion to UK time.

NoelLH: I think while it only happens once + with no variation this is good. I guess until we have charities in other countries, there's not enough value to consider varying times of collection & display TZ by donor – they will likely assume we're using UK ones.

NoelLH (approved):
LGTM

src/Domain/RegularGivingService.php
bdsl: Yep. I think even if we have charities in other countries as long as the site is clearly UK based for something like this it'll be fine to have everyone use UK. For something monthly I don't expect many people will mind if it's off up to 12ish hours one way or the other.

PR #984, opened 2024-08-27
MAT-376: Start collecting pre-authorized donations

bdsl:


NoelLH (approved):

bdsl:
Closing to favour https://github.com/thebiggive/matchbot/pull/985 as it seems like pulling the payment method from Stripe won't really work.

PR #986, opened 2024-08-29
MAT-376: Create command to generate a regular giving mandate for testing

bdsl:
Docblock from new command:

```php
/**
 * Use to create a regular giving agreement for test purposes. Usage example:
 *
 * To create a regular giving mandate for the emergency campaign, with gift aid enabled, for
 * a donor with the given identity server UUID, stripe customer ID and payment method ID, to donate £500 each month
 * on the 20th day of the month, starting this month if today is <19th otherwise starting next month run:
 *
 * ./matchbot matchbot:setup-test-mandate --campaign Emergency --gift-aid
 *      --donor-uuid 72f1e368-65f3-11ef-878f-fb0a039f0650 --donor-stripeid cus_xyxzaxasfdef
*       --donor-pmid pm_1xyzxzyxzyxzyxzyxzyxzyxz 500 20
 *
 * This doesn't create the intitial donation, but it does pre-authorise the 2nd and 3rd donations.
 */
 ```

### DIFF
--- a/matchbot-cli.php
+++ b/matchbot-cli.php
@@ -17,6 +17,7 @@ use MatchBot\Application\Commands\RedistributeMatchFunds;
 use MatchBot\Application\Commands\ResetMatching;
 use MatchBot\Application\Commands\RetrospectivelyMatch;
 use MatchBot\Application\Commands\ScheduledOutOfSyncFundsCheck;
+use MatchBot\Application\Commands\SetupTestMandate;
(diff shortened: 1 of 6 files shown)


bdsl:
@NoelLH what sort of test were you thinking? As `SetupTestMandate` is itself only there for testing purposes I'm not convinced it's worthwhile to write a test for it. The changes to other things in this PR are all trivial so also don't seem worth testing.

PR #987, opened 2024-08-29
MAT-376: WIP (but mergable) create new pending donations based on regular giving man…

bdsl:
Sets up various prerequisites for creating new donations based on regular giving mandates. Adds some new incomplete functions for creating those donations to be continued in a later PR.

Would have been neater to either know the prerequisites in advance and do just them, or do the overall functions first with the prerequisite stuff stubbed out, this is a slightly awkward mix of outside-in and inside-out working, but should be OK.

Would also be nice to have more tests but the work so far was so CRUD-based it's hard to see where to put tests. Next thing may be the logic for converting a sequence number to a date, which will definitely be unit tested, and for creating a donation object with that will will also be unit tested.

### DIFF
--- a/src/Domain/DonationSequenceNumber.php
+++ b/src/Domain/DonationSequenceNumber.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace MatchBot\Domain;
+
+use MatchBot\Application\Assertion;
+
+/**
+ * Used for a donation given as part of a regular giving mandate. May not be duplicated within one mandate, and
+ * indicates the date the donation should be taken, e.g. #1 indicates the donation taken at time of mandate creation,
+ * #2, will be taken one month (or possibly other regular period) later, #3 will be taken two months later etc.
+ *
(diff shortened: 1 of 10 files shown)


src/Domain/DonationSequenceNumber.php
```diff
@@ -0,0 +1,28 @@
...
+ * #2, will be taken one month (or possibly other regular period) later, #3 will be taken two months later etc.
+ *
+ * @psalm-suppress PossiblyUnusedProperty - to be used soon.
+ */
+readonly class DonationSequenceNumber
+{
+    private function __construct(
+        public int $number
+    ) {
+        // having a mandate last for 100 years is ambitious, putting some upper limit in mostly to because its better
```
bdsl, line 19:
```suggestion
        // having a mandate last for 100 years is ambitious, putting some upper limit in mostly because its better
```

NoelLH (approved):
All looks good!

PR #993, opened 2024-09-03
MAT-376: Coplete MandateService::makeNextDonationForMandate function …

bdsl:
…implentation


src/Application/Commands/TakeRegularGivingDonations.php
```diff
@@ -33,7 +27,7 @@ public function __construct(
         private DonationRepository $donationRepository,
         private DonationService $donationService,
         private MandateService $mandateService,
-        private EntityManager $em,
+        private EntityManagerInterface $em,
```
bdsl, line 30: this `EntityManager` mistake I made caused errors that took hours to find

NoelLH (approved):
Looks good!

PR #994, opened 2024-09-03
BG2-2722 – Cache bust as needed for SF change messages

NoelLH:
Ask single Campaign & Fund Salesforce API endpoints not to cache when we were just specifically told there is an update.


bdsl (approved):
Looks good with the related changes in other repos - maybe possible to do things a bit more simply entirely within this repo by adding a random or time based query param to force a cache miss but this also works.

PR #1012, opened 2024-09-12
DON-896: Remove support for confirming donation by stripe payment int…

bdsl:
…ent ID

Do not deploy to prod until after https://github.com/thebiggive/donate-frontend/pull/1644 has been deployed and had time to reach all user browsers


NoelLH (commented):
This'll need a slight tweak if we apply the class typo fix from https://github.com/thebiggive/matchbot/pull/1013, but otherwise all looks good

bdsl:
@NoelLH Thanks, rebased on to origin/develop. Did rebase rather than merge to make sure all changes are visible and nothing is hidden in a merge commit.

NoelLH (approved):

bdsl:
Merged to develop but we shouldn't release to prod for several hours (so in practice until Monday) to make sure there are no copies of the old frontend still active in browsers.

PR #1013, opened 2024-09-12
DON-896 – review tweaks

NoelLH:
Everything else tagged with the issue on `develop` now LGTM


bdsl (approved):
Added one suggested change, otherwise all LGTM.

PR #1024, opened 2024-10-01
MAT-345 – add command to send CloudWatch donation statistics

NoelLH:


composer.lock
bdsl, line 1: I now get `Package phpcsstandards/php_codesniffer is abandoned, you should avoid using it. Use squizlabs/php_codesniffer instead.` . I think squiz must have allowed the new phpcs standards people to take over their namespace after abandoing the project months ago so we should change it a second time.

NoelLH: Yeah, I also noticed this. But both projects' current readmes point at the other one, so I gave up trying to work it out.

src/Application/Commands/SendStatistics.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * Value in default 'Unit' mode accepts doubles, so we leave this with the default for both float
+     * & int values.
+     */
+    private function makeMetric(string $name, float | int | null $value, \DateTimeImmutable $timestamp): array
+    {
+        $appEnv = getenv('APP_ENV');
+        \assert(is_string($appEnv)); // crash if somehow missing.
```
bdsl, line 75: I think it's worth injecting the appEnv - something like this using the existing `\MatchBot\Application\Environment` enum class. Except I'd put move the strtolower inside that class (or replace it with a match)

```suggestion
        $appEnv = strtolower($this->environment->name);
```

src/Application/Commands/SendStatistics.php
```diff
@@ -0,0 +1,82 @@
...
+        private CloudWatchClient $cloudWatchClient,
+        private DonationRepository $donationRepository
+    ) {
+        parent::__construct();
+    }
+
+    protected function doExecute(InputInterface $input, OutputInterface $output): int
+    {
+        $time = time();
+        $startOfThisMinute = new \DateTimeImmutable('@' . ($time - ($time % 60)));
```
bdsl, line 36: The logic for finding the start of the minute is effectively untested, since its repeated in the test class and just compared to itself. I think it probably is worth injecting either a clock or a timestamp so we can use fixed values in the test to see what the result of this calculation is. A timestamp would be enough, if you want to do a clock we'd use it like this:

```suggestion
        $time = $this->clock->now()->getTimestamp();
        $startOfThisMinute = new \DateTimeImmutable('@' . ($time - ($time % 60)));
```

src/Domain/DonationRepository.php
```diff
@@ -455,6 +455,78 @@ public function findWithTransferIdInArray(array $transferIds): array
...
+        $result = $qb->getQuery()->getOneOrNullResult(Query::HYDRATE_ARRAY);
+
+        if ($result === null || $result['donationCount'] < 20) {
+            return null;
+        }
+
+        return (float) $result['completeCount'] / $result['donationCount'];
+    }
+
+    public function getDonationsJustCreated(\DateTimeImmutable $startOfThisMinute): int
```
bdsl, line 498: might be slightly clearer named without assuming the given minute is now, and also count instead of get? e.g. 
```suggestion
    public function countDonationsCreatedInMinuteTo(\DateTimeImmutable $minuteEnd): int
```

The given param is the end of the period of interest, not the beginning. (true its also going to be the start of the current minute at runtime, but repo code doesn't need to know that)

Same applies to the following function.

app/dependencies.php
```diff
@@ -183,6 +184,17 @@ function (ContainerInterface $c): Auth\DonationPublicAuthMiddleware {
             return new Client\LiveStripeClient($c->get(StripeClient::class));
         },
 
+        CloudWatchClient::class => static function (): CloudWatchClient {
+            return new CloudWatchClient([
+                'version' => 'latest',
+                'region' => getenv('AWS_REGION'),
```
bdsl, line 190: this is going to cause a crash any time we run the matchbot CLI in an env that doesn't have the region set in the environment, even if we aren't using the new command, so we'd have to do the infrastructure roll out first. Can we make it default to something that will be OK as placeholder as long as we don't actually need to run the stats command? Or get the env changes rolled out before merging this PR?

src/Domain/DonationRepository.php
```diff
@@ -455,6 +455,78 @@ public function findWithTransferIdInArray(array $transferIds): array
...
+            ->setParameter('end', $oneMinuteAgo)
+            ->setParameter(
+                'completeStatuses',
+                array_map(static fn(DonationStatus $s) => $s->value, DonationStatus::SUCCESS_STATUSES),
+            );
+
+        /**
+         * @var array{donationCount: int, completeCount: int}|null $result
+         */
+        $result = $qb->getQuery()->getOneOrNullResult(Query::HYDRATE_ARRAY);
```
bdsl, line 489: I do still think DQL is quite a bit easier to read and edit than the query builder, and is well supported by PHPStorm. We lose out a bit on Psalm checks but I don't think Psalm does enough to be worthwhile. I converted it here by adding var dumping the output of `$qb->getQuery()->getDql()` and then reformatting by hand:

```suggestion
        $query = $this->getEntityManager()->createQuery(<<<'DQL'
            SELECT
            COUNT(d.id) as donationCount, 
            SUM(CASE WHEN d.donationStatus IN (:completeStatuses) THEN 1 ELSE 0 END) as completeCount
            FROM MatchBot\Domain\Donation d 
            LEFT JOIN d.fundingWithdrawals fw
            WHERE d.createdAt >= :start
            AND d.createdAt < :end HAVING SUM(fw.amount) > 0
        DQL
        );
        $query->setParameter('start', $sixteenMinutesAgo);
        $query->setParameter('end', $oneMinuteAgo);

        /**
         * @var array{donationCount: int, completeCount: int}|null $result
         */
        $result = $query->getOneOrNullResult(Query::HYDRATE_ARRAY);
```

bdsl: Same applies to other uses of `createQueryBuilder` below.

app/dependencies.php
NoelLH: I don't see why we can't roll out the infra changes from https://github.com/thebiggive/infrastructure/pull/284 ahead of this PR going in. I'll merge ahead of review now so we can spot any apply issues in Staging early.

src/Domain/DonationRepository.php
NoelLH: It's true that I'm getting completion suggestions with either option. I do think there are stronger checks when building the query in code, e.g. getting clear errors about entity classes that don't exist, without waiting until runtime. But have changed all the new queries to DQL for now.

bdsl: Yep. We could get https://github.com/phpstan/phpstan-doctrine to do static checks for DQL. But I don't think it's hugely important, we only have to run it once to catch runtime errors. It's not like we'd have to run multiple scenarios for different code paths.

src/Application/Commands/SendStatistics.php
NoelLH: Yep makes sense, forgot we already had that. Done in d2d92a285d23985fa621cb56ac63e47a25f3f5b8

src/Application/Commands/SendStatistics.php
NoelLH: Added testing against a specific time in 4bda1919d6407cc86ef89efec3cc6f1f1da62d67

src/Domain/DonationRepository.php
NoelLH: Updated some var names in 4bda1919d6407cc86ef89efec3cc6f1f1da62d67

src/Application/Commands/CallFrequentTasks.php
```diff
@@ -0,0 +1,46 @@
...
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+#[AsCommand(
+    name: 'matchbot:tick',
+    description: 'Calls all per-minute commands; currently to expire old matching and send statistics'
+)]
+class CallFrequentTasks extends LockingCommand
```
bdsl, line 17: liking having a CallFrequentTasks command a lot if it means we can add new ones without having to make infrastructure repo changes. We could remove some of the duplication here but not much point while its just calling two commands. Eventually I still think using something like https://github.com/peppeocchi/php-cron-scheduler would make sense.

tests/Application/Commands/SendStatisticsTest.php
```diff
@@ -0,0 +1,126 @@
...
+use Symfony\Component\Console\Tester\CommandTester;
+use Symfony\Component\Lock\LockFactory;
+
+class SendStatisticsTest extends TestCase
+{
+    private \DateTimeImmutable $startOfMockClocksCurrentMinute;
+
+    public function setUp(): void
+    {
+        $this->startOfMockClocksCurrentMinute = new \DateTimeImmutable('@1727866800');
```
bdsl, line 23: I might make another PR when I have time just to use a more readable date/time format here and in the MockClock constructor - there's no way to see at a glance that this is the start of a minute but I do see that its two seconds before the clock is set to so is reasonable.

bdsl (approved):
LGTM

src/Application/Commands/CallFrequentTasks.php
bdsl: Replied on slack

PR #1027, opened 2024-10-03
Update API contract

NoelLH:
Also proposes to start migrating `DonorAccount\Create` to an endpoint more consistent with others (supporting both for now).

### DIFF
--- a/api.yaml
+++ b/api.yaml
@@ -32,8 +32,10 @@ paths:
         in: path
         schema:
           type: string
+          format: uuid
         description: Identifier of the donor
         required: true
+
     post:
       operationId: createDonation
       summary: Create a donation on behalf of a specified person
@@ -52,6 +54,28 @@ paths:
           application/json:
             schema:
               $ref: '#/components/schemas/Donation'
+
+    get:
+      operationId: listDonations
+      summary: List existing complete donations made by the person
+      tags: [Past donations]
+      security:
+        - JWT: []
+      responses:
+        '200':
+          description: List of donations
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  donations:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/Donation'
+        '400':
+          description: Required parameters not set as expected
+
   /donations/{donationId}:
     parameters:
       - name: donationId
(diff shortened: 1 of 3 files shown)


api.yaml
```diff
@@ -2,51 +2,94 @@ openapi: 3.0.0
...
+        schema:
+          type: string
+          format: uuid
+        description: Identifier of the donor
+        required: true
+
     post:
-      summary: Creates a donation
-      operationId: addDonation
-      description: Initiates a donation with pending status
```
bdsl, line 22: is there a reason for deleting the mention of the pending status? I don't know if we need to mention the status by name but may be helpful to mention that this is legitimately used when someone is only *considering* donating. "Create a donation" could be read as much more final than that. But maybe it doesn't matter much for the audience of this document.

api.yaml
```diff
@@ -2,51 +2,94 @@ openapi: 3.0.0
...
       requestBody:
         content:
           application/json:
             schema:
               $ref: '#/components/schemas/Donation'
-        description: Donation to create
+
+    get:
+      operationId: listDonations
+      summary: List existing complete donations made by the person
```
bdsl, line 60: do we need to document that it looks like the implementation actually ignores the person ID in the URL and only uses the ID in the JWT header sent with the request? I think this is actually the same for createDonation, and I think I considered not having the person ID in the URL when creating this route for that reason.

bdsl (approved):

api.yaml
NoelLH: Not a specific one, no. This diff is actually the result of deleting the (non-`personId`) endpoint entirely. I will merge the description in with `createDonation`'s to further clarify.

api.yaml
NoelLH: I think the token verifier & middleware do require a match don't they @bdsl? If I'm reading it correctly then a mismatch gives you
> JWT error: Not authorised for person ID ...

It's true that which person the request is for is over-specified but I think it's worth it to have clearer & more consistent URIs when looking at request logs etc.

api.yaml
NoelLH: Done in 8525d41d53139507a331ed530ba06157bbcfdfc9

PR #1028, opened 2024-10-04
Update deps, fix issues found with psalm

bdsl:


composer.json
```diff
@@ -67,12 +67,12 @@
     "require-dev": {
         "doctrine/event-manager": "^2.0",
         "infection/infection": "^0.27.11",
-        "phpcsstandards/php_codesniffer": "^3.10",
         "phpspec/prophecy": "^1.19",
         "phpspec/prophecy-phpunit": "^v2.1.0",
         "phpunit/phpunit": "^9.6",
         "psalm/plugin-phpunit": "^0.19.0",
         "roave/security-advisories": "dev-latest",
+        "squizlabs/php_codesniffer": "^3.10",
```
bdsl, line 75: as discussed at https://github.com/thebiggive/matchbot/pull/1024#discussion_r1784159326

src/Domain/FundRepository.php
```diff
@@ -161,10 +162,15 @@ protected function setAnyFundData(Fund $fund, array $fundData): Fund
 
     protected function getNewFund(array $fundData): Fund
     {
+        $currencyCode = $fundData['currencyCode'] ?? 'GBP';
+        $name = $fundData['name'] ?? '';
```
bdsl, line 166: same values as at will be set just a moment later when the execution reaches here https://github.com/thebiggive/matchbot/blob/7434d20702494dd1a91f0ff7e31d270e7d87abf1/src/Domain/FundRepository.php#L156-L157

tests/Application/Commands/HandleOutOfSyncFundsTest.php
```diff
@@ -19,6 +19,18 @@
...
+    private CampaignFunding $fundingInSync;
+    private CampaignFunding $fundingOverMatched;
+    private CampaignFunding $fundingUnderMatchedWithNothingAllocated;
+
+    public function setUp(): void
+    {
+        $this->fundingUnderMatched = $this->getFundingUnderMatched();
+        $this->fundingInSync = $this->getFundingInSync();
+        $this->fundingOverMatched = $this->getFundingOverMatched();
+        $this->fundingUnderMatchedWithNothingAllocated = $this->getFundingUnderMatchedWithNothingAllocated();
```
bdsl, line 32: as these objects now contain timestamps we can't construct them multiple times and expect them to be equal. If this test was left unchanged it would be failing. So constructing once in setUp instead of having multiple calls to these get functions.

dorota-joanna (approved):
LGTM

PR #1030, opened 2024-10-04
Fix 'PropertyNotSetInConstructor' issues in CampaignFunding class

bdsl:
This refactoring PR does technically add more lines than it removes, but the added lines are substantially simpler than the removed lines.


dorota-joanna (dismissed):
LGTM

PR #1032, opened 2024-10-09
DON-998: add class Actions\RegularGivingMandate\Create with pseudocode

bdsl:


src/Application/HttpModels/MandateCreate.php
```diff
@@ -0,0 +1,38 @@
...
+
+    /** @var Salesforce18Id<Campaign>  */
+    public Salesforce18Id $campaignId;
+
+    /**
+     * @psalm-suppress PossiblyUnusedMethod - called by Symfony Serializer
+     */
+    public function __construct(
+        // not taking any value objects as constructor params as at least with our current config or
+        // any config I've been able to find the Symfony Serializer doesn't initialise their properties.
```
NoelLH, line 27: I think this is all good, but would be helpful to add the new schema & endpoint to the OpenAPI as we may need to get updates to the vendor quite quickly if they come back on a time for testing before CC

src/Application/Actions/RegularGivingMandate/Create.php
```diff
@@ -0,0 +1,112 @@
...
+            $this->logger->info("Mandate Create non-serialisable payload was: $body");
+
+            $message = 'Donation Create data deserialise error';
+            $exceptionType = get_class($exception);
+
+            return $this->validationError(
+                $response,
+                "$message: $exceptionType - {$exception->getMessage()}",
+                $message,
+                empty($body), // Suspected bot / junk traffic sometimes sends blank payload.
```
NoelLH, line 62: I think this is hopefully less likely with this new, more niche endpoint. Should we start with error level logs until we know otherwise?

NoelLH (approved):
Looks good! 2 things inline, not blockers

src/Application/HttpModels/MandateCreate.php
bdsl: Right - I thought I'd implement the consumer in frontend though, and there will probably be a bit of iteration on API design between the two. Do you think we should add stuff to the OpenAPI spec from the start and keep it updated with every change we merge to matchbot?

src/Application/Actions/RegularGivingMandate/Create.php
bdsl: yes I think that's fine. I think I will also change the middleware now to be PersonWithPassword which will probably make bots a lot less likely to access it.

src/Application/HttpModels/MandateCreate.php
NoelLH: I don't think we need to incorporate every change but we might end up backed into having to do things fairly urgently if we don't do an initial draft as we go. It's a bit hard to say until I can get hold of the vendor again, so let's just keep it in mind / chat in standup

src/Application/Actions/RegularGivingMandate/Create.php
bdsl: Done in 822e9d91

src/Application/HttpModels/MandateCreate.php
bdsl: OK, I've pushed a couple of commits to add the create request to the api spec. I also adjusted the spec there that's already there for the returned object as used in `regular-giving/my-donation-mandates` to better match the implementation, and returned the same type of object in the new endpoint.

api.yaml
```diff
@@ -291,7 +291,21 @@ paths:
           description: Donor account features enabled
         '400':
           description: Invalid data or already enabled
-
+  /people/{personId}/regular-giving:
```
NoelLH, line 294: I think this needs a param def to by valid syntax:

> Declared path parameter "personId" needs to be defined as a path parameter at either the path or operation level

Slightly frustratingly it seems this has to be declared repetitively for each path and can't be `$ref`'d to a shared definition.

NoelLH (commented):
There is one thing making the OpenAPI validator unhappy. (I started trying to suggest inline but wasn't sure about the YAML indent here). Otherwise all LGTM now, thanks!

bdsl:
> There is one thing making the OpenAPI validator unhappy. (I started trying to suggest inline but wasn't sure about the YAML indent here). Otherwise all LGTM now, thanks!

Where  are you seeing that? Or how are you running the validator?

NoelLH:
> > There is one thing making the OpenAPI validator unhappy. (I started trying to suggest inline but wasn't sure about the YAML indent here). Otherwise all LGTM now, thanks!
> 
> Where are you seeing that? Or how are you running the validator?

I copied the whole file into SwaggerHub

PR #1034, opened 2024-10-11
Reused confirmation token

bdsl:


NoelLH (commented):
The addition to the list looks good.

I wonder if we should also at this point remove the logic specific to _payment method_ reuse, from declaring `$paymentMethodReuseAttempted` onwards? It seems like maybe now we use confirmation tokens this is redundant. And I'm not sure the special cases add enough value to be worth repeating for ctokens, although perhaps the explanatory comment is worth updating and moving to the list of messages.

bdsl:
@NoelLH yes I think that's a good idea. Done in 5f612f7c

PR #1041, opened 2024-10-16
MAT-377 – message push enhancements: improve traceability, reduce chance of unhelpful reverts

NoelLH:
* [Delay handling of Stripe payment webhooks' queued messages](https://github.com/thebiggive/matchbot/commit/b15ba32c85a029b9a7a0aa1eda1722f2ed050d71); May reduce the risk of these important updates being superseded unhelpfully by overlapping ones
* Give all messages unique IDs
* Log those ID with all sends & receives
* Log donation UUID alongside where applicable

~~I've not tested the middleware yet, so draft for now.~~

### DIFF
--- a/src/Application/Messenger/Middleware/AddOrLogMessageId.php
+++ b/src/Application/Messenger/Middleware/AddOrLogMessageId.php
@@ -17,6 +17,7 @@ readonly class AddOrLogMessageId implements MiddlewareInterface
     {
     }
 
+    /** @psalm-suppress PossiblyUnusedReturnValue sig must match interface */
     public function handle(Envelope $envelope, StackInterface $stack): Envelope
     {
         $existingMessageIdStamp = $envelope->last(MessageId::class);
--- a/src/Application/Messenger/Middleware/AddOrLogMessageId.php
+++ b/src/Application/Messenger/Middleware/AddOrLogMessageId.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MatchBot\Application\Messenger\Middleware;
+
+use MatchBot\Application\Messenger\DonationUpserted;
+use MatchBot\Application\Messenger\Stamp\MessageId;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\Middleware\MiddlewareInterface;
+use Symfony\Component\Messenger\Middleware\StackInterface;
+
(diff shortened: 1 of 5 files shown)


src/Application/Messenger/Middleware/AddOrLogMessageId.php
```diff
@@ -0,0 +1,59 @@
...
+use Symfony\Component\Messenger\Middleware\MiddlewareInterface;
+use Symfony\Component\Messenger\Middleware\StackInterface;
+
+readonly class AddOrLogMessageId implements MiddlewareInterface
+{
+    public function __construct(private LoggerInterface $logger)
+    {
+    }
+
+    /** @psalm-suppress PossiblyUnusedReturnValue sig must match interface */
```
NoelLH, line 20: Somehow the addition of the unit test turns this into an error, unless I suppress. Not clear why given the file is also new and so presumably not in the baseline. Removing the return type isn't possible as it violates the interface.

bdsl: I think for things like this I normally just put a comment to say that its called by the framework (or Symfony Messenger in this case), not by us.

src/Application/Messenger/Stamp/MessageId.php
```diff
@@ -0,0 +1,24 @@
...
+
+declare(strict_types=1);
+
+namespace MatchBot\Application\Messenger\Stamp;
+
+use Ramsey\Uuid\Uuid;
+use Ramsey\Uuid\UuidInterface;
+use Symfony\Component\Messenger\Stamp\StampInterface;
+
+class MessageId implements StampInterface
```
bdsl, line 11:
```suggestion
readonly class MessageId implements StampInterface
```

bdsl (dismissed):

src/Application/Messenger/Middleware/AddOrLogMessageId.php
NoelLH: Right – made it a little clearer in d050e112dd784978c0565f37b49a575f9c8205f7. It's still odd to me that it is the presence of our test call that causes the error and that zero calls in our codebase is allowed – no `PossiblyUnusedMethod` in either case.

NoelLH:
Did a bit more of a local test (which uses Redis but should behave as SQS will). I seem to be getting the INFO logs I wanted on both send and handle of messages. Message UUIDs match up and are printed alongside donation UUID for easy searching.

PR #1043, opened 2024-10-21
DON-998: Partially implement restriction on creating ad-hoc donation …

bdsl:
…for regular giving campaign or vice-versa

Still need to actuall pull data from SF to make a real distinction, and save it into our DB when SF has that data - see BG2-2711


src/Domain/Campaign.php
```diff
@@ -77,6 +77,26 @@ public function __toString(): string
         return "Campaign ID #{$this->id}, SFId: {$this->salesforceId}";
     }
 
+    /**
+     * @return bool does this campaign accept ad-hoc giving, i.e. non-regular giving.
+     * @see self::isRegularGiving()
+     */
+    public function isAdHocGiving(): bool
```
bdsl, line 84: not sure what we want to call the opposite of regular giving. I've gone with ad-hoc giving but there might be a better term.

bdsl: changed to one-off

src/Domain/RegularGivingService.php
```diff
@@ -2,15 +2,160 @@
...
+        );
+        $this->donationService->enrollNewDonation($firstDonation);
+        if (! $firstDonation->isFullyMatched()) {
+            throw new NotFullyMatched(
+                "Donation could not be fully matched, need to match {$firstDonation->getAmount()}," .
+                " only matched {$firstDonation->getFundingWithdrawalTotal()}"
+            );
+        }
+        // @todo-regular-giving - throw if first donation is not fully matched unless donor has said they're OK with
+        //                        that.
```
NoelLH, line 103:
```suggestion
```

NoelLH: And maybe the todo's below? I've slightly lost the thread of which parts were already on `develop`, could be the later parts were done already?

NoelLH (dismissed):
Minor non-blockers inline, LGTM

src/Domain/RegularGivingService.php
```diff
@@ -2,15 +2,158 @@
...
+            mandateSequenceNumber: DonationSequenceNumber::of(1)
+        );
+        $this->donationService->enrollNewDonation($firstDonation);
+        if (! $firstDonation->isFullyMatched()) {
+            throw new NotFullyMatched(
+                "Donation could not be fully matched, need to match {$firstDonation->getAmount()}," .
+                " only matched {$firstDonation->getFundingWithdrawalTotal()}"
+            );
+        }
+        // @todo-regular-giving - collect first donation
```
bdsl, line 102:
```suggestion
        // @todo-regular-giving - collect first donation (currently created as pending, not collected)
```

bdsl:
Thanks @NoelLH

PR #1045, opened 2024-10-23
Also catch `MissingConstructorArgumentsException` and `AssertionFailedException` on Donation Update deserialise

NoelLH:
With our latest libs & constructor, this is now what happens if you leave out e.g. donation amount (missing constructor arg) or provide an invalid format email address (assertion valid).

We want an HTTP 400 and warning log for both of these, not 500 and error.


bdsl (approved):
LGTM

PR #1051, opened 2024-10-30
MAT-382 – fix fee calculation regression

NoelLH:


bdsl (approved):
Looks good - [src/Domain/DonationService.php](https://github.com/thebiggive/matchbot/pull/1051/files#diff-bd27501321f38ab1ad31b1e5c4640c8f32b128dd8b282103ad0a6801476c87f8) is the only file changed that's actually used in production, and the need for the change is very clear. So let's get this into prod as soon as we reasonably can and then think about any additional work we want to after.

PR #1000, opened 2024-09-06
DON-896: Add support for confirming donation by confirmation token ID…

bdsl:
… instead of payment method ID

See
https://docs.stripe.com/payments/payment-element/migration-ct?locale=en-GB and
https://docs.stripe.com/api/confirmation_tokens/object

Must be in prod before the related feature flag in frontend is enabled

PR #1053, opened 2024-10-30
MAT-383 – add a command to report on and return accidentally overcharged fees

NoelLH:


src/Application/Commands/ReturnErroneousExcessFees.php
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        if (!$card) {
+            throw new \LogicException('Cannot continue with no card on charge');
+        }
+
+        $cardBrand = $card->brand;
+        $cardCountry = $card->country;
+
+        return Calculator::calculate(
```
bdsl, line 140: the result of this calculation will be the same as what we recorded on the donation in the database, won't it? I was going to ask if we should use that instead of recalculating, but I guess a better option is to use both, and throw if they're different, for an extra check?

src/Domain/DonationRepository.php
```diff
@@ -452,6 +452,30 @@ public function findWithTransferIdInArray(array $transferIds): array
...
+    {
+        $query = $this->getEntityManager()->createQuery(<<<'DQL'
+            SELECT d
+            FROM MatchBot\Domain\Donation d 
+            WHERE d.collectedAt >= :start
+            AND d.donationStatus IN (:completeStatuses)
+            AND d.paymentMethodType = 'card'
+        DQL
+        );
+        $query->setParameter('start', new \DateTimeImmutable('2024-09-13 00:00:00'));
```
bdsl, line 468: Shouldn't this be September 12 instead of 13? The bug was introduced in https://github.com/thebiggive/matchbot/pull/1000 on Sept 6 and the first deploy to prod after that was https://github.com/thebiggive/matchbot/pull/990 on September 12.  Or we could just put it back to September 6th or even 1st, since checking a extra donations isn't a problem.

src/Application/Commands/ReturnErroneousExcessFees.php
```diff
@@ -0,0 +1,149 @@
...
+            $feeId = (string) $charge->application_fee;
+            $fee = $this->stripeClient->applicationFees->retrieve($feeId);
+
+            if ($fee->amount_refunded > 0) {
+                // Might be because the donation itself was fully refunded and the fee reversed. Carry on.
+                $output->writeln("Donation {$donation->getId()} has already had fee refunded");
+                continue;
+            }
+
+            $feeDifference = $fee->amount - $donation->getAmountToDeductFractional();
```
bdsl, line 82: rename to $feeDiferencePence ?

src/Application/Commands/ReturnErroneousExcessFees.php
```diff
@@ -0,0 +1,149 @@
...
+
+            if ($donation->getCharityFee() !== $this->getCorrectFees($donation, $charge)->coreFee) {
+                $output->writeln("Donation {$donation->getId()} has a different fee than expected");
+                continue;
+            }
+
+            $countOfDonationsToChange++;
+            $sumOfFeeDifference = bcadd($sumOfFeeDifference, (string) ($feeDifference / 100), 2);
```
bdsl, line 90: rename to $sumOfFeeDifferencePounds ?

bdsl (dismissed):
Added comments inline, not blockers. Otherwise all LGTM but I feel like the logic is critical enough that we probably should write tests for it even though its temporary. I agree we can use the 'check' mode as the tests, since all most all the same code runs in 'check' and 'fix' modes.

src/Application/Commands/ReturnErroneousExcessFees.php
bdsl: oh yes where you output `Donation {$donation->getId()} has a different fee than expected`. Makes sense. Could that throw or at least log an error instead of just outputting? We don't expect it to ever happen, do we?

src/Domain/DonationRepository.php
NoelLH: Yes, looks like I inferred wrong from Jira.

src/Application/Commands/ReturnErroneousExcessFees.php
NoelLH: Done in 975a424bb1d2decfcd7f4cd6298666590505b3cf

src/Application/Commands/ReturnErroneousExcessFees.php
NoelLH: Done in 975a424bb1d2decfcd7f4cd6298666590505b3cf

bdsl (approved):

src/Application/Commands/ReturnErroneousExcessFees.php
NoelLH: Sorry just saw this. I was thinking we would probably search all the output for it on a bigger run. Can make it an error on my next PR too.

PR #1054, opened 2024-10-30
Get shared stamp class from `thebiggive/messages`

NoelLH:
And update Symfony libs in tandem as needed

Includes removing some now-extraneous annotation help, presumably due to Symfony providing better typing


bdsl:
Approving here rather than with the button for merging only when we're ready to deploy to prod with https://github.com/thebiggive/claimbot/pull/64.

PR #1061, opened 2024-10-31
Don 1082 donation funds tip auto cancel take ii

bdsl:


src/Application/Actions/Donations/Update.php
```diff
@@ -162,15 +162,18 @@ protected function action(Request $request, Response $response, array $args): Re
                     try {
                         $this->donationService->cancel($donation);
                     } catch (DonationAlreadyFinalised $e) {
+                        $this->entityManager->rollback();
```
bdsl, line 165: had to move around quite a bit of transaction stuff. Would like to try and keep to a rule of always having any calls related to the same transaction in the same function - I made a mistake before by moving code with a transaction commit out of here into a service class while keeping the transaction start and rollbacks here.

Could force that by using \MatchBot\Application\Persistence\RetrySafeEntityManager::wrapInTransaction instead but that's doing more stuff and is a bigger change, especially since our implementation of that has retrying built-in.

src/Domain/DonationService.php
```diff
@@ -467,15 +467,13 @@ public function cancel(Donation $donation): void
...
             $this->logger->info("Donation ID {$donation->getUuid()} was already Cancelled");
-            $this->entityManager->rollback();
 
             return;
         }
 
         if ($donation->getDonationStatus()->isSuccessful()) {
             // If a donor uses browser back before loading the thank you page, it is possible for them to get
             // a Cancel dialog and send a cancellation attempt to this endpoint after finishing the donation.
-            $this->entityManager->rollback();
```
bdsl, line 478: this rollback is moved into the catch in the action class

src/Application/Actions/Donations/Update.php
NoelLH: Yeah, I think there's a trade off between those for sure. I like `wrapInTransaction()` better conceptually, but above a certain size of transaction I feel like it makes the code harder to follow again.

Am I right in thinking these changes reverse what I found about `DonationService::cancel()` needing an open transaction? If so then my comment on the fn from yesterday should go too to avoid adding more confusion.

bdsl: Yes, thanks for spotting. Done in 9c4db075

bdsl: Not about to try to build it but I guess it might be nice if there was an interface where you could do something like:

```
$transaction = $entityManager->beginTransaction(); // (or $this->transaction = ...)
/// lots of code
$wasSucessful ? $transaction->commit(); $transaction->rollback();
 ```

NoelLH (approved):

PR #1063, opened 2024-11-04
Mat 381 pull all

bdsl:


src/Application/Commands/PullMetaCampaignFromSF.php
```diff
@@ -0,0 +1,72 @@
...
+        parent::__construct();
+    }
+    public function configure(): void
+    {
+        $this->addArgument('metaCampaignSlug', InputArgument::REQUIRED);
+    }
+    protected function doExecute(InputInterface $input, OutputInterface $output): int
+    {
+        // Steps:
+        // 1. Use SF API to get list of campaigns withi meta campaign by hitting e.g.
```
dorota-joanna, line 48:
```suggestion
        // 1. Use SF API to get list of campaigns within meta campaign by hitting e.g.
```

dorota-joanna (approved):
LGTM

NoelLH (commented):
Mostly looks good to me, minor things inline. Following up wider discussion about the API on Jira.

bdsl:
@NoelLH I've updated this to fetch up to 10,000 campaigns from SF by paging through the API, instead of just taking the first page of 100. Still think we could make it quite a lot more efficient by changing the SF side to either return a list of IDs only or return multiple campaigns at once as needed by matchbot.

src/Domain/CampaignRepository.php
```diff
@@ -98,6 +98,45 @@ public function findOneBySalesforceId(Salesforce18Id $salesforceId): ?Campaign
...
+        $count = count($campaignIds);
+
+        $i = 0;
+        foreach ($campaignIds as $id) {
+            $i++;
+            $this->logger->info("Fetching campaign $i of $count\n");
+            $existingCampaignInDB = $this->findOneBySalesforceId($id);
+
+            if ($existingCampaignInDB) {
+                $this->updateFromSf($existingCampaignInDB, withCache: false, autoSave: true);
```
NoelLH, line 129: Should we also add `$this->fundRepository->pullForCampaign($campaign);` around here as you suggested earlier? I think that would make the effect pretty much identical to what `UpdateCampaigns` does for already known ones.

bdsl: Yes I think so. I didn't want to make CampaignRepo depend on FundRepo so I've done it as a second phase after this loop completes in f80ca87efa0

NoelLH: Nice, looks good!

NoelLH (approved):

PR #1074, opened 2024-11-14
DON-1085 – add Donation Funds tip cancellation support

NoelLH:
Technically allow cancelling all donations to any one campaign + payment method type, for the current donor. Also group routes a bit more to reduce duplicated definition of relevant middlewares.


dorota-joanna (approved):
LGTM

PR #1076, opened 2024-11-18
Skip over new donations without transaction / payment intent IDs when…

bdsl:
… pushing to SF

This happens if there was an error creating the payment intent ID. Currently it generates an alarming unhandled execption, this PR should reduce it to a warning.


dorota-joanna (approved):
LGTM

PR #1085, opened 2024-11-20
Catch invalid billing details JSON and return a 400

NoelLH:


bdsl (approved):
LGTM. Probably there are quite a few other places where do json_decode so might want to apply the same thing in future.

PR #1092, opened 2024-11-26
MAT-391 – delete pledges loaded in advance of CC24 that are no longer available

NoelLH:


src/Migrations/Version20241126163843.php
```diff
@@ -0,0 +1,124 @@
...
+                'amount' => 2_500,
+            ],
+        ];
+
+        Assertion::count($deleteSets, 14);
+
+        foreach ($deleteSets as $deleteSet) {
+            $this->addSql('DELETE FROM Campaign_CampaignFunding WHERE campaign_id = :campaignId AND campaign_funding_id = :campaignFundingId LIMIT 1', $deleteSet);
+            $this->addSql('DELETE FROM CampaignFunding WHERE id = :campaignFundingId AND fund_id = :fundId AND amount = :amount LIMIT 1', $deleteSet);
+            $this->addSql('DELETE FROM Fund WHERE fundType = "pledge" AND id = :fundId LIMIT 1', $deleteSet);
```
bdsl, line 116: don't think I ever noticed before that `\Doctrine\Migrations\AbstractMigration::addSql` takes more than one argument!

PR #1093, opened 2024-11-27
Hotfix for sf pull type error

bdsl:
Not sure why this is null though, I thought the data from sf would always include status. Not an issue for now for it to be null in matchbot.

### DIFF
--- a/src/Domain/Campaign.php
+++ b/src/Domain/Campaign.php
@@ -238,7 +238,7 @@ class Campaign extends SalesforceReadProxy
     public function updateFromSfPull(
         Charity $charity,
         string $currencyCode,
-        string $status,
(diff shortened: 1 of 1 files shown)


src/Domain/Campaign.php
```diff
@@ -238,7 +238,7 @@ public function isReady(): bool
     public function updateFromSfPull(
         Charity $charity,
         string $currencyCode,
-        string $status,
+        ?string $status,
```
bdsl, line 241: see caller passing `status: $campaignData['status']` at https://github.com/thebiggive/matchbot/blob/2101a75e525b12976aa44b4066e69c7a1475d0df/src/Domain/CampaignRepository.php#L182

PR #1096, opened 2024-11-28
MAT-392 – further CC24 funding patches

NoelLH:


src/Migrations/Version20241128165753.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        /** @var array{array{campaignId: int, campaignFundingId: int, fundId: int, amount: float}} $deleteSets */
+        $deleteSets = [
+            [
+                'campaignId' => 8289, // SF a056900002TPSp9AAH
+                'campaignFundingId' => 36212,
+                'fundId' => 29542, // SF a0AWS000001R1p32AC
+                'amount' => 900,
+            ],
+        ];
```
bdsl, line 31: this seems to match up with what's on JIRA & slack.

bdsl (approved):
LGTM, all seems to match up with info on Slack & Jira.

PR #1101, opened 2024-12-02
DON-998: Pull Campaign.isRegularGiving and .regularGivingCollectionEnd fields from Salesforce

bdsl:


src/Domain/CampaignRepository.php
```diff
@@ -164,16 +233,6 @@ protected function doUpdateFromSf(SalesforceReadProxy $proxy, bool $withCache):
...
-        $charity = $this->pullCharity(
-            salesforceCharityId: $campaignData['charity']['id'],
-            charityName: $campaignData['charity']['name'],
-            stripeAccountId: $campaignData['charity']['stripeAccountId'],
-            giftAidOnboardingStatus: $campaignData['charity']['giftAidOnboardingStatus'],
-            hmrcReferenceNumber: $campaignData['charity']['hmrcReferenceNumber'],
-            regulator: $campaignData['charity']['regulatorRegion'],
-            regulatorNumber: $campaignData['charity']['regulatorNumber'],
-        );
-
```
bdsl, line 176: we know this is always going to update the existing charity that's attached to the campaign, not pull a new one.

dorota-joanna (approved):
LGTM

integrationTests/CampaignRepositoryTest.php
```diff
@@ -65,13 +67,15 @@ public function testItFindsNo10MonthOldCampaignEvenIfCharityAwaitingGiftAidAppro
         $campaign = new Campaign(
             Salesforce18Id::ofCampaign('xxxxxxxxxxxxxxxxxx'),
             $this->getCharityAwaitingGiftAidApproval(),
-            startDate: new \DateTimeImmutable('-9 months'),
+            startDate: new \DateTimeImmutable('-11 months'),
```
bdsl, line 70: added check in Campaign model to make sure a campaign can't start before it ends. I think that probably makes sense but there's an argument that any checks like that should be done in SF if needed and by the time the data gets to matchbot its too late and we should just accept whatever SF gives us, since we can't make the exception bubble up to the UI.

src/Domain/Campaign.php
```diff
@@ -91,19 +105,25 @@ public function __construct(
...
+            currencyCode: $currencyCode,
+            status: $status,
+            endDate: $endDate,
+            isMatched: $isMatched,
+            name: $name,
+            startDate: $startDate,
+            ready: $ready,
+            isRegularGiving: $isRegularGiving,
+            regularGivingCollectionEnd: $regularGivingCollectionEnd,
+        );
```
bdsl, line 126: delegating to updateFromSfPull as that now does some sanity checks on the data that are equally relevant at construction time. Also to help make sure that any future changes to fields pulled from SF get reflected both in new campaigns and updated campaigns.

src/Domain/Campaign.php
```diff
@@ -243,22 +260,43 @@ public function isReady(): bool
         return $this->ready;
     }
 
-    public function updateFromSfPull(
+    final public function updateFromSfPull(
```
bdsl, line 263: has to be final for Psalm to be satisfied that this can reliably take on some of the constructor's job of setting fields that don't have default values.

dorota-joanna, line 263: What's the reason for this change?

dorota-joanna (approved):
One question about _final_, otherwise LGTM

src/Domain/Campaign.php
bdsl: Without this change psalm would show these errors:

```
ERROR: PropertyNotSetInConstructor
at /home/barney/projects/matchbot/src/Domain/Campaign.php:77:18
Property MatchBot\Domain\Campaign::$ready is not defined in constructor of MatchBot\Domain\Campaign or in any private or final methods called in the constructor (see https://psalm.dev/074)
    private bool $ready;


ERROR: PropertyNotSetInConstructor
at /home/barney/projects/matchbot/src/Domain/Campaign.php:80:18
Property MatchBot\Domain\Campaign::$isRegularGiving is not defined in constructor of MatchBot\Domain\Campaign or in any private or final methods called in the constructor (see https://psalm.dev/074)
    private bool $isRegularGiving;


ERROR: PropertyNotSetInConstructor
at /home/barney/projects/matchbot/src/Domain/Campaign.php:91:35
Property MatchBot\Domain\Campaign::$regularGivingCollectionEnd is not defined in constructor of MatchBot\Domain\Campaign or in any private or final methods called in the constructor (see https://psalm.dev/074)
    protected ?\DateTimeImmutable $regularGivingCollectionEnd;
```

Because if the method is not final then in principle a subclass could override it, which would mean those properties would never get set, violating our expectation that they will always be set when the object is constructed.

PR #1104, opened 2024-12-04
Don't alarm on stripe "Invalid account" error for updated payment method

bdsl:
This can happen when a donor is updating their payment method at /my-account, as seen yesterday.
Not something matchbot can prevent, they would need to contact their bank.

We could improve the UX a bit by showing a friendly message but I don't think that's high priorty. This PR doesn't change the UX at all, only changes the log type from error to info

### DIFF
--- a/src/Application/Actions/UpdatePaymentMethod.php
+++ b/src/Application/Actions/UpdatePaymentMethod.php
@@ -21,6 +21,7 @@ class UpdatePaymentMethod extends Action
         'Your card has expired.',
         'Your card does not support this type of purchase.',
         'Your card\'s expiration month is invalid.',
+        'Invalid account',
     ];
 
     #[Pure]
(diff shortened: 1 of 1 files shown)


src/Application/Actions/UpdatePaymentMethod.php
```diff
@@ -21,6 +21,7 @@ class UpdatePaymentMethod extends Action
         'Your card has expired.',
         'Your card does not support this type of purchase.',
         'Your card\'s expiration month is invalid.',
+        'Invalid account',
```
NoelLH, line 24:
```suggestion
        'Invalid account.',
```

I think we need Stripe's punctuation as we're after an exact match upon comparison.

dorota-joanna (dismissed):
LGTM

src/Application/Actions/UpdatePaymentMethod.php
bdsl: You're right, thanks.

PR #1105, opened 2024-12-04
MAT-393 – add a possible extra fund release lock

NoelLH:


NoelLH:
Closing as very likely obsolete: https://github.com/thebiggive/matchbot/pull/1117 applies a simpler short term fix with similar reasoning and is in Production now; https://github.com/thebiggive/matchbot/pull/1119 and maybe a later PR with a similar strategy will hopefully give us an improved long-term fix later.

PR #1106, opened 2024-12-04
MAT-393 – patch a donation and CampaignFunding to fix a rare allocation blip

NoelLH:


src/Application/Commands/RedistributeMatchFunds.php
```diff
@@ -16,16 +18,48 @@
...
+    /**
+     * @throws \Symfony\Component\Lock\Exception\LockAcquiringException
+     */
+    private function ensureOneTimePatchNotDone(): void
+    {
+        $oneTimeFixLock = $this->lockFactory->createLock(
+            resource: 'redistribute-2024-12-04-patch',
+            ttl: 2 * 24 * 60 * 60, // 2 days
+        );
+        $oneTimeFixLock->acquire();
```
bdsl, line 63: shouldn't this be in blocking mode? Looks like we don't want to wait for this lock to be released, we just want to make sure we only ever acquire it once:
```suggestion
        $oneTimeFixLock->acquire(blocking: true);
```

bdsl (approved):
Added one suggestion, not a blocker, otherwise all LGTM.

src/Application/Commands/RedistributeMatchFunds.php
NoelLH: Hm, yep it [sounds like](https://symfony.com/doc/current/components/lock.html) either blocking mode or checking the return value is probably needed.
> By default, when a lock cannot be acquired, the acquire method returns false immediately.

I think I guessed wrong about the behaviour because of the `acquire()` doc block saying
> LockAcquiringException If the lock cannot be acquired

Seems like maybe that's only true for blocking mode.

Also, it looks like the Redis store doesn't implement `BlockingStoreInterface`, so this will just keep trying in a loop basically. Should work OK for what we need though.

PR #1115, opened 2024-12-05
DON-1009: WIP send emails for new regular giving mandates

bdsl:
WIP but should be fine to merge given that we know we need to address todo before launching the service or marking ticket done.

### DIFF
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -506,6 +506,7 @@ return function (ContainerBuilder $containerBuilder) {
         },
 
         ClockInterface::class => fn() => new NativeClock(),
+        Psr\Clock\ClockInterface::class  => fn() => new NativeClock(),
 
         EventDispatcherInterface::class => fn() => new EventDispatcher(),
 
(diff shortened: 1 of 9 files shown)


app/dependencies.php
```diff
@@ -502,6 +502,7 @@ static function (ContainerInterface $c): Matching\Adapter {
         },
 
         ClockInterface::class => fn() => new NativeClock(),
+        Psr\Clock\ClockInterface::class  => fn() => new NativeClock(),
```
bdsl, line 505: took me a while to work out why I was getting an error about ClockInterface not being instantiate despite line 504 already being there before I worked out it was because we only put the Symfony interface in here not the PSR interface. Arguably we should stick to one for consistency but it doesn't do any harm using the PSR one when we don't need the features from Symfony, and especially since they have the same name its a somewhat confusing error if we don't have the PSR one defined.

Also there's some inconsistency because in some services we inject a clock and in others we inject just a `$now` \DateTime.

NoelLH (approved):
Looks good so far, agree it's fine to merge knowing we're not using live before it's more ready

PR #1116, opened 2024-12-05
MAT-393 – revert everything used for temporary data patch which is done now

NoelLH:
(except semi-related minor refactoring)


bdsl (approved):
LGTM. Not urgently needed in prod so could go into the unfrozen branch, but it seems safe enough so I wouldn't object to breach code freeze for this either.

NoelLH:
> LGTM. Not urgently needed in prod so could go into the unfrozen branch, but it seems safe enough so I wouldn't object to breach code freeze for this either.

Thanks @bdsl, I'll stick with `develop` for now but won't do a Production deploy until we have something else that we want to get out alongside.

PR #1117, opened 2024-12-05
MAT-393 – hold the existing funds release lock longer, to fix a race condition

NoelLH:
Same goal as https://github.com/thebiggive/matchbot/pull/1105 but simpler and without introducing an extra lock.

Could do with testing, but probably better than the other option if I've not missed anything and it works.

### DIFF
--- a/src/Domain/DonationRepository.php
+++ b/src/Domain/DonationRepository.php
@@ -184,10 +184,6 @@ class DonationRepository extends SalesforceWriteProxyRepository
 
         $this->getEntityManager()->flush();
 
-        // After new funds are allocated, we want releasing them to be allowed without waiting for lock expiry, in case
-        // e.g. a donor changes their donation amount for a second time within that window.
-        $this->getFundsReleaseLock($donation)->release();
-
         return $amountNewlyMatched;
     }
 
--- a/src/Domain/DonationRepository.php
+++ b/src/Domain/DonationRepository.php
@@ -20,6 +20,7 @@ use MatchBot\Client\NotFoundException;
 use MatchBot\Domain\DomainException\MissingTransactionId;
 use Symfony\Component\Lock\Exception\LockAcquiringException;
 use Symfony\Component\Lock\LockFactory;
+use Symfony\Component\Lock\LockInterface;
 use Symfony\Component\Messenger\Envelope;
 use Symfony\Component\Messenger\MessageBusInterface;
 
@@ -183,6 +184,10 @@ class DonationRepository extends SalesforceWriteProxyRepository
 
         $this->getEntityManager()->flush();
 
+        // After new funds are allocated, we want releasing them to be allowed without waiting for lock expiry, in case
+        // e.g. a donor changes their donation amount for a second time within that window.
+        $this->getFundsReleaseLock($donation)->release();
+
         return $amountNewlyMatched;
     }
 
@@ -210,9 +215,12 @@ class DonationRepository extends SalesforceWriteProxyRepository
         // were able to observe. This is why we now go straight to trying to `acquire()`. If
         // this returns false, we are in the contested lock case and know to drop this attempt.
 
-        $fundsReleaseLock = $this->lockFactory->createLock("release-funds-{$donation->getUuid()}");
+        $fundsReleaseLock = $this->getFundsReleaseLock($donation);
 
         try {
+            // We no longer `release()` the lock, holding it for the default 5 mins instead unless a thread newly
+            // allocates funds during that time. This avoids race condition bugs between explicit cancel actions by
+            // donors and funds auto expiry.
             $gotLock = $fundsReleaseLock->acquire(false);
         } catch (LockAcquiringException $exception) {
             // According to the method (but not the exception) docs, `LockConflictedException` is thrown only
@@ -257,8 +265,6 @@ class DonationRepository extends SalesforceWriteProxyRepository
 
         $this->logInfo("Taking from ID {$donation->getUuid()} released match funds totalling {$totalAmountReleased}");
         $this->logInfo('Deallocation took ' . round($lockEndTime - $lockStartTime, 6) . ' seconds');
-
-        $fundsReleaseLock->release();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Domain/DonationRepository.php
```diff
@@ -210,9 +215,12 @@ public function releaseMatchFunds(Donation $donation): void
         // were able to observe. This is why we now go straight to trying to `acquire()`. If
         // this returns false, we are in the contested lock case and know to drop this attempt.
 
-        $fundsReleaseLock = $this->lockFactory->createLock("release-funds-{$donation->getUuid()}");
+        $fundsReleaseLock = $this->getFundsReleaseLock($donation);
```
bdsl, line 218: extracting line into function unchanged. :+1:

src/Domain/DonationRepository.php
```diff
@@ -183,6 +184,10 @@ public function allocateMatchFunds(Donation $donation, bool $mistrustOrmFundingC
 
         $this->getEntityManager()->flush();
 
+        // After new funds are allocated, we want releasing them to be allowed without waiting for lock expiry, in case
+        // e.g. a donor changes their donation amount for a second time within that window.
+        $this->getFundsReleaseLock($donation)->release();
```
bdsl, line 189: confused about why we seem to be acquiring the lock and releasing it on the same line - shouldn't we have a variable for a lock aquired earlier that we're now ready to release.

NoelLH: I was thinking about amount changes, which means the release and allocation are different API calls so the lock isn't in memory. However I also forgot that that means the amended version of the donation is a _new_ donation; I don't think we patch core amounts. That being the case, maybe this is redundant. I can't immediately think of another reason we'd need to release, allocate and release again within 5 minutes for the same donation.

bdsl: Yes we don't allow changes to donation amount on any given donation. https://github.com/thebiggive/matchbot/blob/bc124b3e2bbf6d8ef513fd05eb16aea06f7b37eb/src/Application/Actions/Donations/Update.php#L251

src/Domain/DonationRepository.php
```diff
@@ -210,9 +215,12 @@ public function releaseMatchFunds(Donation $donation): void
         // were able to observe. This is why we now go straight to trying to `acquire()`. If
         // this returns false, we are in the contested lock case and know to drop this attempt.
 
-        $fundsReleaseLock = $this->lockFactory->createLock("release-funds-{$donation->getUuid()}");
+        $fundsReleaseLock = $this->getFundsReleaseLock($donation);
 
         try {
+            // We no longer `release()` the lock, holding it for the default 5 mins instead unless a thread newly
```
bdsl, line 221: Minor but I think using "no longer" makes it more work to read the code as you're asking the reader to compare what we do now to what we did in a previous version of matchbot.

```suggestion
            // We don't `release()` the lock, holding it for the default 5 mins instead unless a thread newly
```

src/Domain/DonationRepository.php
```diff
@@ -210,9 +215,12 @@ public function releaseMatchFunds(Donation $donation): void
         // were able to observe. This is why we now go straight to trying to `acquire()`. If
         // this returns false, we are in the contested lock case and know to drop this attempt.
 
-        $fundsReleaseLock = $this->lockFactory->createLock("release-funds-{$donation->getUuid()}");
+        $fundsReleaseLock = $this->getFundsReleaseLock($donation);
 
         try {
+            // We don't `release()` the lock, holding it for the default 5 mins instead unless a thread newly
```
bdsl, line 221: where does 5 minutes come from? Maybe worth adding a reference to that in the comment.

src/Domain/DonationRepository.php
```diff
@@ -257,8 +265,6 @@ public function releaseMatchFunds(Donation $donation): void
 
         $this->logInfo("Taking from ID {$donation->getUuid()} released match funds totalling {$totalAmountReleased}");
         $this->logInfo('Deallocation took ' . round($lockEndTime - $lockStartTime, 6) . ' seconds');
-
-        $fundsReleaseLock->release();
```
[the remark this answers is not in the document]
NoelLH: Sorry, I should have copied more context to here.

I don't think we can deal with it that cleanly because two separate processes are involved. From my initial investigations on Slack:

> Right, so looks like a very unlucky situation where the natural expiry command and an explicit Cancel from the browser (e.g. from changing amount) led to the same release of match funds within 0.5 seconds.

> Kind of a tricky one to fully preempt because we almost certainly want to keep releasing funds immediately on Cancel. Even waiting a minute means someone changing amounts might be told match funds are unavailable, unnecessarily.
> The challenge comes because there is a decent chance that these steps in order lead to two processes wanting to release match funds:
> 1. regular cron reads donations ready to release funds and finds Collected donation X (database query and looping through takes up to a few seconds)
> 2. during the few seconds, donor cancels donation X
> 3. web process doesn’t know about cron so immediately releases funds, finishes its work and releases the lock
> 4. cron has the previous info in memory as it works through the loop and so also still wants to release funds, and is able to acquire a new lock and do so
>
> It seems like our best option is probably some kind of longer lasting lock. Certainly for when things happen this way around, maybe both ways around? If it’s the latter then it could be that we just ought to rely on the TTL and not release the lock explicitly after any release of match funds; perhaps making an exception if we need to newly allocate match funds to an existing donation a second time.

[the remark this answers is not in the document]
bdsl: @NoelLH Thanks. Lets have a call after standup. That helps but I'm still not fully clear.

It makes me think maybe what we need then is for the cron process to hold the lock for basically the entire time `MatchFundsRedistributor::redistributeMatchFunds` is running, from before it fetches data from the DB until its done. Or maybe better split that function up so that it queries for a list of donation IDs not full donation objects, then for each one acquires the lock, loads it from db, then releases the lock.

Also probably longer term thing but maybe instead of just sending a cancel command the client should send a command to replace one donation with another (effectively changing the amount on a donation) so that matchbot can do the entire thing atomically. That's what it looks like in the UI anyway.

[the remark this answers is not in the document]
bdsl: Also wondering if there's something we can do to stop using or de-emphasize the matchbot.CampaignFunding.amountAvailable column . As I understand the issue is overlapping processes to update this column that can mean it ends up with the wrong value. Maybe instead of storing the amountAvailable in MySQL we can store the full list of things that increase and decrease that amount, and calculate the current total on demand. As that list would only ever need to appended to it wouldn't need locking. If necessary we could also keep the amountAvailable as a performance optimisation but the full list would be the main source of truth.

Like my bank account has a balance but the source of truth is the list of transactions from which the balance can be calculated, not the balance itself.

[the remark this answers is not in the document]
bdsl: Clarified with @NoelLH , the thing we're waiting for before releasing the lock is any copy of the donation thats in memory on a different thread from before this lock was aquired needs to be released. As threads do not run for more than 5 minutes we can be fairly sure that that will happen if we just wait 5 minutes.

bdsl (approved):
LGTM for now - would still like to make more changes to not rely on the expire process taking less than 5 minutes.

PR #1118, opened 2024-12-06
MAT-394 – remove Gift Aid flag on a donation where it was claimed in error

NoelLH:


bdsl (approved):
LGTM, I double checked the uuid and transaction ID against the case and the prod DB.

PR #1119, opened 2024-12-06
MAT-393 – Use donation DB level lock for releasing funds instead…

bdsl:
…; temporary test updates, logging etc.


src/Domain/DonationRepository.php
```diff
@@ -257,18 +221,18 @@ public function releaseMatchFunds(Donation $donation): void
...
-     * @return Donation[]
+     * @return UuidInterface[]
      */
     public function findWithExpiredMatching(\DateTimeImmutable $now): array
     {
         $cutoff = $now->sub(new \DateInterval('PT' . self::EXPIRY_SECONDS . 'S'));
 
         $qb = $this->getEntityManager()->createQueryBuilder()
-            ->select('d')
+            ->select('d.uuid')
```
NoelLH, line 235: Nice

src/Domain/DonationRepository.php
```diff
@@ -283,9 +247,12 @@ public function findWithExpiredMatching(\DateTimeImmutable $now): array
         // `ExpireMatchFunds`, it makes more sense to opt it out of result caching
         // here rather than take the performance hit of a full query cache clear
         // after every single persisted donation.
-        return $qb->getQuery()
+        /** @var list<array{uuid: UuidInterface}> $rows */
+        $rows = $qb->getQuery()
             ->disableResultCache()
```
NoelLH, line 252: I think this existing line and the comment above are misleading, as I since learnt we're probably not using the result cache in general. But I'm in two minds about removing it because we have a ticket to probably change that soon-ish.

src/Domain/DonationService.php
```diff
@@ -488,6 +489,9 @@ function () use ($donation) {
     /**
      * Sets donation to cancelled in matchbot db, releases match funds, cancels payment in stripe, and updates
      * salesforce.
+     *
+     * Call this from inside a transaction and with a locked donation to avoid double releasing funds associated with
```
NoelLH, line 493: Nice, makes sense. Took me a moment to see why the cancel call via `MatchBot\Application\Actions\Donations\Update` is safe despite not using `releaseMatchFundsInTransaction()` but I think it is. Possibly worth a comment in `Update` to emphasise how they relate and the importance of not removing the `beginTransaction()` stuff?

tests/Application/Actions/Donations/UpdateTest.php
```diff
@@ -1923,4 +1884,27 @@ private function setDoublesInContainer(
         $container->set(DonorAccountRepository::class, $this->prophesize(DonorAccountRepository::class)->reveal());
         $container->set(ClockInterface::class, new MockClock());
     }
+
+    /**
+     * @return ObjectProphecy<RetrySafeEntityManager>
+     */
+    public function prophesizeEM(bool $persist = false, bool $flush = false, bool $commit = false): ObjectProphecy
```
NoelLH, line 1891: Nice, much cleaner

NoelLH (approved):
Looks good!

bdsl:
Testing this on my local:

1. I've created a pending donation for £10.
2. Verified in the database that the relevant CampaignFunding now shows amount 5,000 and amountAvailable 4990 as we should expect, since £10 is reserved for my pending donation.

3. Ran the ExpireMatchFunds command, modified to have an 60s sleep between fetching the list of expired donation UUIDs and releasing the funds, and also with the find function modified so it finds all recent donations instead of only slightly older ones.

4. `d ./matchbot matchbot:expire-match-funds`

5. During that 60s pause, modified the donation via the dona UI to change the amount from £10 to £1 (internally treated as cancelling a £10 donation and creating a new pending £1 donation). Verified that the amountAvailable  is now updated to  £4999 as expected.

6. When the 60 seconds is up found a TypeError crash due to uncommitted extra logging code I added, so will fix that and go back to start and try again.


----


Repeated steps 1-5  from above.

Saw output from command:

"Released 1 donations' matching"

But the amountAvailable correctly remains at £4999.00 . The "Released 1 donations' matching" message is slightly misleading, its the count of donations that appeared to need matching released when the command was started, and includes the 1 that turned out not to need anything released since it was released by step 5 instead.

src/Domain/DonationRepository.php
bdsl: Right, I think I'll merge as-is then, especially since this PR is already a big diff.

src/Domain/DonationService.php
bdsl: Right, that's safe because the donation is locked at https://github.com/thebiggive/matchbot/blob/b9ff400bc78ad48a23f9865cbc1bfff3863d0cd4/src/Application/Actions/Donations/Update.php#L118 I think.

I'll add a comment for now. I did also wonder about if we could have a `LockedDonation` type somehow. Not sure about that.

bdsl: Done in c30771bdc4684fe530ac

PR #1126, opened 2024-12-09
DON-1106 – map properties needed for Donation Create to save Gift Aid info

NoelLH:
Needed for Donation Funds "tips" to Big Give to have the correct info saved without an Update.

Targeting `develop` as I don't expect we would choose to merge this much before noon tomorrow anyway.

Feels like ideally we'd add tests too but I'm not sure what would be a valuable one as the gap we missed here sort of sits between apps. Maybe expanding the scope of (or adding) a regression test is the best option.

### DIFF
--- a/src/Domain/Donation.php
+++ b/src/Domain/Donation.php
@@ -172,7 +172,6 @@ class Donation extends SalesforceWriteProxy
 
     /**
      * Whether the Donor has asked for Gift Aid to be claimed about this donation.
-     * @var bool
      */
     #[ORM\Column()]
     protected bool $giftAid = false;
@@ -357,6 +356,10 @@ class Donation extends SalesforceWriteProxy
         ?string $tipAmount,
         ?RegularGivingMandate $mandate,
         ?DonationSequenceNumber $mandateSequenceNumber,
+        ?bool $giftAid = null,
+        ?bool $tipGiftAid = null,
+        ?string $homeAddress = null,
+        ?string $homePostcode = null,
     ) {
         $this->setUuid(Uuid::uuid4());
         $this->fundingWithdrawals = new ArrayCollection();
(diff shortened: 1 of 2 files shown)


src/Domain/Donation.php
```diff
@@ -357,6 +356,10 @@ public function __construct(
         ?string $tipAmount,
         ?RegularGivingMandate $mandate,
         ?DonationSequenceNumber $mandateSequenceNumber,
+        ?bool $giftAid = null,
```
bdsl, line 359: While the property is non-nullable I think it's clearer to make the param non-null too, and then we wouldn't need the false default set on line 395. Is there any advantage to having it default null here?

```suggestion
        bool $giftAid = false,
```

bdsl: Also not a blocker but I'm not sure if we should have a default at all, maybe better to not set a default here and pass the value from all callers to try and make sure we don't forget to consider it in any case we construct a donation. IntellJ can fill it in in all callers as part of the change signature refactoring command.

src/Domain/Donation.php
```diff
@@ -389,6 +392,10 @@ public function __construct(
         $this->setDonorName($donorName);
         $this->setDonorEmailAddress($emailAddress);
 
+        $this->giftAid = $giftAid ?? false;
```
bdsl, line 395: see https://github.com/thebiggive/matchbot/pull/1126/files#r1878287896

src/Domain/Donation.php
NoelLH: In the more common case of the main donation form, we currently start off intentionally `null` on donation create. I think it's quite useful to distinguish between `false` and `null` there, and I'm not sure that we're less likely to miss a similar bug to this one in the main form if we were to switch to initialising it `false`. It kind of feels more risky than what we do now to me.

But then I'm casting to `false` anyway so maybe this is a symptom of having been awake too long, not sure.

bdsl (approved):

PR #1128, opened 2024-12-10
DON-998: rename mandate donation amount

dorota-joanna:


bdsl (dismissed):

bdsl:
@dorota-joanna do you want to merge this?

bdsl:
ah looks like this has no changes now because it was duplicated by https://github.com/thebiggive/matchbot/pull/1138 so can be closed.

PR #1130, opened 2024-12-10
DON-998: get one mandate by uuid and return to FE

dorota-joanna:


src/Application/Actions/RegularGivingMandate/Get.php
```diff
@@ -0,0 +1,47 @@
...
+        parent::__construct($logger);
+    }
+
+    protected function action(Request $request, Response $response, array $args): Response
+    {
+        if (! $this->environment->isFeatureEnabledRegularGiving()) {
+            throw new HttpNotFoundException($request);
+        }
+        if (empty($args['mandateId'])) {
+            throw new DomainRecordNotFoundException('Missing donation ID');
```
bdsl, line 30: Liking the validation but I think its worth avoiding  `empty` in general. We could add a rule against empty to psalm config. See https://localheinz.com/articles/2023/05/10/avoiding-empty-in-php/

Probably fine not to do any particular check here actually, fine to rely on the throw when the regularGivingMandateRepository doesn't return anything.

src/Domain/RegularGivingMandateRepository.php
```diff
@@ -53,6 +54,14 @@ public function allActiveForDonorWithCharities(PersonId $donor): array
         return $this->getMandatesWithCharities($query);
     }
 
+    /**
+     * @param UuidInterface $uuid
+     * @return RegularGivingMandate|object|null
+     */
+    public function findOneBy(UuidInterface $uuid) {
```
bdsl, line 61: Liking the specific method for findOneBy UUID .

Couple of suggestions:
Rename to `findOneByUuid`. PHP doesn't have method overloading so if we want to findOnBy something else that method would need a different name.

Remove the docblock - we can really on native PHP types - we only need to write types types in docblocks when we want to go beyond the capabilities of the native PHP type system.

src/Application/Actions/RegularGivingMandate/Get.php
bdsl: not a blocker, may try and remove usage of empty more generally at some point.

src/Application/Actions/RegularGivingMandate/Get.php
```diff
@@ -0,0 +1,60 @@
...
+use MatchBot\Domain\PersonId;
+use Psr\Http\Message\ResponseInterface as Response;
+use Psr\Http\Message\ServerRequestInterface as Request;
+use Psr\Log\LoggerInterface;
+use Ramsey\Uuid\Uuid;
+use Slim\Exception\HttpNotFoundException;
+
+class Get extends Action
+{
+    #[Pure]
```
bdsl, line 24: Don't think it makes sense to say the constructor is pure.
```suggestion
```

bdsl (approved):
One linting error now but otherwise all LGTM, feel free to merge without asking me again when it fixed.

PR #1132, opened 2024-12-11
Remove clashing (in some situations?) 'l' option shortcut

NoelLH:


bdsl (approved):

bdsl:
Found that the 'l' shortcut I added to the app globally clashes with this one from the Symfony Consume Messages command [https://github.com/symfony/messenger/blob/7.2/Command/ConsumeMessagesCommand.php#L71 . ](https://github.com/symfony/messenger/blob/cc0e820c02a0a887a88ddb52b7c4de4634677ce6/Command/ConsumeMessagesCommand.php#L71) . Didn't spot it in dev env or CircleCI as we're not routinely running that command there.

PR #1133, opened 2024-12-11
Fix Messenger consumer command DI

NoelLH:
My local now gets as far as trying to read messages (and crashing because my local queue has the old stamp path). I think that's probably enough to be semi-confident that it will fix the real queues.

### DIFF
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -59,6 +59,7 @@ use Symfony\Component\Lock\Store\DoctrineDbalStore;
 use Symfony\Component\Messenger\Bridge\AmazonSqs\Middleware\AddFifoStampMiddleware;
 use Symfony\Component\Messenger\Bridge\AmazonSqs\Transport\AmazonSqsTransportFactory;
 use Symfony\Component\Messenger\Bridge\Redis\Transport\RedisTransportFactory;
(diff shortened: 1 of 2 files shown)


bdsl (approved):
Looks good, I see this is fixing what I broke in https://github.com/thebiggive/matchbot/commit/7c5901c20e4465b128c945b3692dfb2c1f7cfa62 when I removed

```
    new ConsumeMessagesCommand(
        $psr11App->get(RoutableMessageBus::class),
        $messengerReceiverLocator,
        new EventDispatcher(),
        $psr11App->get(LoggerInterface::class),
        [$messengerReceiverKey],
    ),
``` 

from the cli script. Apologies!

Not sure if/why the message receiver needs a different Container to everything else. Maybe worth adding a comment to explain if you know or to say if you're also unsure.

NoelLH:
> Not sure if/why the message receiver needs a different Container to everything else. Maybe worth adding a comment to explain if you know or to say if you're also unsure.

Not totally sure, been quite a while since I looked at it. I think that the param being called `$receiverLocator` and seeing other examples that seemed to construct a smaller container, I never really considered using the whole app's container. Not sure it's necessary that it's like that.

PR #1134, opened 2024-12-11
DON-1066: Show link to mandate from donation

bdsl:


src/Domain/DoctrineDonationRepository.php
bdsl, line 1: This file/class is basically renamed with only minor changes (and docs moved to the other file), but git isn't showing it as a rename. To see changes on CLI run `git diff origin/develop:src/Domain/DonationRepository.php HEAD:src/Domain/DoctrineDonationRepository.php`

src/Domain/Donation.php
```diff
@@ -562,6 +568,15 @@ public function toFrontEndApiModel(): array
             $data['matchReservedAmount'] = (float) $this->getFundingWithdrawalTotal();
         }
 
+        if ($this->mandate) {
+            // Not including the entire mandate details as that would be wasteful, just parts FE needs to display with
+            // the donation.
+            $data['mandate']['uuid'] = $this->mandate->getUuid()->toString();
```
bdsl, line 574: This will cause an extra db query to load each mandate. Not ideal and we could eagerly load them to stop that but probably not an issue for the current feature as one donor is never going likely to have hundreds of mandates.

Might be a problem for things that deal with donations more in bulk though in a future when we have hundreds or thousands of mandates in the database.

integrationTests/RedistributeMatchingCommandTest.php
```diff
@@ -247,4 +223,24 @@ private function prepareCampaign(bool $isOpen): array
 
         return [$campaign, $championFundCampaignFunding];
     }
+
+    public function runCommand(): BufferedOutput
```
NoelLH, line 227: Nice

src/Domain/DoctrineDonationRepository.php
NoelLH: Cool, makes sense I think. "the other file" here is the original `DonationRepository` which is now an interface which this implements.

tests/TestCase.php
```diff
@@ -4,6 +4,7 @@
 
 namespace MatchBot\Tests;
 
+use Cassandra\Uuid;
```
NoelLH, line 7: I think this is unintentional and importing the other Uuid would make the below more concise instead?

src/Application/Messenger/DonationUpserted.php
```diff
@@ -8,14 +8,11 @@
 
 class DonationUpserted extends AbstractStateChanged implements MessageGroupAwareInterface
 {
-    private function __construct(public string $uuid, public array $jsonSnapshot)
+    public function __construct(public string $uuid, public array $jsonSnapshot)
```
NoelLH, line 11: I see this has made a shared test helper 2 lines shorter – not totally sure that's a benefit worthwhile for losing a belt and braces guarantee that the message always comes from a real `Donation`, seems like a move away from value objects by default? But I may be missing other benefits I'm yet to get to or not understanding a wider goal?

tests/Application/Commands/RedistributeMatchFundsTest.php
```diff
@@ -5,6 +5,7 @@
 namespace MatchBot\Tests\Application\Commands;
 
 use Doctrine\ORM\EntityManagerInterface;
+use JmesPath\Env;
```
NoelLH, line 8:
```suggestion
```

tests/Application/Commands/RedistributeMatchFundsTest.php
```diff
@@ -15,6 +16,7 @@
 use MatchBot\Domain\ChampionFund;
 use MatchBot\Domain\Donation;
 use MatchBot\Domain\DonationRepository;
+use MatchBot\Domain\DonationService;
```
NoelLH, line 19:
```suggestion
```

tests/Application/Commands/PushDonationsTest.php
```diff
@@ -16,17 +20,13 @@ class PushDonationsTest extends TestCase
...
-        $bus = $this->prophesize(RoutableMessageBus::class)->reveal();
-        $now = new \DateTimeImmutable();
-        $donationRepoProphecy = $this->prophesize(DonationRepository::class);
-        $donationRepoProphecy->abandonOldCancelled()
-            ->willReturn(0)
-            ->shouldBeCalledOnce();
-        $donationRepoProphecy->pushSalesforcePending(now: $now, bus: $bus,)
-            ->willReturn(1)
-            ->shouldBeCalledOnce();
+        list($bus, $now, $donationRepoProphecy) = $this->getTestDoubles(numberCancelled: 0);
```
NoelLH, line 23: Nice

tests/Application/Actions/Hooks/StripePaymentsUpdateTest.php
```diff
@@ -142,12 +139,7 @@ public function testSuccessfulPayment(): void
         $body = json_encode($webhookContent);
         $webhookSecret = $this->getValidWebhookSecret($container);
         $time = (string) time();
-
-        $donationRepoProphecy = $this->prophesize(DonationRepository::class);
-        $donationRepoProphecy
-            ->findAndLockOneBy(['transactionId' => 'pi_externalId_123'])
-            ->willReturn($donation)
-            ->shouldBeCalledOnce();
+        $this->donationRepository->store($donation);
```
NoelLH, line 142: Liking this. Seems like right now, this is the main win from the extra layer of abstraction having the repo as an interface, but the expectation is there will be a longer term gain as we implement more mocked methods in the in-memory version for more tests?

NoelLH (approved):
Looks good, minor things inline, no blockers

tests/TestCase.php
bdsl: Yes, will change.

src/Application/Messenger/DonationUpserted.php
bdsl: This change is left over by accident from when I thought Donations couldn't fetch mandates. I will undo.

tests/Application/Actions/Hooks/StripePaymentsUpdateTest.php
bdsl: Yes I think in many tests it will be useful to have an in memory repo and using that should make it a lot easier to change the repo api since the many tests don't each constrain it directly.

But it is still a bit of an awkward design since there are still many non persistence related functions in the repo.

tests/TestCase.php
bdsl: Done in 0f5768e8

src/Application/Messenger/DonationUpserted.php
bdsl: Undone in 0f5768e8

tests/Application/Commands/RedistributeMatchFundsTest.php
bdsl: Deleted as part of 0f5768e8

NoelLH (approved):

PR #1135, opened 2024-12-12
DON-1009: Fetch campaign thank you messages from SF and include in re…

dorota-joanna:
…gular giving notification email

Also dropped other fields from regular giving notification email that we don't have and think we may not need.

### DIFF
--- a/src/Domain/CampaignRepository.php
+++ b/src/Domain/CampaignRepository.php
@@ -251,17 +251,17 @@ class CampaignRepository extends SalesforceReadProxyRepository
 
 
         $campaign->updateFromSfPull(
-            status: $campaignData['status'],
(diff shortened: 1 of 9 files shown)


src/Domain/CampaignRepository.php
```diff
@@ -109,7 +109,6 @@ public function pullNewFromSf(Salesforce18Id $salesforceId): Campaign
         $this->getEntityManager()->flush();
 
         $this->logInfo('Done persisting new campiagn ' . $campaign->getSalesforceId());
```
dorota-joanna, line 111:
```suggestion
        $this->logInfo('Done persisting new campiagn ' . $campaign->getSalesforceId());
        
```

dorota-joanna (commented):
LGTM - continued with @bdsl

PR #1136, opened 2024-12-12
DON-1009: Start adding donation properties to mandate email request t…

bdsl:
…o mailer


src/Domain/RegularGivingNotifier.php
```diff
@@ -36,22 +40,58 @@ public function notifyNewMandateCreated(
...
+            'donationAmount' => $firstDonation->getAmount(),
+            'donationDatetime' => $firstDonationCollectedAt?->format('c'),
+            'charityName' => $charity->getName(),
+            'transactionId' => $firstDonation->getTransactionId(),
+            'matchedAmount' => $firstDonation->getFundingWithdrawalTotal(),
+            'statementReference' => $campaign->getCharity()->getStatementDescriptor(),
+
+            // @todo-regular-giving - implement giftAidAmountClaimed. I think we want to store
+            // this in the DB, not rely on being able to calculate in demand, to allow for the
+            // tax rate to change.
```
bdsl, line 85: giftAidAmountClaimed is used in mailer's existing template for non-regular giving emails here:

https://github.com/thebiggive/mailer/blob/75ce26214ad719f794b44c29cf4b910805c6bdd5/templates/donor-donation-success.html.twig#L37 . 

I think the past tense 'claimed' is maybe a misnomer, or it means claimed by the donor rather than claimed already from HMRC.

tests/Domain/RegularGivingNotifierTest.php
```diff
@@ -2,39 +2,145 @@
...
-        $campaign = TestCase::someCampaign(thankYouMessage: 'Thank you for setting up your regular donation to us!');
-        $personId = PersonId::of(Uuid::uuid4()->toString());
+                ]
+            ]
+        ]);
+
+        $this->whenWeNotifyThemThatTheMandateWasCreated($clock, $mandate, $donor, $campaign, $firstDonation);
+    }
+
+    public function markDonationCollected(Donation $firstDonation, \DateTimeImmutable $collectionDate): void
```
NoelLH, line 86: I think a few supporting fns could be `private` to help see at a glance which are tests vs. test support?

NoelLH (approved):
Looks good, no blockers

NoelLH (approved):
Latest commit also LGTM

Re
> Wondering if we should store total of withdrawals in the DB instead of
always summing at the time. Particularly when we're looking at an old
donation it feels like it would be more natural to be looking at a
record of something calculated on the donation day rather than doing a
sum at that point.

there are quite a lot of situations where this changes after initial allocation. I'm not sure if the proposal would be to denormalise once on first allocation or try to keep it in sync? It feels like once we need a lot of detail, the logs have it already and storing it in the database in a useful way for any investigative scenario may be tricky.

PR #1137, opened 2024-12-13
DON-998 - add more info to mandate not found exception message

dorota-joanna:


src/Application/Actions/RegularGivingMandate/Get.php
```diff
@@ -36,17 +37,22 @@ protected function action(Request $request, Response $response, array $args): Re
...
+
+        Assertion::keyExists($args, "mandateId");
+        $mandateId = $args["mandateId"];
+        Assertion::string($mandateId);
         if (empty($args['mandateId'])) {
-            throw new DomainRecordNotFoundException('Missing donation ID');
+            throw new DomainRecordNotFoundException('Missing mandate ID ' . $mandateId);
         }
 
+        // @todo-regular-giving: make sure that the donor requesting the mandate is the owner of the mandate
```
bdsl, line 48: I think since this is a confidentiality requirement we should do this before we merge this PR to make sure it can't be forgotten.

src/Domain/RegularGivingMandate.php
```diff
@@ -108,7 +108,9 @@ public function activate(\DateTimeImmutable $activationDate): void
 
     public function toFrontEndApiModel(Charity $charity, \DateTimeImmutable $now): array
     {
-        Assertion::same($charity->salesforceId, $this->charityId);
+        //var_dump(['charity' => $charity->salesforceId, 'charityId' => $this->charityId]);
+        // the sf id is not set on the proxy object for the charity, so we need to get from DB
```
bdsl, line 112:
```suggestion
```

bdsl (commented):
Added comments inline, everything else LGTM

src/Application/Actions/RegularGivingMandate/Get.php
```diff
@@ -36,19 +36,24 @@ protected function action(Request $request, Response $response, array $args): Re
...
         $donorId = $request->getAttribute(PersonWithPasswordAuthMiddleware::PERSON_ID_ATTRIBUTE_NAME);
         \assert($donorId instanceof PersonId);
         $uuid = Uuid::fromString((string) $args['mandateId']);
         $mandate = $this->regularGivingMandateRepository->findOneByUuid($uuid);
-
```
dorota-joanna, line 47: Cannot throw an exception here because sa:check says $mandate is never falsy, but we are throwing exception up a level in the `regularGivingMandateRepository->findOneByUuid($uuid)` when the mandate wasn't found

bdsl: @dorota-joanna that's probably because you've asserted that the mandate is not null - if you assert it Psalm will believe you.

src/Application/Actions/RegularGivingMandate/Get.php
```diff
@@ -36,19 +36,24 @@ protected function action(Request $request, Response $response, array $args): Re
...
-        if (!$mandate) {
-            throw new DomainRecordNotFoundException('Mandate not found');
+        \assert($mandate !== null);
+        \assert($donorId === $mandate->donorId);
+        if ($donorId !== $mandate->donorId) {
+            throw new DomainRecordNotFoundException('Donor Id: ' . strval($donorId) . ' on request does not match donor Id on the mandate: ' . strval($mandate->donorId));
```
bdsl, line 55: strval is a sort of aggressive way to make sure you've got a string. It's better to work with the object by doing `$donorId->id`

bdsl (approved):

PR #1138, opened 2024-12-17
Don 998 make api fields match frontend

dorota-joanna:


src/Domain/RegularGivingMandate.php
```diff
@@ -247,7 +248,7 @@ public function getUuid(): UuidInterface
 
     public function getAmount(): Money
```
bdsl, line 249: worth also renaming this function. PHPStorm should be able to rename it and update all usages at once.

dorota-joanna: I tried, but it's overridden method

bdsl (approved):
LGTM. I'd just say on the commit message for c465621 "DON-998 - fix integration test" - that change isn't just to fix the integration test, it's an essential part of the rename. Matchbot would be broken and unable to work with mandates if we had merged this without that, so the integration tests did their job of alerting us to it being necessary.

PR #1141, opened 2024-12-18
DON-1009: Fix error in \MatchBot\Domain\Money::fromNumericStringGBP

bdsl:
This was randomly throwing an assertion exception in a test when given the input "258.40" because multiplying by 100 using floating point gives 25,839.9999999999963, not exactly 25,840.

This would have caused a real bug, more seriously if the assertion was not there, since `(int)(100 * "258.40")` returns 25839 not the 25840 you'd expect.

### DIFF
--- a/src/Domain/Money.php
+++ b/src/Domain/Money.php
@@ -115,9 +115,10 @@ readonly class Money implements \JsonSerializable, \Stringable
      */
     public static function fromNumericStringGBP(string $amount): self
     {
-        $amountInPence = $amount * 100;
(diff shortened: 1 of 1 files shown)


NoelLH (approved):

NoelLH:
@bdsl Approved but then thought, could `fromPoundsGBP()` have the same issue?

bdsl:
> `fromPoundsGBP`

No, because that takes an int not a string. And the only caller to it that's really for production passes `0`.

image

image

PR #1144, opened 2024-12-18
MAT-397 – reduce avoidable database work when considering allocating match funds

NoelLH:
Should improve performance in various places but most noticeably when running retrospective matching at the end of a big campaign.


src/Domain/DoctrineDonationRepository.php
```diff
@@ -156,11 +155,13 @@ public function allocateMatchFunds(Donation $donation): string
             $amountNewlyMatched = bcadd($amountNewlyMatched, $newWithdrawalAmount, 2);
         }
 
+        if (count($newWithdrawals) > 0) {
+            $this->getEntityManager()->flush();
+        }
+
```
bdsl, line 161: I'm not sure if this will give us any appreciable speedup, since `\Doctrine\ORM\UnitOfWork::commit` already looks like it returns without doing any DB work if there's nothing to update:

```
        if (
            ! ($this->entityInsertions ||
                $this->entityDeletions ||
                $this->entityUpdates ||
                $this->collectionUpdates ||
                $this->collectionDeletions ||
                $this->orphanRemovals)
        ) {
            $this->dispatchOnFlushEvent();
            $this->dispatchPostFlushEvent();

            $this->postCommitCleanup($entity);

            return; // Nothing to do.
        }
```

NoelLH: @bdsl That sounds very likely right. It's avoiding the explicit, extraneous transaction in `Adapter::saveFundingsToDatabase()` which I expect to make the main difference.

src/Domain/DoctrineDonationRepository.php
```diff
@@ -566,9 +547,11 @@ private function safelyAllocateFunds(Donation $donation, array $fundings, string
             $currentFundingIndex++;
         }
 
-        $this->queueForPersist($donation);
+        if ($withdrewAnything) {
+            $this->queueForPersist($donation);
+            $this->matchingAdapter->saveFundingsToDatabase();
```
bdsl, line 552: similarly I'm not sure `saveFundingsToDatabase` is ever doing anything useful. It only does `persist` but the CampaignFundings are not new at this point so surely they're already persisted? We only construct a new CampaignFunding object in `FundRepository::pullForCampaign`

bdsl: (to confirm we don't appear to use the [Deferred Explicit](https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/reference/change-tracking-policies.html#deferred-explicit) strategy anywhere - we use the default Deferred Implicit. 

Which means we don't have to ever call `persist` on an already doctrine-managed entity - its only needed for newly instantiated entities.

src/Domain/DoctrineDonationRepository.php
bdsl: OK - IMHO unless there's a reason you think making this flush conditional might speed things up or you want to test how it performs then I think we should keep it unconditional for simplicity.

src/Domain/DoctrineDonationRepository.php
```diff
@@ -566,9 +547,11 @@ private function safelyAllocateFunds(Donation $donation, array $fundings, string
             $currentFundingIndex++;
         }
 
-        $this->queueForPersist($donation);
+        if ($withdrewAnything) {
+            $this->queueForPersist($donation);
```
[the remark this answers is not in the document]
NoelLH: Perhaps not, I'll do some more testing later.

src/Domain/DoctrineDonationRepository.php
NoelLH: Yeah this does sound right – it would be great news if so as it's a commonly used and currently expensive fn. I'll look more closely later and double check that everything calling it has a flush after, as the current approach looks like it might well be a side effect of me misunderstanding this a while ago.

bdsl: Right. Afaik calling persist on an object that's doctrine-managed is useless whether or not there's a `flush` anywhere. Calling flush is the only bit that matters.

bdsl: If `Matching\Adapter::saveFundingsToDatabase` is useless as it seems to be then `Matching\Adapter::addAmountWithoutSavingFundingsToDB` is misleadingly named - it's always without saving to DB until we do a 'flush'.

src/Application/Matching/Adapter.php
```diff
@@ -56,14 +52,14 @@ public function __construct(
...
-        return $fundBalance;
+        return $this->addAmountWithoutSavingFundingsToDB($funding, $amount);
     }
 
-
     /**
+     * Note that we use Doctrine in deferred implicit mode – so the name of this fn is meant as a
+     * reminder to flush work overall. Specific per-entity action isn't needed if Doctrine already
+     * knows the `$funding`.
+     *
```
bdsl, line 62: IMHO we'd be better to rename this, or better inline it into `Matching\Adapter::addAmount`. 

Deferred implicit is the doctrine default, I don't see a reason to mention it just here. There's nothing special about this function not calling 'flush', we would put `WithoutSavingToDB` all over our codebase if we wanted to mark all functions that don't flush.

src/Application/Matching/Adapter.php
```diff
@@ -56,14 +52,14 @@ public function __construct(
      */
     public function addAmount(CampaignFunding $funding, string $amount): string
     {
-        $fundBalance = $this->addAmountWithoutSavingFundingsToDB($funding, $amount);
-        $this->saveFundingsToDatabase();
-
-        return $fundBalance;
+        return $this->addAmountWithoutSavingFundingsToDB($funding, $amount);
```
bdsl, line 55: IMHO we'd be better to rename this, or better inline it into `Matching\Adapter::addAmount`. 

Deferred implicit is the doctrine default, I don't see a reason to mention it just here. There's nothing special about this function not calling 'flush', we would put `WithoutSavingToDB` all over our codebase if we wanted to mark all functions that don't flush.

bdsl: Particularly since `addAmount` also doesn't do its own flush.

src/Domain/DoctrineDonationRepository.php
NoelLH: Yeah, I was thinking about inlining as the next step. It looks like this got renamed from addAmountWithoutNewTransaction() https://github.com/thebiggive/matchbot/commit/ff2a8d7d025771afe3d14fd6c6644288bf6fbd83 and maybe the goal was to emphasise that we should flush too? But I don't think either name really makes this clearer or is likely to be preventing bugs.

bdsl: I think the confusion has been in the code for a long time - at least since 2019 when we were calling `runTransactionally` but passing it a a closure that afaik didn't actually do an database work:

https://github.com/thebiggive/matchbot/blob/733d1db09f872bebd2cea29c22442ad3c6d4667b/src/Domain/DonationRepository.php#L115

src/Domain/DoctrineDonationRepository.php
NoelLH: I think which is clearer is debatable but reverted + added a comment to help understand which entities are involved when quickly scanning the code, in bc8cfdb042530c06e4da5a5f8df5ea1b6867ff58

src/Domain/DoctrineDonationRepository.php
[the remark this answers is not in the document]
NoelLH: I think a couple of places weren't yet flushing and should have been, hopefully all addressed in simplifications on 7633343037a55ad2e477274878ab4860d2e71e49

src/Application/Matching/Adapter.php
NoelLH: Done, more on overlapping thread at https://github.com/thebiggive/matchbot/pull/1144#pullrequestreview-2514010257

src/Domain/DoctrineDonationRepository.php
NoelLH: @bdsl I think back then we were trying out Redis for the first time and still had Doctrine as an option for the primary matching adapter. So probably the transaction made sense, kind of, depending on which one was in service. Then we kept building on it and forgot to simplify when it became Redis-only.

src/Application/Matching/Adapter.php
```diff
@@ -235,29 +222,12 @@ private function buildKey(CampaignFunding $funding): string
...
-            foreach ($this->fundingsToPersist as $funding) {
-                $this->entityManager->persist($funding);
-            }
-        });
-    }
-
     /**
      * @psalm-param numeric-string $newValue
      */
     private function setFundingValue(CampaignFunding $funding, string $newValue): void
```
bdsl, line 228: `setFundingValue` looks like it might as well be inlined to its callers now. The name isn't any more meaningful that `setAmountAvailable` is it?

src/Application/Matching/Adapter.php
```diff
@@ -46,29 +43,19 @@ class Adapter
...
-    /**
+     * Callers are responsible for flushing after this works on `CampaignFunding`s – typically once per larger operation
+     * if looping over many.
+     *
      * @param CampaignFunding $funding
      * @param numeric-string $amount
      * @return numeric-string New fund balance
      */
-    private function addAmountWithoutSavingFundingsToDB(CampaignFunding $funding, string $amount): string
+    public function addAmount(CampaignFunding $funding, string $amount): string
```
bdsl, line 58: Could we also rename `subtractAmountWithoutSavingToDB` to just `subtractAmount` for symmetry?

bdsl:
Looks like this is going to be a big performance boost and a big code ownership cost reduction at the same time!

image

src/Application/Commands/HandleOutOfSyncFunds.php
```diff
@@ -137,6 +139,8 @@ protected function doExecute(InputInterface $input, OutputInterface $output): in
             $this->outOfSyncFundFound = true;
         }
 
+        $this->entityManager->flush();
```
bdsl, line 142: is there a reason for adding this flush? Doesn't look like it will do any harm to performance or anything else but would be good to know. Afaik we're not removing a flush anywhere in this PR that would mean we need to add this. Think it's OK to add if it's just on a belt-and-braces principle.

src/Domain/DoctrineDonationRepository.php
```diff
@@ -169,6 +164,7 @@ public function releaseMatchFunds(Donation $donation): void
         $startTime = microtime(true);
         try {
             $totalAmountReleased = $this->matchingAdapter->releaseAllFundsForDonation($donation);
+            $this->getEntityManager()->flush();
```
bdsl, line 167: Similar to https://github.com/thebiggive/matchbot/pull/1144/files#r1893745565 what's the reason for adding this?

Looks like there are three callers to this in prod. Two already unconditionally call `flush` (assuming we don't throw / close the entity manager inside `removeAllFundingWithdrawalsForDonation`, but `\MatchBot\Domain\DonationService::cancel` doesn't flush if the call to Stripe throws an exception.

Wonder if we actually want to flush in case of an unrecoverable DBAl exception later though - might it not be better to leave that release unflushed so the funds don't get out of sync with the funding withdrawals?

src/Domain/DonationService.php
```diff
@@ -345,6 +345,8 @@ function () use ($donation) {
                         // would be reduced again when the donation expires.
                         $this->donationRepository->removeAllFundingWithdrawalsForDonation($donation);
 
+                        $this->entityManager->flush();
```
bdsl, line 348: looks like this might be worth trying but but in most cases if an exception has been thrown during allocateMatchFunds it may mean the EM is closed and we won't be able to flush anyway. Does that seem right to you @NoelLH ?

tests/Application/Matching/AdapterTest.php
bdsl, line 1: liking the amount of simplification in this file

src/Application/Commands/HandleOutOfSyncFunds.php
NoelLH: Having added the comment that's now on `Adapter::addAmount()`, I tried to trace back its direct callers and their callers to check that everything did in fact flush at some point. This was one place I couldn't find it doing so. And it's been so long since we ran an out of sync funds _correction_ that it seems quite possible we could have broken it at some point – so it seemed safer to add a flush to be sure persistence happens where we intended.

src/Application/Matching/Adapter.php
NoelLH: Yep! Done in c55d239727d83a90c1b0bd9b6472a07f482dc6ae

bdsl (approved):
Looks like this should be a big improvement in performance!

src/Application/Matching/Adapter.php
NoelLH: Nope, good spot. Done in ae8a1b145b7d69e787e9fb66be5522a02afb32d8

src/Domain/DoctrineDonationRepository.php
NoelLH: I think that in the case of Stripe API errors we are better to go ahead and release match funds so the impact doesn't necessarily extend to potential new donors. If we're looking in the same place (the latter half of `DonationService::cancel()`), then I think ignoring the Stripe error but making other things self-consistent is probably the safest thing we can do. We should not be in a position to incorrectly try again because MatchBot will refuse to process a cancelled donation on later calls based on its own database.

It's true that adding the flush here might not have been the most readable way to achieve that, not sure. I'm fairly ambivalent between keeping this line vs. adding a flush in `DonationService::cancel()` directly instead.

src/Domain/DonationService.php
NoelLH: Yes, certainly sounds possible. There's a chance the EM retry methods work to sort this out in some cases, but I don't think anything has hit this case recently enough for me to be confident either way.

src/Domain/DoctrineDonationRepository.php
NoelLH: I'll merge with this as-is for now – definitely open to moving this line if you feel the other option is a lot better.

PR #1148, opened 2024-12-20
Don 998 tweaks

bdsl:


src/Application/Commands/TakeRegularGivingDonations.php
```diff
@@ -108,7 +108,6 @@ private function confirmPreCreatedDonationsThatHaveReachedPaymentDate(
         /* @todo-regular-giving
             Still to do to improve this before launch:
             - Record unsuccessful payment attempts to limit number or time extent of retries
-            - Send metadata to stripe so to identify the payment as regular giving when we view it there.
```
bdsl, line 111: looks like this was put into the payment intent metadata a while ago and this comment wasn't updated.

NoelLH (approved):

PR #1152, opened 2024-12-27
MAT-379 – handle Person upsert messages from Identity

NoelLH:


src/Application/Messenger/Handler/PersonHandler.php
```diff
@@ -0,0 +1,57 @@
...
+ */
+#[AsMessageHandler]
+readonly class PersonHandler
+{
+    /** @psalm-suppress PossiblyUnusedMethod Used by Messenger mapping & tests. */
+    public function __construct(
+        // apparently at the time this is constructed in tests the container isn't ready to give
+        // us a donorAccountRepository, so taking a ref to the container instead and getting the
+        // repository inside __invoke
+        private ContainerInterface $container,
```
NoelLH, line 25: Strategy & comment borrowed from charity update counterpart. I found that when I injected the repo itself, actual runtime worked but tests hit strange errors from repo's not being instantiable in DI and ending up null.

dorota-joanna (approved):
LGTM

PR #1151, opened 2024-12-27
DON-998: set PI just before processing the payment instead of at the

dorota-joanna:
time of creating the donation

PR #1157, opened 2025-01-07
DON-998: Stop assuming donations will always have transaction IDs

bdsl:
This throw causes crashes in the regular giving form, and the assumption that almost every donation will have a non-null transaction ID stops being true when we launch regular giving, as for the 2nd and 3rd donations of each mandate there will be 1 and 2 month delays between donation creation and transaction id creation.


src/Domain/Donation.php
```diff
@@ -1688,4 +1665,27 @@ public function totalCharityValueAmount(): string
             ])
         )->toNumericString();
     }
+
+    /**
+     * These assertions are required both for a donation to be confirmed and for it to be pre-authorized. They do
+     * NOT include an assertion that the donation has a transaction ID, as that is only required for confirmation, not
+     * for pre-auth.
+     */
+    private function assertionsForConfirmOrPreAuth(): \Assert\LazyAssertion
```
NoelLH, line 1674: Liking that these end up very clear in terms of what they allow, and DRY

NoelLH (dismissed):
Nice simplifications along with the fix, looks good!

bdsl:
> Nice simplifications along with the fix, looks good!

Thanks! I think I'll wait until after https://github.com/thebiggive/matchbot/pull/1151 is merged before merging this to avoid creating merge conflicts over there.

PR #1160, opened 2025-01-08
DON-998: Confirm first donation in regular giving & save payment …

bdsl:
…method ID to account

### DIFF
--- a/integrationTests/CreateRegularGivingMandateTest.php
+++ b/integrationTests/CreateRegularGivingMandateTest.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace MatchBot\IntegrationTests;
 
+use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
 use GuzzleHttp\Psr7\ServerRequest;
 use MatchBot\Client\Mailer;
 use MatchBot\Domain\DonorAccountRepository;
--- a/integrationTests/CreateRegularGivingMandateTest.php
+++ b/integrationTests/CreateRegularGivingMandateTest.php
@@ -31,13 +31,13 @@ class CreateRegularGivingMandateTest extends IntegrationTest
         $stripeProphecy->createPaymentIntent(
             Argument::that(fn(array $payload) => ($payload['amount'] === $pencePerMonth))
         )
-            ->shouldBeCalledTimes(1)
+            ->shouldBeCalledOnce()
             ->will(fn() => new PaymentIntent('payment-intent-id-' . IntegrationTest::randomString()));
         $stripeProphecy->confirmPaymentIntent(
             Argument::type('string'),
             Argument::cetera()
         )
-            ->shouldBeCalledTimes(1)
+            ->shouldBeCalledOnce()
             ->will(fn(array $args) => new PaymentIntent($args[0]));
 
         $this->getContainer()->set(Stripe::class, $stripeProphecy->reveal());
--- a/integrationTests/CreateRegularGivingMandateTest.php
+++ b/integrationTests/CreateRegularGivingMandateTest.php
@@ -31,8 +31,15 @@ class CreateRegularGivingMandateTest extends IntegrationTest
         $stripeProphecy->createPaymentIntent(
             Argument::that(fn(array $payload) => ($payload['amount'] === $pencePerMonth))
         )
-            ->shouldBeCalledTimes(3)
+            ->shouldBeCalledTimes(1)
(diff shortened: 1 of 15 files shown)


src/Domain/RegularGivingNotifier.php
```diff
@@ -32,6 +32,7 @@ public function notifyNewMandateCreated(
                 'templateKey' => 'donor-mandate-confirmation',
                 'recipientEmailAddress' => $donorAccount->emailAddress->email,
                 'params' => [
+                    'donorName' => $donorAccount->donorName->fullName(),
```
bdsl, line 35: not related to other changes but necessary bug fix to be able to send this email.

integrationTests/CreateRegularGivingMandateTest.php
```diff
@@ -29,8 +31,15 @@ public function testItCreatesRegularGivingMandate(): void
         $stripeProphecy->createPaymentIntent(
             Argument::that(fn(array $payload) => ($payload['amount'] === $pencePerMonth))
         )
-            ->shouldBeCalledTimes(3)
+            ->shouldBeCalledTimes(1)
```
NoelLH, line 34:
```suggestion
            ->shouldBeCalledOnce()
```
Marginally clearer/ more searchable maybe? Similar in one other place below.

src/Domain/DonationService.php
```diff
@@ -650,4 +641,18 @@ public function findAllCompleteForCustomerAsAPIModels(StripeCustomerId $stripeCu
...
+
+    public function confirmDonationWithSavedPaymentMethod(Donation $donation, StripePaymentMethodId $paymentMethod): void
+    {
+        $paymentIntentId = $donation->getTransactionId();
+        Assertion::notNull($paymentIntentId);
+        $paymentIntent = $this->stripe->confirmPaymentIntent(
+            $paymentIntentId,
+            ['payment_method' => $paymentMethod->stripePaymentMethodId]
+        );
+        if ($paymentIntent->status !== 'succeeded') {
```
NoelLH, line 653: Maybe marginally more searchable / statically analyse-able to use `PaymentIntent::STATUS_SUCCEEDED`, but we're already not consistent about it. Definitely not a blocker.

src/Domain/DonationService.php
```diff
@@ -650,4 +641,18 @@ public function findAllCompleteForCustomerAsAPIModels(StripeCustomerId $stripeCu
...
+            // @todo-regular-giving: create a new db field on Donation - e.g. payment_attempt_count and update here
+            // decide on a limit and log an error (or warning) if exceeded
```
NoelLH, line 655: Yep – and maybe we want to auto cancel when the limit is reached? Although I guess releasing spare match funds isn't a high priority so long as the time windows for joining Regular Giving remain shorter than 1 month.

src/Domain/DonationService.php
```diff
@@ -345,7 +336,7 @@ function () use ($donation) {
             );
         }
 
-        if ($donation->getPsp() === 'stripe') {
+        if ($donation->getPsp() === 'stripe' && $donation->getDonationStatus() === DonationStatus::Pending) {
```
NoelLH, line 339:
```suggestion
        // Regular Giving enrolls donations with `DonationStatus::PreAuthorized`, which get Payment Intents later instead.
        if ($donation->getPsp() === 'stripe' && $donation->getDonationStatus() === DonationStatus::Pending) {
```
Maybe just because of the context shown by default in GitHub but it took me a minute to confirm the rationale for this. Maybe worth a comment for future glances at this area?

NoelLH (approved):
LGTM, no blockers

dorota-joanna (dismissed):
LGTM

src/Domain/DonationService.php
bdsl: Thanks! Suggestion applied.

src/Domain/DonationService.php
bdsl: Right. This comment isn't new it's just moved, but I've expanded it based on this in 856079fb13457abbf

src/Domain/DonationService.php
bdsl: Thanks - this is just moved code not new, but this seems like a good idea, done in e1ee63317670e33

integrationTests/CreateRegularGivingMandateTest.php
bdsl: Thanks, done in 59de6847fe9

PR #1161, opened 2025-01-08
BG2-2798 – push fund used amounts to Salesforce upon campaign close

NoelLH:


src/Domain/Fund.php
```diff
@@ -75,4 +89,54 @@ public function getCurrencyCode(): string
...
+
+            $totalAmount = $totalAmount->plus($thisAmount);
+            $usedAmount = $usedAmount->plus($thisAmountUsed);
+        }
+
+        return [
+            'totalAmount' => $totalAmount->toNumericString(),
+            'usedAmount' => $usedAmount->toNumericString(),
+            'currencyCode' => $this->currencyCode,
+        ];
```
bdsl, line 117: IMHO better to delay calling toNumericString until as late as possible, so we have as little as possible stringly-typed. In this case that would mean calling it inside `toAmountUsedUpdateModel` not here. I think returning the money objects would make the test clearer and reduce the chance of mistakes in future calls to `getAmounts`

src/Domain/Fund.php
```diff
@@ -75,4 +89,54 @@ public function getCurrencyCode(): string
...
+     * after a campaign closes.
+     *
+     * @return array{totalAmount: numeric-string, usedAmount: numeric-string, currencyCode: string}
+     */
+    public function getAmounts(): array
+    {
+        $totalAmount = Money::fromPoundsGBP(0);
+        $usedAmount = Money::fromPoundsGBP(0);
+
+        foreach ($this->campaignFundings as $campaignFunding) {
```
bdsl, line 104: We can only add currency codes if they're equal. IMHO it might be worth making a decision to properly remove multi-currency from matchbot at some point if we don't see any likelyhood of using a different currency, but while its there we can add this

```suggestion
        foreach ($this->campaignFundings as $campaignFunding) {
        Assertion::eq($campaignFunding->getCurrencyCode(), $this->currencyCode);
        
```

src/Domain/Fund.php
```diff
@@ -41,10 +48,17 @@ abstract class Fund extends SalesforceReadProxy
...
+     * @var Collection<int, CampaignFunding>
+     */
+    #[ORM\OneToMany(mappedBy: 'fund', targetEntity: CampaignFunding::class)]
+    protected Collection $campaignFundings;
+
     final public function __construct(string $currencyCode, string $name)
     {
         $this->createdAt = new \DateTime();
         $this->updatedAt = new \DateTime();
+        $this->campaignFundings = new ArrayCollection();
```
bdsl, line 61: maybe worth adding an (optional?) constructor param for campaignFundings so we can test what this does with them without reflection.

src/Domain/Fund.php
NoelLH: So far `getAmounts()` is only used by the fn right after it and isn't tested on its own. I think I expected the 2 new fns to end up longer when I made them distinct. I'm now wondering if it makes more sense to merge them if no separate use cases some up so that we don't have to think about this distinction, what do you think?

src/Domain/Fund.php
NoelLH: I just started trying this but the test's `CampaignFunding` needs `$fund` in its constructor so I think this leaves us with a circular chain when setting up the data? We could add a public setter but if it's only for tests I think that's perhaps no better than using reflection, what do you reckon?

bdsl: hmm good question, will have a think.

src/Domain/Fund.php
bdsl: I think it's probably worth keeping this separate for testability, since its implementing some important and non-trivial domain logic.

NoelLH: Thanks, I've aimed to do that in 77f77e5a9aa08dd0f55d3111c61bc234b22e2d56

src/Domain/Fund.php
bdsl: OK so this is the typical problem with aggregates as in this 2018 blog post [[Doctrine ORM and DDD aggregates](https://matthiasnoback.nl/2018/06/doctrine-orm-and-ddd-aggregates/)](https://matthiasnoback.nl/2018/06/doctrine-orm-and-ddd-aggregates/).

I think ideally if possible we should make the dependency one-way, probably keeping the new collection and removing the `\MatchBot\Domain\CampaignFunding::$fund` field. It looks like in prod the only use for that field is to do an instanceOf check on the Fund to find out if it's Pledge or Champion. We could denormalize that info into the CampaignFunding.

Having both fields introduces a possible source of errors since [Doctrine only checks the field on the owning side for updates](https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/reference/unitofwork-associations.html). Currently the owning side is `\MatchBot\Domain\CampaignFunding::$fund`. 

But reading https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/reference/association-mapping.html I'm learning that we can't lose the $fund field, because "A one-to-many association has to be bidirectional, unless you are using a join table. ", and I don't think it's worth setting up that join table even though it would make the PHP model a bit nicer.

But I don't think the reflection is great since it makes it hard to find that we've got a test using that field - I was a bit surprised to see that neither PHPStorm nor Psalm seem to follow the call from `new \ReflectionProperty($fund, 'campaignFundings');` to to the property definition, even if I replace $fund with `Fund::class`

I'll make a PR to this PR with the best I can come up with.

src/Application/Messenger/FundTotalUpdated.php
```diff
@@ -0,0 +1,33 @@
...
+
+use MatchBot\Application\Assertion;
+use MatchBot\Domain\Fund;
+use Symfony\Component\Messenger\Bridge\AmazonSqs\MessageGroupAwareInterface;
+
+readonly class FundTotalUpdated implements MessageGroupAwareInterface
+{
+    public array $jsonSnapshot;
+
+    protected function __construct(public string $salesforceId, array $json)
```
bdsl, line 13: Not a but, might be worth documenting a more specific type than `array`. Less critical since I think this is only used in one or two places.

Also not sure if there's any reason to use the promoted property for the salesforceId but not for the jsonSnapshot

src/Domain/FundRepository.php
```diff
@@ -156,17 +156,42 @@ protected function getNewFund(array $fundData): Fund
...
 
         return $fund;
     }
 
+    /**
+     * @return Fund[]
+     */
+    public function findForCampaignsClosedSince(DateTime $closedSinceDate): array
+    {
+        $now = new DateTime('now');
```
bdsl, line 174: I'd be happier passing this in for testability/flexibility instead of accessing the system clock directly here.

tests/Domain/FundTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+        $this->fund->addCampaignFunding(new CampaignFunding(
+            fund: $this->fund,
+            amount: '123.45',
+            amountAvailable: '100.00',
+            allocationOrder: 100,
+        ));
+    }
+
+    public function testGetAmounts(): void
```
bdsl, line 28: probably worth testing this a bit more thoroughly, e.g. test cases with zero, one and many amounts to sum. Might be a good fit for a data-driven test.

bdsl (dismissed):
LGTM - Added some comments inline, no blockers

src/Application/Messenger/FundTotalUpdated.php
NoelLH: True, I think it was a general `array` mostly because I first planned to extend `AbstractStateChanged` but that currently wants everything to have a UUID, and that didn't seem a good enough reason to give Funds an extra identifier.

I think because it's used quite narrowly I will probably leave it for now in the hopes of getting some version working today.

Good catch on the inconsistent promoted property use, I don't think there was a reason here or in `AbstractStateChanged`. Fixed both in https://github.com/thebiggive/matchbot/pull/1161/commits/355e05b884cdbea1f85f3dc38364e334a93c5b1b. This also helped me spot that there was a needless constructor in `DonationUpserted` so deleted that too.

src/Domain/FundRepository.php
NoelLH: Fair enough, I was mostly matching what the donations counterpart does without thinking. I don't think I will use it for tests yet but moved it to a parameter just in case. https://github.com/thebiggive/matchbot/pull/1161/commits/15f7b77c1497f33e421bf94bae10e35771a48bed

tests/Domain/FundTest.php
```diff
@@ -0,0 +1,103 @@
...
+     * @return list{array{
+     *     amountAvailable: Money,
+     *     expectedUsedAmount: Money,
+     *     totalAmount: Money,
+     * }}
+     */
+    private function amountsToSummarise(): array
+    {
+        /**
+         * @var list{array{
```
NoelLH, line 72: Psalm seems adamant that there is a mismatch between the implicitly detected types and the `@return` unless this is specified here, gave up trying to work out why. It doesn't complain of any issues at the point of assignment now so maybe a bug?

PR #1165, opened 2025-01-09
DON-998: Add action to allow testing collection of regular giving funds via http

dorota-joanna:
To use, point your browser at URL such as

http://localhost:30030/v1/test-donation-collection-for-date/2020-03-01?secret=<secret>

and you should see something like:

```
Big Give Matchbot
============================================================================================

Actual current date: Thu, 09 Jan 2025 18:33:07
Running regular giving collection process with simulated timestamp Sun, 01 Mar 2020 00:00:00 


This endpoint is only available in test/dev environments, not in production.

--------------
matchbot:collect-regular-giving starting!
Simulating running on 2020-03-01 00:00:00

 0 mandates may have donations to create at this time                           

 0 donations are due to have Payment Intent set at this time                    

 0 donations are due to be confirmed at this time                               

matchbot:collect-regular-giving complete!
--------------
```

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -83,6 +83,11 @@ return function (App $app) {
             ->add(PersonWithPasswordAuthMiddleware::class)
             ->add($ipMiddleware)
             ->add(RateLimitMiddleware::class);
+
+        $versionGroup->get(
+            '/test-donation-collection-for-date/{date}',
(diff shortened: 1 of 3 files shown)


NoelLH (approved):
Nice!

PR #1167, opened 2025-01-10
Don-998 show number of donations matched

bdsl:


src/Application/Commands/TakeRegularGivingDonations.php
```diff
@@ -98,6 +98,11 @@ private function createNewDonationsAccordingToRegularGivingMandates(\DateTimeImm
 
         foreach ($mandates as [$mandate]) {
             // @todo-regular-giving: catch the exception when missing address on account
+
+            // @todo-regular-giving: Make sure we don't create a donation here after the campaign has closed.
```
NoelLH, line 102: I think in this context it's _Regular Giving Collection End_ we would want to compare to, not End Date which is the last time _new_ donors can join the campaign. We'll need to expose the new field from SF to close this todo.

```suggestion
            // @todo-regular-giving: Don't create a donation after the campaign Regular Giving Collection End date, if any.
```

NoelLH (approved):
Looks good, suggested a comment tweak but not 100% certain about it.

src/Application/Commands/TakeRegularGivingDonations.php
bdsl: Right, it's also `campaign.Campaign_Status__c` which gets exposed to matchbot as part of `CampaignServiceModel.ready`.

PR #1170, opened 2025-01-13
Bg2 2797 push mandates to sf

bdsl:


.env.example
```diff
@@ -40,6 +40,7 @@ JWT_DONATION_SECRET=myLocalJwtSecret
 JWT_ID_SECRET=myLocalIdSecret
 
 SALESFORCE_CLIENT_TIMEOUT=15
+SALESFORCE_API_BASE=https://sf-api-staging.thebiggivetest.org.uk/
```
bdsl, line 43: IMHO this should replace the four environment variables below - but we can't delete them until after this new one is set in prod.

src/Application/Messenger/Handler/MandateUpsertedHandler.php
```diff
@@ -0,0 +1,63 @@
...
+        private EntityManagerInterface $entityManager,
+        private Clock $clock,
+    ) {
+    }
+
+    public function __invoke(MandateUpserted $message): void
+    {
+        $this->logger->info("MUH invoked for UUID: $message->uuid");
+        $sfId = $this->client->createOrUpdate($message);
+        $this->setSalesforceFields($message->uuid, $sfId);
```
bdsl, line 39: surprisingly to me this already seems to be working on my local - the mandates are getting SF ID's set in my local DB. I thought we would still need to create the endpoint in SF to accept the mandate upserts.

Logging in to full isn't working for me now though so I can't actually see how how the mandates look there.

src/Domain/RegularGivingMandate.php
```diff
@@ -352,4 +352,12 @@ public function createPendingFirstDonation(Campaign $campaign, DonorAccount $don
             billingPostcode: $donor->getBillingPostcode(),
         );
     }
+
+    public function toSFApiModel(): array
+    {
+        return [
+            'id' => $this->uuid,
+            // todo - fill in other properties, consider sharing code with toFrontEndApiModel
```
NoelLH, line 360:
```suggestion
            // @todo-regular-giving - fill in other properties, consider sharing code with toFrontEndApiModel
```

src/Client/Mandate.php
```diff
@@ -0,0 +1,33 @@
...
+namespace MatchBot\Client;
+
+use GuzzleHttp\Exception\RequestException;
+use MatchBot\Application\Messenger\DonationUpserted;
+use MatchBot\Application\Messenger\MandateUpserted;
+use MatchBot\Domain\DonationRepository;
+use MatchBot\Domain\Salesforce18Id;
+
+/**
+ * Client to push / upsert copies of donations to Salesforce.
```
NoelLH, line 14:
```suggestion
 * Client to push / upsert copies of Regular Giving mandates to Salesforce.
```

src/Client/Common.php
```diff
@@ -43,4 +47,91 @@ protected function getUri(string $uri, bool $withCache): string
 
         return $uri;
     }
+
+    /**
+     * @param 'donation'|'mandate' $entityType
+     */
+    protected function postUpdateToSalesforce(string $uri, array $jsonSnapshot, string $uuid, string $entityType): Salesforce18Id
```
NoelLH, line 54: I'm wondering if it would be slightly cleaner to pass the class name of the queued message verbatim, and use that along with static properties to derive `$uri` and `$entityType` here? Might be a little less error prone when editing and reduce the argument list size.

src/Application/Persistence/RegularGivingMandateEventSubscriber.php
```diff
@@ -0,0 +1,61 @@
...
+use Doctrine\ORM\Event\PostUpdateEventArgs;
+use Doctrine\ORM\Events;
+use MatchBot\Application\Assertion;
+use MatchBot\Application\Messenger\MandateUpserted;
+use MatchBot\Domain\RegularGivingMandate;
+use Stripe\Mandate;
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\RoutableMessageBus;
+
+class RegularGivingMandateEventSubscriber implements EventSubscriber
```
NoelLH, line 17: Nice!

NoelLH (dismissed):
Seems like a good simplification to get rid of `SalesfroceWriteProxyRepository`, especially as the pattern wasn't a good fit for pushing Funds and so that new way of pushing didn't extend it anyway.

Looking really good! Minor things inline, no blockers.

src/Application/Messenger/Handler/MandateUpsertedHandler.php
bdsl: realised it wasn't working,  was just getting confused by seeing artifacts of automated testing in my local db

src/Client/Common.php
bdsl: Yes. I think it's not that error prone for now as Psalm checks that entityType is one of the two allowed values from the docblock. Deriving the uri based on the entity type seems like a good plan but I think we can do that as part of the change to use the `SALESFORCE_API_BASE` env var after its set in all environments.

At that point it might also be worth merging the clients that talk to SF  (Donation, Mandate, maybe also Campaign and Fund) into a single class as another simplification. So `\MatchBot\Client\Campaign::getById` would become `\MatchBot\Client\Salesforce:::getCampaignById` and `\MatchBot\Client\Mandate::createOrUpdate` would become `\MatchBot\Client\Salesforce::createOrUpdateMandate`.  Not sure about that part.

PR #1186, opened 2025-01-21
Include donor details in mandate sent to SF so that SF can upser…

bdsl:
…t them as a contact

### DIFF
--- a/src/Application/Persistence/RegularGivingMandateEventSubscriber.php
+++ b/src/Application/Persistence/RegularGivingMandateEventSubscriber.php
@@ -9,17 +9,26 @@ use Doctrine\ORM\Event\PostUpdateEventArgs;
 use Doctrine\ORM\Events;
 use MatchBot\Application\Assertion;
 use MatchBot\Application\Messenger\MandateUpserted;
+use MatchBot\Domain\DonorAccount;
+use MatchBot\Domain\DonorAccountRepository;
 use MatchBot\Domain\RegularGivingMandate;
+use Psr\Container\ContainerInterface;
 use Stripe\Mandate;
 use Symfony\Component\Messenger\Envelope;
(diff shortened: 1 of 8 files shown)


src/Domain/RegularGivingMandate.php
```diff
@@ -37,7 +37,7 @@ class RegularGivingMandate extends SalesforceWriteProxy
     private readonly UuidInterface $uuid;
 
     #[ORM\Embedded(columnPrefix: 'person')]
-    public PersonId $donorId;
+    private PersonId $donorId;
```
bdsl, line 40: Unless I found an exception previously that I now can't remember, then all Doctrine properties should be private (or at least not public) to allow lazy-loading to work.

NoelLH (dismissed):

src/Application/Persistence/RegularGivingMandateEventSubscriber.php
```diff
@@ -9,17 +9,26 @@
...
 class RegularGivingMandateEventSubscriber implements EventSubscriber
 {
+    private ?RoutableMessageBus $bus = null;
+    private ?DonorAccountRepository $donorRepository = null;
+
     /**
      * @psalm-suppress PossiblyUnusedMethod
+     *
+     * Accepts whole container to lazy-load properties on first use, and allow for a circular dependency - this depends
+     * on a repository which depends on the EntityManager which depends on this.
```
bdsl, line 29: The circular dependency isn't great of course. I also tried breaking it by adding the subscriber to the EM in app bootstrap or index php, after the container has returned the EM, but sequencing that with `IntegrationTests\CreateRegularGivingMandateTest::setUp` that runs the bootstrap and then put test doubles in the container is awkward.

PR #1194, opened 2025-01-29
DON-1002: Add Big Give & Charity comms opt-in fields to Regular Givin…

bdsl:
…g Mandate

For now these are always false, will be set according to donor preferences when mandate is created from FE.


dorota-joanna (approved):
LGTM

PR #1196, opened 2025-01-30
DON-1002: Copy donors marketing communcation prefs from regular givin…

bdsl:
…g mandate to 1st donation linked to mandate

This should mean it gets picked at SF as it would for an ad-hoc donation and we won't have to change anything there.

Also setting the prefs on later donations in the mandate to null so that when those donations are pushed to SF they don't update any comms preferences - we wouldn't want to update preferences in the following months as it may no longer reflect the user's latest wishes.


dorota-joanna (approved):
LGTM

PR #1197, opened 2025-02-04
BG2-2798 – fund amounts push: fix types for compatibility with Salesforce `Decimal`

NoelLH:


bdsl (approved):
LGTM. The float for money makes me slightly nervous but I can't think of any case where it won't do rounding as we'd want it to when rendering the float in JSON. More ideally I think we'd either send the number as an integer of pence or add a feature to PHP to allow embedding decimal numbers directly in JSON.

PR #1198, opened 2025-02-04
DON-999 – error log if a known fund's value decreases

NoelLH:
Instead of ignoring silently

### DIFF
--- a/src/Domain/FundRepository.php
+++ b/src/Domain/FundRepository.php
@@ -103,6 +103,13 @@ class FundRepository extends SalesforceReadProxyRepository
 
                     $campaignFunding->setAmount($amountForCampaign);
                 }
+
+                if (bccomp($increaseInAmount, '0.00', 2) === -1) {
+                    $this->logger->error(
+                        "Funding ID {$campaignFunding->getId()} balance could not be negative-increased by " .
+                        "£{$increaseInAmount}"
+                    );
+                }
             } else {
                 // Not a previously existing campaign -> create one and set balances without checking for existing ones.
                 $campaignFunding = new CampaignFunding(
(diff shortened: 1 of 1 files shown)


src/Domain/FundRepository.php
```diff
@@ -103,6 +103,13 @@ public function pullForCampaign(Campaign $campaign): void
 
                     $campaignFunding->setAmount($amountForCampaign);
                 }
+
+                if (bccomp($increaseInAmount, '0.00', 2) === -1) {
+                    $this->logger->error(
+                        "Funding ID {$campaignFunding->getId()} balance could not be negative-increased by " .
+                        "£{$increaseInAmount}"
```
bdsl, line 110: Probably also worth logging the SF Fund ID here since I guess this will generally happen after someone has edited that fund? I haven't tested but I think it would be something like this:

```suggestion
                        "Funding ID {$campaignFunding->getId()} balance could not be negative-increased by " .
                        "£{$increaseInAmount}" + " SF Fund ID: " + $fundData['id']
```

bdsl (dismissed):
Added a suggestion inline. Also:

The comment about "Note that a balance DECREASE on the API side is unsupported and would be ignored" could be updated but not a blocker.

I was thinking about if we'd want to not only log an error but also throw but I guess not, although I'm not sure. Ideally we'd throw an error that would go back to the UI inside salesforce to stop the fund amounts being decreased in the first place.

src/Domain/FundRepository.php
NoelLH: Yep, good idea. Did that with slightly different syntax in https://github.com/thebiggive/matchbot/pull/1198/commits/23454e7e3b75e007a5e4d46fd30562ea763a5908

NoelLH:
> I was thinking about if we'd want to not only log an error but also throw but I guess not, although I'm not sure. Ideally we'd throw an error that would go back to the UI inside salesforce to stop the fund amounts being decreased in the first place.

Yeah, I don't think there is any built-in mechanism in Salesforce that would allow blocking a record update based on a callout unfortunately.

PR #1203, opened 2025-02-06
BG2-2798 – log fund update HTTP errors better & tidy up

NoelLH:


bdsl (approved):
As on slack, wondering what's the reason for [Revert "BG2-2798 – add FundTotalUpdated message class to `RoutableMessageBus` container"](https://github.com/thebiggive/matchbot/pull/1203/commits/037a4ce7f452c00655c519f1cc8ec69418c96256)? Should FundTotalUpdated not be in the list of message types with the others?

But looks like it should work either way not a blocker, and all other changes here LGTM.

PR #1204, opened 2025-02-06
BG2-2774: Send donation tip refund amount to Salesforce

bdsl:
This can replace Salesforce inferring a tip refund amount


integrationTests/DonationPersistenceTest.php
```diff
@@ -104,6 +104,7 @@ private function donationRow(): array
             'preAuthorizationDate' => null,
             'mandate_id' => null,
             'mandateSequenceNumber' => null,
+            'tipRefundAmount' => null,
```
NoelLH, line 107: Would be good to sync `api.yaml` to make it easier to find the expected type when working across apps.

src/Domain/Donation.php
```diff
@@ -1719,4 +1745,20 @@ public function thisIsInDateRangeToConfirm(DateTimeImmutable $now): bool
...
+        return $this->refundedAt;
+    }
+
+    public function getTipRefundAmount(): ?Money
+    {
+        if ($this->tipRefundAmount === null) {
+            return null;
+        }
+
+        // cheating a little by asserting that currency is GBP, since it is always likely to be.
```
NoelLH, line 1760:
```suggestion
        // @todo-multi-currency cheating a little by asserting that currency is GBP, since it is always likely to be.
```
I think this is fine, but shall we start a new todo marker to help whenever we have a use case for international support again?

bdsl:
@NoelLH it's worth reviewing this alongside related PR for SF.

NoelLH (approved):
LGTM, no blockers

integrationTests/DonationPersistenceTest.php
bdsl: Although we could add the new property into the API that FE uses - either now or at the time we need it. We could start to usefully display it since we now let people view old donations on the site.

src/Domain/Donation.php
bdsl: Happy to commit that change, although I was also wondering in the last months about if we should go the other way - if multi currency is unlikely enough to happen that we should do more to simplify the code by removing the support for multi currency that's there, or at least making a policy that we won't support it in new code.

So for instance we'd had that policy before we created the Money VO it wouldn't have had a currency property, it would just have been documented to always refer to GBP.

integrationTests/DonationPersistenceTest.php
NoelLH: @bdsl My preference is to use it as the description currently suggests:
> As Salesforce & MatchBot share a Donation schema, endpoints for both are documented here
    together; however operation `salesforceUpsert` is implemented only in Salesforce
    and all others are implemented only in MatchBot.

src/Domain/Donation.php
NoelLH: My expectation is that it will happen, just probably not this year.

bdsl: OK, will try and keep building mostly things based on the idea that we support a list of currencies that happens to currently include only GBP without relying on that being the only one.

PR #1215, opened 2025-02-13
Don 1003 allow regular giving without matching

bdsl:


tests/Domain/RegularGivingServiceTest.php
bdsl, line 1: this test is edited because I thought I was going to add the test for unmatched behaviour here. But then since in this test class the donation service is mocked it didn't seem like there was much I could actually test - the mandate service mostly just passes the value through to donation service.

So ended up not adding a test here and testing in integration tests instead. I think the refactors are worth keeping though.

src/Domain/Donation.php
```diff
@@ -796,7 +796,7 @@ public function getDonorCountryCode(): ?string
      */
     public function getFundingWithdrawalTotal(): string
     {
-        $withdrawalTotal = '0.0';
```
bdsl, line 799: adjusted to match what the Money class returns otherwise an assert in production code fails, and I think the format that it would return if non-zero.

NoelLH (approved):

PR #1221, opened 2025-02-17
DON-1008: Regular giving GA: Store homeIsOutsideUK field on donor acc…

bdsl:
…ounts

### DIFF
--- a/src/Migrations/Version20250217175219.php
+++ b/src/Migrations/Version20250217175219.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MatchBot\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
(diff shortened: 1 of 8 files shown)


src/Migrations/Version20250217175219.php
```diff
@@ -0,0 +1,25 @@
...
+final class Version20250217175219 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add field DonorAccount ADD homeIsOutsideUK';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE DonorAccount ADD homeIsOutsideUK TINYINT(1) DEFAULT NULL');
```
bdsl, line 18: all donor account records in prod currently have other home address details null, so this fits the pattern. Only expecting them to have non-null when we start collecting details for use with regular giving mandates.

NoelLH (approved):

PR #1222, opened 2025-02-18
DON-1022: Add controller for donor to cancel / discontinue a regular…

bdsl:
… giving mandate

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -83,7 +83,7 @@ return function (App $app) {
             ->add(PersonWithPasswordAuthMiddleware::class)
             ->add($ipMiddleware)
             ->add(RateLimitMiddleware::class);
-        $versionGroup->get('/regular-giving/my-donation-mandates/{mandateId:[a-z0-9-]{36}}/cancel', RegularGivingMandate\Cancel::class)
+        $versionGroup->post('/regular-giving/my-donation-mandates/{mandateId:[a-z0-9-]{36}}/cancel', RegularGivingMandate\Cancel::class)
             ->add(PersonWithPasswordAuthMiddleware::class)
             ->add($ipMiddleware)
             ->add(RateLimitMiddleware::class);
--- a/app/routes.php
+++ b/app/routes.php
@@ -83,6 +83,10 @@ return function (App $app) {
             ->add(PersonWithPasswordAuthMiddleware::class)
             ->add($ipMiddleware)
             ->add(RateLimitMiddleware::class);
+        $versionGroup->get('/regular-giving/my-donation-mandates/{mandateId:[a-z0-9-]{36}}/cancel', RegularGivingMandate\Cancel::class)
+            ->add(PersonWithPasswordAuthMiddleware::class)
+            ->add($ipMiddleware)
+            ->add(RateLimitMiddleware::class);
 
         $versionGroup->get(
             '/test-donation-collection-for-date/{date}',
--- a/src/Domain/MandateCancellationType.php
+++ b/src/Domain/MandateCancellationType.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace MatchBot\Domain;
+
+enum MandateCancellationType: string
+{
+    /**
+     * Auto cancelled at creation because enrolling or matching one of the initial donations failed
+     */
+    case EnrollingDonationFailed = 'EnrollingDonationFailed';
+
+    /**
+     * Auto cancelled at creation because the first donation payment could not be collected
+     */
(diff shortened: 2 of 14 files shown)


src/Domain/MandateCancellationType.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace MatchBot\Domain;
+
+enum MandateCancellationType: string
```
bdsl, line 5: not sure if introducing this enum and associated column on mandate table makes sense. Could instead have split the existing `\MatchBot\Domain\MandateStatus::Cancelled` case into three to discriminate between the three possibilities, and could still map them all to 'Cancelled'  the same thing when sending to SF and FE.

app/routes.php
```diff
@@ -83,6 +83,10 @@
             ->add(PersonWithPasswordAuthMiddleware::class)
             ->add($ipMiddleware)
             ->add(RateLimitMiddleware::class);
+        $versionGroup->get('/regular-giving/my-donation-mandates/{mandateId:[a-z0-9-]{36}}/cancel', RegularGivingMandate\Cancel::class)
+            ->add(PersonWithPasswordAuthMiddleware::class)
+            ->add($ipMiddleware)
+            ->add(RateLimitMiddleware::class);
```
bdsl, line 89: may be worth refactoring at this point (or when this is being manually tested with FE implementation) to combine with other routes into a group with shared middlewares.

Also (and I'm sure I already wrote this comment but now can't find it) not sure if it's really worthwhile having the mandateId in both the URI and the request body. It allows checking that they match in the controller but not sure that really does much, both are likely to be set based on the same variable in FE. What do you think @NoelLH ?

NoelLH: I do find it helpful to have IDs in the route for things like this @bdsl; it quite frequently makes it easier to work with Apache logs (as well as following what I've historically understood to be RESTful API design received wisdom). But it's also a bit odd and annoying to work with an object that's missing its most fundamental property for unique identification. So although I don't like the redundancy of doing both, I think it might still be the least bad option.

bdsl: OK, will leave it having both.

NoelLH (commented):
LGTM. No blockers except the field lengths thing.

src/Domain/MandateCancellationType.php
bdsl: OK will keep as is. For clarity this is adding both the enum for the general type of cancellation (e.g. donor cancelled vs initial payment failed) as well as a free text field for any reason given by the donor. I think there will be some business logic that will depend on the distinction, e.g. when donor looks at their list of mandates we'd want to show ones they cancelled but not ones we auto-cancelled. Although we could say that logic is only show mandates that have at least 1 successful donation each.

tests/Domain/DonationServiceTest.php
```diff
@@ -160,7 +161,12 @@ public function testRefusesToConfirmPreAuthedDonationForNonActiveMandate(): void
 
         $this->donorAccountRepoProphecy->findByStripeIdOrNull($stripeCustomerId)->willReturn($donor);
 
-        $mandate->cancel();
+        $mandate->cancel(
+            reason: '',
```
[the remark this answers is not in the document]
bdsl: Do you mean here in the test case? I don't see what that would help. The production code does already fill something in to this field in every case.

tests/Domain/RegularGivingServiceTest.php
```diff
@@ -540,6 +541,35 @@ public function testDoesNotLeaveDonorAddressChangedIfDonationServiceThrows(): vo
...
+
+    public function testCancellingMandateCancelsPendingDonations(): void
+    {
+        $sut = $this->makeSut(new \DateTimeImmutable('2024-11-29T05:59:59 GMT'));
+        $mandate = $this->getMandate(2, '2024-09-03T06:00:00 BST', 1);
+
+        $donation = self::someDonation();
+        $this->donationRepositoryProphecy->findPendingAndPreAuthedForMandate($mandate->getId())->willReturn([$donation]);
+
+        $this->donationServiceProphecy->cancel($donation)->shouldBeCalled();
```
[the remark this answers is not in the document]
bdsl: I don't think it really matters. If you call it twice all it does the second time is log an info "Donation ID {$donation->getUuid()} was already Cancelled".

tests/Domain/DonationServiceTest.php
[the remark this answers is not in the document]
NoelLH: I lost track of the context – all good!

PR #1224, opened 2025-02-19
DON-1023 – add `RegularGivingMandate\CancelAsAdmin` for Salesforce callouts

NoelLH:


NoelLH:
Came to slightly regret doing a non-integration functional test here (especially while my local Docker setup is flaky and sometimes not running the latest code/tests). We need to mock a lot of repo's now for the Regular Giving service. I would probably go for an integration test instead next time, but I'm also wondering if we should be splitting up the service a bit. I had to mock several things that the action under test doesn't touch.

src/Domain/RegularGivingService.php
```diff
@@ -390,14 +390,18 @@ private function confirmWithNewPaymentMethod(Donation $firstDonation, StripeConf
...
      */
-    public function cancelMandate(RegularGivingMandate $mandate, string $reason): void
-    {
+    public function cancelMandate(
+        RegularGivingMandate $mandate,
+        string $reason,
+        MandateCancellationType $cancellationType,
+    ): void {
         $mandateId = $mandate->getId();
         Assertion::notNull($mandateId);
```
bdsl, line 402:
```suggestion
        Assertion::notNull($mandateId);
        Assertion::inArray($cancellationType, [MandateCancellationType::DonorRequestedCancellation, MandateCancellationType::BigGiveCancelled]);
```

bdsl:
> Came to slightly regret doing a non-integration functional test here (especially while my local Docker setup is flaky and sometimes not running the latest code/tests). We need to mock a lot of repo's now for the Regular Giving service. I would probably go for an integration test instead next time, but I'm also wondering if we should be splitting up the service a bit. I had to mock several things that the action under test doesn't touch.

Yes I generally find it's not worth doing non-integration tests for controllers, there's too much boilerplate type stuff that we need which takes time when writing the test initially and also makes it harder for us to change stuff.

I think it's probably better to either test at the level under controllers, or do integration tests so use the same DI setup as production, or if we want to do more tests at controller level have a shared test setup routine, e.g. creating a fake of every repository that's either empty or could even have some shared fixture data. We could create versions of `dependencies.php` and `repositories.php` that are for tests, maybe replace each repository with an in-memory repo.

Also not opposed at all to splitting up the RegularGivingService if we find a good way to split it.

bdsl (approved):
LGTM

PR #1227, opened 2025-02-20
BG2-2838 – support allocating topup pledges last

NoelLH:


integrationTests/IntegrationTest.php
```diff
@@ -272,7 +272,7 @@ public function addFundedCampaignAndCharityToDB(
             isRegularGiving: $isRegularGiving
         );
         ['fundId' => $fundId, 'campaignFundingId' => $campaignFundingId] =
-            $this->addFunding($campaignId, $fundWithAmountInPounds, 1, Pledge::DISCRIMINATOR_VALUE);
+            $this->addFunding($campaignId, $fundWithAmountInPounds, 1, 'pledge');
```
NoelLH, line 275: Not ideal to be going back to strings in tests like this. But the below changes have an attribute relying on a const `DISCRIMINATOR_MAP`, all outside the class, because of the attribute. I think making that critical code DRYer is a bigger win than avoiding the strings in tests. Open to ideas of how we can cleanly do both.

src/Domain/TopupPledge.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MatchBot\Domain;
+
+use Doctrine\ORM\Mapping as ORM;
+
+#[ORM\Entity]
```
bdsl, line 9: Could we write a docblock to explain what a TopupPledge is? I could guess but I don't think I know well enough to make a suggestion.

Also could the names or class structure be worth rethinking - based on the names you might think a TopupPledge is a subtype of Pledge but that's not what we have in the class structure.

And I'm wondering whether we should keep the multiple classes, given that they don't have substantially different behaviour - they only differ in their `NORMAL_ALLOCATION_ORDER` values.

NoelLH: I find the discriminator map behaviour quite good to work with for these in general; not sure about losing the class-per-type.

I'm not sure that it really makes it clearer to have e.g. two concrete pledges extend an abstract one – mostly I was trying to prioritise aligning with terminology in Salesforce and existing docs.

I agree there should be a doc block on this class, added a first pass in 589dda710b06b2a81d0ca3792822d6d376a9d756

bdsl: Thanks. I wanted to see how it would look without the class per type. This isn't fully working code but is most of the way there. See what you think. https://github.com/thebiggive/matchbot/pull/1228

NoelLH: Thanks for that @bdsl, I agree it looks like it could be a definite improvement. I think I've been biased by other situations where discriminator maps maybe added a bit more value.

NoelLH:
Merging this approach for now as discussed. We'll probably flatten the models shortly and drop the discriminator map entirely, rather than refining how we use it.

PR #1228, opened 2025-02-20
BG2-2838: WIP collapse four Fund classes into one - maybe simpler?

bdsl:
I started doing this to test out if it makes the code simpler, I think it probably does although I gave up part way through updating the tests.

Would be probably better still using an enum instead of a string for the type.

### DIFF
--- a/src/Domain/Fund.php
+++ b/src/Domain/Fund.php
@@ -154,7 +154,7 @@ class Fund extends SalesforceReadProxy
         return [
             'currencyCode' => $amounts['totalAmount']->currency->isoCode(),
             'fundId' => $this->getId(),
-            'fundType' => $this->getDiscriminatorValue(),
+            'fundType' => $this->fundType,
             'salesforceFundId' => $sfId,
             'totalAmount' => (float) $amounts['totalAmount']->toNumericString(),
             'usedAmount' => (float) $amounts['usedAmount']->toNumericString(),


src/Domain/Fund.php
```diff
@@ -27,28 +21,26 @@
...
-    ...DISCRIMINATOR_MAP,
-])]
 #[ORM\HasLifecycleCallbacks]
-abstract class Fund extends SalesforceReadProxy
+class Fund extends SalesforceReadProxy
 {
-    use TimestampsTrait;
-
-    /** @var positive-int */
-    public const int NORMAL_ALLOCATION_ORDER = 999;
```
bdsl, line 42: the 999 I think was never used for anything since it's overidden by all child classes.

src/Domain/Fund.php
```diff
@@ -183,4 +173,13 @@ public function addCampaignFunding(CampaignFunding $funding): void
...
     }
+
+    public function allocationOrder(): int
+    {
+        return match ($this->fundType) {
+            'championFund' => 200,
+            'pledge' => 100,
+            'topupPledge' => 300,
+        };
+    }
```
bdsl, line 184: I think having all the order numbers together like this is quite valuable to be see at a glance what order they're used in (especially if we sorted them to numerical order in the code)

src/Domain/Fund.php
NoelLH: Yeah, should have just been abstract probably as the class can't be used as-is.

bdsl: I don't think you can have an abstract const, but it could be an abstract method. In PHP 8.4 it could be an abstract property.

bdsl:
Closing in favour of https://github.com/thebiggive/matchbot/pull/1230

PR #1230, opened 2025-02-21
BG2-2838: Extract and move function \MatchBot\Domain\Fund::getAllocat…

bdsl:
…ionOrder to FundRepository

### DIFF
--- a/src/Domain/Fund.php
+++ b/src/Domain/Fund.php
@@ -183,4 +183,14 @@ abstract class Fund extends SalesforceReadProxy
 
         $this->campaignFundings->add($funding);
     }
+
+    /**
+     * @return positive-int
+     */
+    public function getAllocationOrder(): int
+    {
+        $order = $this::NORMAL_ALLOCATION_ORDER;
+        \assert(is_int($order) && $order > 0);
+        return $order;
+    }
 }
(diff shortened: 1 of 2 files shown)


src/Domain/Donation.php
```diff
@@ -849,8 +848,7 @@ private function getConfirmedPledgeWithdrawalTotal(): string
 
         $withdrawalTotal = '0.0';
         foreach ($this->fundingWithdrawals as $fundingWithdrawal) {
-            // Rely on Doctrine `SINGLE_TABLE` inheritance structure to derive the type from the concrete class.
-            if ($fundingWithdrawal->getCampaignFunding()->getFund() instanceof Pledge) {
+            if ($fundingWithdrawal->getCampaignFunding()->getFund()->getFundType() === FundType::Pledge) {
```
bdsl, line 851: @NoelLH I think this is a bug (both before and after this PR is merged). We should include TopUpPledges along with regular pledges in this total, shouldn't we? Or not?

Either way I can switch to using `match` instead of `===` so we're forced to consider it if we add any other cases to the enum. It would be nice if Psalm had alerted me at the point I had an instanceof that would always return true, but it didn't.

NoelLH: Yes, good catch! Thanks. I was just looking at this code and still missed it 🤦

NoelLH (dismissed):

src/Domain/Donation.php
bdsl: Fixed here https://github.com/thebiggive/matchbot/pull/1230/commits/66e8cfcbf486837481c0dbf80fa20f1900628ca7

PR #1231, opened 2025-02-21
DON-1003 – send champion matching for preauth'd donations to Salesforce

NoelLH:


src/Domain/Donation.php
```diff
@@ -827,15 +828,7 @@ private function getConfirmedChampionWithdrawalTotal(): string
...
-        $withdrawalTotal = '0.0';
-        foreach ($this->fundingWithdrawals as $fundingWithdrawal) {
-            // Rely on Doctrine `SINGLE_TABLE` inheritance structure to derive the type from the concrete class.
-            if ($fundingWithdrawal->getCampaignFunding()->getFund() instanceof ChampionFund) {
-                $withdrawalTotal = bcadd($withdrawalTotal, $fundingWithdrawal->getAmount(), 2);
-            }
-        }
-
-        return $withdrawalTotal;
+        return $this->getWithdrawalTotalByFundType(ChampionFund::class);
```
bdsl, line 831: this is going to conflict with the PR I've just created (see my comment at https://github.com/thebiggive/matchbot/pull/1230/files#r1965884768 )

src/Domain/Donation.php
```diff
@@ -847,10 +840,31 @@ private function getConfirmedPledgeWithdrawalTotal(): string
             return '0.0';
         }
 
+        return $this->getWithdrawalTotalByFundType(Pledge::class);
+    }
+
+    /**
+     * @return numeric-string
+     */
+    private function getPreAuthorizedChampionWithdrawalTotal(): string
```
bdsl, line 849: I might have to step back to take a wider look, but this seems confusing as a concept, and also hard to see when just focusing on this class why it's only ChampionFund withdrawals that we want a preauthed figure for and not the other two types of Fund.

bdsl: OK I see that returning zero depending on the status is following an existing pattern even if it might be surprising.

This and the similar functions are all always called together, so I think it would be worth combining them into one that just iterates through `$this->fundingWithdrawals` once and returns all the totals together. This would be a bit more performant but the main advantage is that then it can throw (and possibly also fail a static check) if it encounters any fund of an unexpected type. 

Currently we'd ignore any fund of an unexpected type, which I think may be a bug in the case of a topup fund.

src/Domain/Donation.php
```diff
@@ -847,10 +840,31 @@ private function getConfirmedPledgeWithdrawalTotal(): string
...
+    {
+        if ($this->getDonationStatus() !== DonationStatus::PreAuthorized) {
+            return '0.0';
+        }
+
+        return $this->getWithdrawalTotalByFundType(ChampionFund::class);
+    }
+
+    /**
+     * @param class-string $fundType
```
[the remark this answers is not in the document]
bdsl: suggestion no longer relevant given merge conflicts

NoelLH:
Closing in favour of https://github.com/thebiggive/matchbot/pull/1232

PR #1234, opened 2025-02-24
DON-1119: Allow donor to complete 3D Secure authentication for fi…

bdsl:
I think this is good to merge as-is, although there's some more stuff we could do:

- Tidy up by never activating a mandate during the confirm request from frontend, and relying only on activation based on charge event from Stripe. We have to be able to do the latter for 3DS mandates so may as well do for all. That also means we can remove the new exception and just use return values.

- Add controller that FE can use to cancel the mandate when 3DS fails. In theory it could use `\MatchBot\Application\Actions\RegularGivingMandate\Cancel` but I think it's better to keep that for cancellation explicitly invoked by donor after mandate fully activated.

- Auto cancel pending mandate when same donor requests a new mandate for same campaign. May do this as part of a separate upcoming ticket.

- Decide if we're OK with updating the home address & postcode of a donor based on a new mandate even if they fail or walk away from 3DS authentication. I think it's fine to update it and quite a bit more work to have both the new and old details persisted during the time they're looking at the 3DS screen.

- Decide if we're happy with 5 minutes as the limit for a donor to complete 3DS authentication

- Add some tests.

### DIFF
--- a/src/Application/Commands/ExpirePendingMandates.php
+++ b/src/Application/Commands/ExpirePendingMandates.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MatchBot\Application\Commands;
+
+use MatchBot\Domain\DonationRepository;
+use MatchBot\Domain\DonationService;
+use MatchBot\Domain\MandateCancellationType;
+use MatchBot\Domain\RegularGivingMandateRepository;
+use MatchBot\Domain\RegularGivingService;
+use Psr\Clock\ClockInterface;
+use Ramsey\Uuid\Uuid;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+#[AsCommand(
+    name: 'matchbot:expire-pending-mandates',
+    description: "Cancels regular giving mandates left pending (e.g. while donor considered 3DS authentication)",
+)]
+class ExpirePendingMandates extends LockingCommand
+{
+    public function __construct(
+        private RegularGivingMandateRepository $regularGivingMandateRepository,
+        private RegularGivingService $regularGivingService,
+        private ClockInterface $clock,
+    ) {
+        parent::__construct();
+    }
+
+    protected function doExecute(InputInterface $input, OutputInterface $output): int
+    {
+        $fiveMinutes = new \DateInterval('PT5M');
+        $now = $this->clock->now();
+
+        $mandatesToCancel = $this->regularGivingMandateRepository->findAllPendingSinceBefore(
+            $now->sub($fiveMinutes)
+        );
+
+        foreach ($mandatesToCancel as $mandate) {
+            $this->regularGivingService->cancelMandate(
+                $mandate,
+                'pending mandate expired at ' . $now->format('c') . ", donor may have walked away from 3DS",
+                MandateCancellationType::FirstDonationUnsuccessful
+            );
+        };
+
+        return 0;
+    }
+}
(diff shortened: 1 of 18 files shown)


src/Application/Commands/ExpirePendingMandates.php
```diff
@@ -0,0 +1,58 @@
...
+        private RegularGivingService $regularGivingService,
+        private ClockInterface $clock,
+        private LoggerInterface $log,
+    ) {
+        parent::__construct();
+    }
+
+    protected function doExecute(InputInterface $input, OutputInterface $output): int
+    {
+        $fiveMinutes = new \DateInterval('PT5M');
```
NoelLH, line 37: This seems a little low to me. I know at least one bank I use has a 9 minute timeout on their side; might be worth a straw poll of the team to get more examples. My instinct is to go for around 15 minutes, to minimise the risk that 3DS could succeed and then we still have an expired mandate. (I'm not yet clear what happens if a donor hits this scenario?)

src/Domain/DoctrineDonationRepository.php
```diff
@@ -185,6 +185,7 @@ public function findWithExpiredMatching(\DateTimeImmutable $now): array
             ->innerJoin('d.fundingWithdrawals', 'fw')
             ->where('d.donationStatus IN (:expireWithStatuses)')
             ->andWhere('d.createdAt < :expireBefore')
+            ->andWhere('d.mandate is null')
```
NoelLH, line 188: I'm not sure I get why we need this condition vs. relying on the statuses?

NoelLH (approved):
Nice, LGTM. Minor things inline; I think we may want to go higher than 5 minutes but not a blocker for testing.

Re. one of the description points, I don't know that we would want frontend to immediately cancel a mandate on 3DS failure? If the donor is to get another try could we keep it active, or are you thinking it's better to do that on a new mandate and reset the timer?

src/Application/Commands/ExpirePendingMandates.php
bdsl: We cancel the payment attempt via the stripe API when this expires, so the donor should get a payment failure message and our DB will be left basically as it would be if they had just walked away when they saw the 3DS popup. I think I asked most of the team yesterday, 5 seemed OK but I can't immediately think what the risk would be if its too high, although it does become more important to proactively cancel if they fail authentication or if they close the tab and then attempt to start a new mandate.

src/Domain/DoctrineDonationRepository.php
bdsl: the 1st donation in a mandate would be `pending`, and for regular giving if the donor walks away from 3DS we don't want to just remove the match funds from it as this would do, we want to fully cancel it and cancel the associated mandate. I don't want a situation where a donor can lose out on matching without knowing it just by being slow to complete 3DS.

In practice with current numbers this wouldn't be an issue because the mandate donations are auto cancelled quicker than the ad-hoc donations have match funds released, but I don't want to rely on that here.

bdsl:
> Re. one of the description points, I don't know that we would want frontend to immediately cancel a mandate on 3DS failure? If the donor is to get another try could we keep it active, or are you thinking it's better to do that on a new mandate and reset the timer?

As it works currently the existing mandate isn't auto cancelled but if the donor tries again it will be a new mandate anyway. So it's an improvement from that to cancel on failure instead of having two at once. But I agree as suggested here https://github.com/thebiggive/donate-frontend/blob/0e7040a587ef8f53011da77fd5887e3251dbffef/src/app/regular-giving/regular-giving.component.ts#L340 that it's probably better to let them retry with the existing mandate, which might mean we'd want to find a way to reset the timer.

src/Domain/DoctrineDonationRepository.php
NoelLH: OK, fair enough. Possibly worth a comment so it's clear why later?

PR #1236, opened 2025-02-26
DON-1119 – start activating mandate only on charge hook

NoelLH:


tests/Domain/RegularGivingServiceTest.php
```diff
@@ -181,9 +181,9 @@ public function testItCreatesRegularGivingMandate(): void
         $this->assertEquals(Country::fromEnum(CountryAlpha2::Kiribati)->alpha2->value, $this->donorAccount->getBillingCountryCode());
         $this->assertSame('KI0107', $this->donorAccount->getBillingPostcode());
 
-        $this->assertSame(MandateStatus::Active, $mandate->getStatus());
-
-        $this->regularGivingNotifierProphecy->notifyNewMandateCreated(Argument::cetera())->shouldBeCalled();
+        // Pending, and no notification, until Stripe first donation charge succeeded webhook.
+        $this->assertSame(MandateStatus::Pending, $mandate->getStatus());
+        $this->regularGivingNotifierProphecy->notifyNewMandateCreated(Argument::cetera())->shouldNotBeCalled();
```
bdsl, line 186: IMHO we can delete this line, we can never assert all the things that should not be called and prophecy will assume it should not be called just by us not saying it should be called.

src/Domain/RegularGivingService.php
```diff
@@ -212,8 +214,6 @@ public function setupNewMandate(
             );
```
bdsl, line 214: the throw just above should now never happen - if the donation was not collected presumably `PaymentIntentNotSucceeded` was thrown at line 188 above, so I don't think the if check here is useful.

bdsl (approved):
LGTM, I think there's more to clear up here but this is fine to stop the immediate errors.

tests/Domain/RegularGivingServiceTest.php
NoelLH: Yep OK, removed in https://github.com/thebiggive/matchbot/commit/3f63f378e0610bed891780bb64e4db654f65e9c2

src/Domain/RegularGivingService.php
NoelLH: I think we've been discussing two alternative changes around that exception which are mutually exclusive, so I'm not really clear on the best change here now, and don't want to apply one unilaterally. If we narrow the exception's meaning to be _failure_ statuses (`canceled`) and not pending ones, then I think this remains needed.

NoelLH: Yeah, compared to always catching there is no difference. I don't feel strongly that we need the exception.

PR #1237, opened 2025-02-26
DON-1119: Fix payment method not saved to donor account for regular giving mandate with 3DS

bdsl:


src/Domain/RegularGivingService.php
```diff
@@ -369,31 +354,6 @@ private function enrollAndMatchDonations(array $donations, RegularGivingMandate
         }
     }
 
-    /**
-     * @throws RegularGivingDonationToOldToCollect
-     * @throws StripeApiErrorException
-     * @throws PaymentIntentNotSucceeded
-     */
-    private function confirmWithNewPaymentMethod(Donation $firstDonation, StripeConfirmationTokenId $confirmationTokenId): StripePaymentMethodId
```
bdsl, line 377: this method goes away because since since we have to rely on the mandate being activated in the stripe hook only the first line is useful, so that gets inlined at the call site.

NoelLH (commented):
Thanks, all LGTM

PR #1241, opened 2025-02-27
Don 995 auto cancel mandates collection end

bdsl:


src/Application/Commands/TakeRegularGivingDonations.php
```diff
@@ -118,9 +133,10 @@ private function createNewDonationsAccordingToRegularGivingMandates(\DateTimeImm
             try {
                 $donation = $this->makeDonationForMandate($mandate);
                 if ($donation) {
+                    $this->reportableEventHappened = true;
                     $io->writeln("created donation {$donation}");
                 } else {
-                    $io->writeln("no donation created for {$mandate} as collection end date passed");
```
bdsl, line 123: this message wasn't accurate - `makeDonationForMandate` also returns null if it doesn't need to create a donation now because no payment is due until later in the month.

PR #1244, opened 2025-02-28
DON-1005: Disallow donor from having two active or pending mandates for one campaign

bdsl:
…tes for same campaign

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -242,5 +242,12 @@
             "@sa:check",
             "@lint:check"
         ]
+    },
+    "extra": {
+        "patches": {
+            "doctrine/dbal": {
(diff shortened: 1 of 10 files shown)


composer.json
```diff
@@ -240,5 +242,12 @@
             "@sa:check",
             "@lint:check"
         ]
+    },
+    "extra": {
+        "patches": {
+            "doctrine/dbal": {
+                "Doctrine migrations functional index workaround": "dbal-funcational-index.patch.txt"
+            }
+        }
```
bdsl, line 251: Having copied the method of indexing out of identity and then found the need to patch dbal, I copied this patch back into identity in https://github.com/thebiggive/identity/pull/193

PR #1247, opened 2025-03-03
DON-1021: WIP allow changing payment card used for regular giving

bdsl:


src/Application/Actions/RegularGivingMandate/UpdatePaymentMethod.php
```diff
@@ -0,0 +1,72 @@
...
+        \assert(is_string($paymentMethodId));
+
+        $methodId = StripePaymentMethodId::of($paymentMethodId);
+
+        try {
+            $paymentMethod = $this->stripe->retrievePaymentMethod($donor->stripeCustomerId, $methodId);
+        } catch (InvalidRequestException $e) {
+            throw new HttpBadRequestException($request, 'Could not load new payment method:' . $e->getMessage());
+        }
+
```
bdsl, line 64: I think we also need to disconnect the previous regular giving payment from the account at this point, which probably also means we should move the work out of this class.

NoelLH (approved):

NoelLH (approved):

PR #1250, opened 2025-03-06
DON-400: Refactor, prepare to start sending donation success emails from here instead of Salesforce

bdsl:


src/Domain/DonationNotifier.php
```diff
@@ -0,0 +1,78 @@
...
+            "collectedAt should not be null for successful donation: {$donation}"
+        );
+
+
+        $campaign = $donation->getCampaign();
+        $charity = $campaign->getCharity();
+
+        $fundingWithdrawalsByType = $donation->getWithdrawalTotalByFundType();
+
+        $matchedAmount = bcadd($fundingWithdrawalsByType['amountMatchedByPledges'], $fundingWithdrawalsByType['amountMatchedByChampionFunds'], 2);
```
bdsl, line 40: move this domain logic into the donation class, if there isn't an equivalent already there.

bdsl: hmm so donation does already have `getFundingWithdrawalTotal` which I think should return the same thing https://github.com/thebiggive/matchbot/blob/1f7b529bcd14975eb3bf98a739a663fa5c238a1d/src/Domain/Donation.php#L806

but that's looking at the total of all withdrawals, whereas the existing SF logic calculating totalMatchedAmount in pardotUtils is specifically looking at only pledges and champion funds. Afaik those are the only two types of funds so that's a distinction without an immediate difference and using a total of all is if anything better.

@NoelLH does that seem right to you? I'll switch this over to use the total of everything.

src/Domain/SendEmailCommand.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace MatchBot\Domain;
+
+/**
+ * @psalm-type emailParams array<string,string|null|int|float|array>
+ */
+readonly class SendEmailCommand
```
NoelLH, line 8: I'm wondering if there's a slight semantic clash here, with Symfony and existing MatchBot conventions of using 'Command' mostly to mean console commands?

Personally I would also maybe put it somewhere more specific to how it's called than `Domain` – perhaps the namespace could be `Application\Notify` or something and the class could be `Email`?

src/Domain/DonationNotifier.php
bdsl: Done in 59155cc7

NoelLH (dismissed):
Thoughts on class naming inline but LGTM, no blockers.

src/Domain/SendEmailCommand.php
bdsl: Yep I was thinking it might be better to call it email or something like that, and then change the function name so we do `$mailer->send($email)` instead of `$mailer->accept($command)`. I think I was thinking of it as domain since it's only depending on I/O free domain stuff like donation, but there's no rule that we can't have classes like that anywhere else. Will see where it could go.

bdsl: Probably influenced because we just read this refactoring in the book in my book club https://refactoring.com/catalog/replaceFunctionWithCommand.html

src/Application/Email/EmailMessage.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace MatchBot\Application\Email;
+
+use MatchBot\Domain\EmailAddress;
+
+/**
+ * A message to be sent email, via our Mailer service
```
bdsl, line 8:
```suggestion
 * A message to be sent by email, via our Mailer service
```

src/Domain/DonationNotifier.php
NoelLH: Yes that seems right @bdsl, even with topup pledges there are no plans to involve a third Salesforce object; so the sum of the two we have now should be the same as the sum of all funding.

PR #1252, opened 2025-03-07
MAT-400: Send emails via mailer directly

bdsl:
Email doesn't yet include optional elements like charity postal address, but should be enough to check the basics work.


tests/Domain/DonationTest.php
```diff
@@ -256,7 +256,7 @@ public function testToSfAPIModel(): void
                 'homePostcode' => null,
                 'lastName' => 'N/A',
                 'mandate' => null,
-                'matchedAmount' => 0,
```
bdsl, line 259: not sure how this needed changing in this PR, but seems like a harmless change.

NoelLH (approved):

PR #1257, opened 2025-03-11
MAT-400: Include charity phone number and address when sending donation confirmat…

bdsl:
…ion email

Also preparing to include postal and email addresses


src/Client/Campaign.php
```diff
@@ -16,6 +16,15 @@
  *      id: string,
  *      name: string,
  *      logoUri: ?string,
+ *      phoneNumber: ?string,
+ *      emailAddress: ?string,
```
bdsl, line 20: this is technically a lie - sf doesn't actually return emailAddress here yet, but I think we need to add it. I'll do that next and make matchbot tolerate it being missing.

src/Domain/PostalAddress.php
```diff
@@ -0,0 +1,61 @@
...
+        ?string $city,
+        ?string $postalCode,
+        ?string $country,
+    ): self {
+        return new self($line1, $line2, $city, $postalCode, $country);
+    }
+
+    private function __construct(
+        #[Column(nullable: true)]
+        public string $line1,
```
bdsl, line 28: looks like Doctrine recognises that this property is non-nullable and gives the Charity a null instead of an address object if line1 is null in the db, which is what I want.

src/Domain/PostalAddress.php
```diff
@@ -0,0 +1,61 @@
...
+        ?string $line2,
+        ?string $city,
+        ?string $postalCode,
+        ?string $country,
+    ): self {
+        return new self($line1, $line2, $city, $postalCode, $country);
+    }
+
+    private function __construct(
+        #[Column(nullable: true)]
```
bdsl, line 27: took me a while trying to work out why doctrine seemed to be ignoring the `Embedded` annotation on the Charity until I realised I also had to annotate each property here.

NoelLH (dismissed):
LGTM! Liking the Embeddable PostalAddress, definitely a nice fit here.

bdsl:
I realised this is going to crash if it gets to prod as it is before https://github.com/thebiggive/salesforce/pull/2244 - so I think I need to add a bit to make it tolerate the phone number and postal address keys being missing in data from SF.

PR #1258, opened 2025-03-11
MAT-403 – push to Salesforce via a separate, lower-priority queue

NoelLH:


app/dependencies.php
```diff
@@ -500,6 +511,23 @@ static function (ContainerInterface $c): Matching\Adapter {
             return new RoutableMessageBus($busContainer, $bus);
         },
 
+        SalesforcePushTransport::class => static function (): TransportInterface {
+            $transportFactory = new TransportFactory([
+                new AmazonSqsTransportFactory(),
+                new InMemoryTransportFactory(), // For unit tests (without Docker Compose setup).
```
NoelLH, line 517:
```suggestion
                new InMemoryTransportFactory(), // For unit tests.
```
(Docker Compose was only relevant before I put the `force`d vars in phpunit config)

app/dependencies.php
```diff
@@ -339,12 +343,19 @@ static function (ContainerInterface $c): Matching\Adapter {
...
             $sendersLocator = new SendersLocator(
                 [
+                    // Outbound for ClaimBot; Redis queue in Production.
                     Messages\Donation::class => [ClaimBotTransport::class],
+
+                    // Outbound, priority, for MatchBot worker; SQS queue in Production.
+                    // `CharityUpdated` does call out to Salesforce, to read data, but it's rarer and
+                    // occasionally more time-sensitive than the group below which push data.
                     CharityUpdated::class => [TransportInterface::class],
                     StripePayout::class => [TransportInterface::class],
```
bdsl, line 353: this is sort of confusing since both queues implement TransportInterface so it isn't immediatly obvious which one this refers to. Would be clearer IMHO if we have two new interfaces to mark the two queues. Or don't use classlike names as the keys in the senders locator and just use plain strings. As suggested on the infra PR could be clearer if they're named something like "high-priority-transport" and "low-priority-transport".

app/dependencies.php
```diff
@@ -500,6 +511,23 @@ static function (ContainerInterface $c): Matching\Adapter {
...
+            if ($dsn === false) {
+                throw new \Exception('SALESFORCE_PUSH_MESSENGER_TRANSPORT_DSN must be defined in environment');
+            }
+            return $transportFactory->createTransport(
+                $dsn,
+                [],
+                new PhpSerializer(),
+            );
+        },
+
```
bdsl, line 530: as we already have an entry defined in the container for `TransportInterface`:

I think it's quite confusing to have entries for both a type its subtype defined together, returning different things. If I'm able to type-hint TransportInterface on something and have it auto-wired I think I'd assume that that means there's only one implementation of that interface available in the container. But here there are two which are equally reasonable in different applications.

It doesn't look like we do ever use these in auto-wiring so maybe better to use non-classlike keys.

And also most of the 16 lines above are duplicated from the TransportInterface and ClaimBotTransport equivalents, so I think it'd be worth extracting a factory function from the three. I think it may only be the dsn string that needs to vary between them.

src/Application/Messenger/Transport/SalesforcePushTransport.php
```diff
@@ -0,0 +1,16 @@
...
+use Symfony\Component\Messenger\Transport\TransportInterface;
+
+/**
+ * This needs to be distinct from the service we inject with `TransportInterface` generically
+ * (this app's default) so we can set it up for publishing lower priority messages which upsert
+ * Donations in Salesforce.
+ *
+ * It remains abstract so that it can be configured for either SQS or Redis in dependencies.php.
+ */
+abstract class SalesforcePushTransport implements TransportInterface
```
bdsl, line 14:
```suggestion
 */
interface SalesforcePushTransport extends TransportInterface
```

But as discussed in other comments we never actually use this as a type declaration and I don't think anything concrete class was ever going to extend or implement this at runtime so probably better to use a plain string identifier instead of a classlike.

src/Domain/DoctrineDonationRepository.php
```diff
@@ -185,6 +185,9 @@ public function findWithExpiredMatching(\DateTimeImmutable $now): array
             ->innerJoin('d.fundingWithdrawals', 'fw')
             ->where('d.donationStatus IN (:expireWithStatuses)')
             ->andWhere('d.createdAt < :expireBefore')
+            // First of a regular giving series is Pending during 3DS. If we ever make the timeout for
+            // that longer than the timeout for matching, we still want to ensure matching can't be
+            // lost while 3DS is in progress.
```
bdsl, line 190: OK - somewhat duplicates the comment on this method in the interface. That is one disadvantage of having interfaces, can be harder to see the docs.

https://github.com/thebiggive/matchbot/blob/73d6c83e0bf5f136467b400fbb87519ec6bf7270/src/Domain/DonationRepository.php#L60

NoelLH: Yeah, I think I probably wrote this before that was merged in, was just trying to include some uncommitted bits I had in-memory while I somewhat remembered why they were there. Can get rid or combine the two if you think it's better in the doc block?

app/dependencies.php
NoelLH: I'm not sure I completely understand what that would look like, and how `SendersLocator` would get a usable transport. Would we start using the plain strings as keys on the main DI container?

bdsl: Yes, there's no rule that says the lookup keys in the DI locator have to be types or class-strings. If you look at mentions of e.g. SalesforcePushTransport it's only used in dependancies.php and in contexts where all we need is a unique string - so could just be a constant:

image

If the interface was used on the right of line 514 I think that would crash with a TypeError it doesn't actually construct an instance of the interface.

src/Domain/DoctrineDonationRepository.php
bdsl: I don't have a strong feeling, could be worth combining but not a blocker.

src/Application/Messenger/Transports.php
```diff
@@ -0,0 +1,52 @@
...
+use Symfony\Component\Messenger\Transport\InMemory\InMemoryTransportFactory;
+use Symfony\Component\Messenger\Transport\Serialization\PhpSerializer;
+use Symfony\Component\Messenger\Transport\TransportFactory;
+use Symfony\Component\Messenger\Transport\TransportInterface;
+
+class Transports
+{
+    public const string TRANSPORT_CLAIMBOT = 'claimbot';
+    public const string TRANSPORT_LOW_PRIORITY = 'low_priority';
+    public const string TRANSPORT_HIGH_PRIORITY = 'high_priority';
```
bdsl, line 16: These need to be unique when used as keys for the container, so let's include the context in the strings:

```suggestion
    public const string TRANSPORT_CLAIMBOT = 'transport_claimbot';
    public const string TRANSPORT_LOW_PRIORITY = 'transport_low_priority';
    public const string TRANSPORT_HIGH_PRIORITY = 'transport_high_priority';
```

Will also help with debugging in case we see the string values somewhere.

bdsl (commented):
Added comment inline about making const string values more self-documenting & unique, otherwise all LGTM.

PR #1262, opened 2025-03-13
MAT-400: Fix transposed postal address postcode and country

bdsl:


src/Domain/CampaignRepository.php
```diff
@@ -370,8 +370,8 @@ static function (?string $string) {
             line1: $postalAddress['line1'],
             line2: $postalAddress['line2'],
             city: $postalAddress['city'],
-            postalCode: $postalAddress['country'],
-            country: $postalAddress['postalCode'],
+            postalCode: $postalAddress['postalCode'],
+            country: $postalAddress['country'],
```
bdsl, line 374: this is the actual bugfix, everything else in this PR is just clarifications.

PR #1263, opened 2025-03-13
MAT-400: Allow BG to trigger resend of donation thanks notification f…

bdsl:
…or existing donation

### DIFF
--- a/src/Domain/DonationNotifier.php
+++ b/src/Domain/DonationNotifier.php
@@ -89,12 +89,6 @@ class DonationNotifier
 
     public function notifyDonorOfDonationSuccess(Donation $donation): void
     {
-        if ($donation->isRegularGiving()) {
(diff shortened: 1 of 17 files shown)


src/Domain/DonationNotifier.php
```diff
@@ -89,12 +89,6 @@ public static function emailMessageForCollectedDonation(Donation $donation): Ema
 
     public function notifyDonorOfDonationSuccess(Donation $donation): void
     {
-        if ($donation->isRegularGiving()) {
-            // Regular giving donors get an email confirming the setup of the mandate, but not an email for
-            // each individual donation.
-            return;
-        }
-
```
bdsl, line 97: I think when we're manually choosing to (re)send a donation email there's no need to exclude regular giving donations, so this logic is moved to `DonationRepository::updateDonationStatusFromSucessfulCharge`

NoelLH (approved):

PR #1273, opened 2025-03-18
MAT-400: Switch on sending donation thanks notifications directly not…

bdsl:
… via salesforce

This is set to enable for any donations confirmed after 2.30pm today. In principle should deploy before that, although I think it may actually be fine to deploy at any point after that as well to turn on immediatly - I think I was wrong to think we need to use the collectedAt date and it's actually fine to just replace `$donation->getCollectedAt() > new \DateTimeImmutable(Donation::MAT_400_ENABLE_TIMESTAMP` with a hard-coded true.


bdsl:
> I think I was wrong to think we need to use the collectedAt date

Because the date is set on the donation record and the email request is sent to matchbot synchronously in the same process, before anything is sent to SF. It would be different if we had a process in matchbot looking at old donations and generating emails for them it would be different.

NoelLH (approved):

PR #1274, opened 2025-03-18
Bg2 2787 : Allow BG team (via Salesforce button yet to be built) to remove gift aid declaration from a donation

bdsl:


src/Application/Actions/Action.php
```diff
@@ -44,6 +51,33 @@ public function __invoke(Request $request, Response $response, $args): Response
...
+                ))))
+                ->block((new SlackSectionBlock())->text($body))
+                ->iconEmoji(':o')
+        );
+    }
+
+    protected function argToUuid(array $args, string $argName): UuidInterface
+    {
+        Assertion::keyExists($args, $argName);
+        $donationUUID = $args['' . $argName . ''];
```
NoelLH, line 72:
```suggestion
        $donationUUID = $args[$argName];
```

src/Application/Actions/Hooks/StripePaymentsUpdate.php
```diff
@@ -418,18 +418,10 @@ private function warnAboutOverRefund(
...
-                '[%s] %s',
-                (string) getenv('APP_ENV'),
-                'Over-refund detected',
-            ))))
-            ->block((new SlackSectionBlock())->text($detailsMessage))
-            ->iconEmoji(':o');
-        $chatMessage->options($options);
-
-        $this->chatter->send($chatMessage);
+        $this->chatter->send($this->prepareSlackMessage(
```
NoelLH, line 421: Nice

src/Domain/Donation.php
```diff
@@ -295,10 +302,10 @@ class Donation extends SalesforceWriteProxy
 
     /**
      * @psalm-suppress PossiblyUnusedProperty - used in DB queries
-     * @var ?DateTime   When a queued message that should lead to a Gift Aid claim was sent.
+     * @var ?DateTimeImmutable When a queued message that should lead to a Gift Aid claim was sent.
      */
     #[ORM\Column(type: 'datetime', nullable: true)]
-    protected ?DateTime $tbgGiftAidRequestQueuedAt = null;
+    protected ?DateTimeImmutable $tbgGiftAidRequestQueuedAt = null;
```
NoelLH, line 308: I see this doesn't need a DB migration because the type in the schema remains `'datetime'`. But I'm wondering if there might be a risk of confusion later from the PHP type diverging?

src/Domain/Donation.php
```diff
@@ -1806,4 +1810,49 @@ public function isRegularGiving(): bool
...
+            throw new CannotRemoveGiftAid(
+                "Cannot remove gift aid from donation {$salesforceId}, gift aid " .
+                "already removed at {$formattedQueuedDate}"
+            );
+        }
+
+        if (!$this->giftAid && !$this->tipGiftAid) {
+            throw new CannotRemoveGiftAid(
+                "Cannot remove gift aid from donation {$salesforceId}, gift aid " .
+                "was never claimed"
```
NoelLH, line 1846:
```suggestion
                "eligibility was never reported by donor"
```
Just trying to disambiguate as I think "claimed" could imply the other meaning.

tests/Domain/DonationTest.php
```diff
@@ -1154,4 +1155,27 @@ public function createWithdrawal(FundType $fundType, string $fundAmount): Fundin
...
+        $donation->setTbgGiftAidRequestQueuedAt(new \DateTimeImmutable('2025-01-01T00:00:00Z'));
+
+        $this->expectException(CannotRemoveGiftAid::class);
+        $this->expectExceptionMessage(
+            "Cannot remove gift aid from donation donationId12345678, request already queued to send to HMRC at 2025-01-01 00:00"
+        );
+        $donation->removeGiftAid(new \DateTimeImmutable());
+    }
+
+    public function testCannotRemoveGiftAid(): void
```
NoelLH, line 1173:
```suggestion
    public function testCanRemoveGiftAid(): void
```

NoelLH (approved):
LGTM. Minor things inline, no blockers

src/Domain/Donation.php
bdsl: I don't think the PHP type is diverging from anything is it? The DB doesn't have a distinction between DateTime and DateTimeImmutable and doesn't need one since it doesn't have reference types.

NoelLH: The DB engine itself doesn't but Doctrine puts a `DC2Type` comment on the field. I don't really know what it uses that for so I'm not sure if we should worry.

src/Domain/Donation.php
bdsl: > 
```suggestion
                "was not requested by donor"
```

src/Domain/Donation.php
bdsl: Right let me see what that's about and if the diff command will generate a migration...

bdsl: I tried running the diff command and I don't think it does anything. Reading https://www.doctrine-project.org/projects/doctrine-dbal/en/3.7/explanation/dc2type-comments.html I think we're best ignoring the DC2Type comments. 
They don't seem to do anything for us. They could be useful if we wanted to define our SQL schema first and then have the PHP classes autogenerated. They'll be removed when we upgrade to DBAL 4.

src/Domain/Donation.php
bdsl: Thanks, I made a slightly different change to the message.

src/Domain/Donation.php
NoelLH: It does nothing in this case because of the `type: 'datetime'` though I think. But fair enough, sounds fine to ignore.

bdsl: My understanding is that these comments do nothing in any case - DBAL 3 just generates them for backward compatibility with DBAL 2. DBAL 4 won't generate them at all. The ORM doesn't use them, it needs all mapping definitions to be in the class file (or xml mapping or whatever).

See discussion here https://github.com/doctrine/migrations/issues/1441

PR #1277, opened 2025-03-21
MAT-410 – support pledges becoming Topup ones after first load

NoelLH:


src/Domain/FundType.php
```diff
@@ -37,4 +37,9 @@ public function allocationOrder(): int
         // check tests cover match above.
         return $order * 100;
     }
+
+    public function isPledge(): bool
+    {
+        return $this === self::Pledge || $this === self::TopupPledge;
```
bdsl, line 43: I think it's worth using match so we can't forget to update this if we ever have any other type of pledge:
```suggestion
        return match ($this) {
            self::Pledge, self::TopupPledge => true,
            self::ChampionFund => false,
        };
```

src/Domain/FundRepository.php
```diff
@@ -148,6 +148,19 @@ protected function setAnyFundData(Fund $fund, array $fundData): Fund
...
+
+        // For now, we let some charities collect non-Topup pledges with the Salesforce form, only
+        // after champion funds are used. So the mid-campaign matching behaviour is "usually" unaffected
+        // by the Topup status (with possible exceptions if e.g. there are refunds that release champion funds).
+        // At the end of a campaign however we want to treat them as Topups during redistribution. So until we have
+        // a form that gets it right from the start, we allow a type change from Pledge to TopupPledge based on
+        // Salesforce Pledge__c record edits.
+        /** @var string $type */
+        $type = $fundData['type'];
+        if ($fund->hasBeenPersisted() && $type !== 'pledge' && $fund->getFundType() === FundType::Pledge) {
```
bdsl, line 160: not sure what the hasBeenPersisted test is about?

Also do we need any of the other checks on this line? If the type is 'topupPledge` then the changeType will be a harmless no-op. If it's `championFund` then `changeType` will throw and set off an alarm. Don't we want an alarm if the fund in SF is in a state that matchbot can't accept.

As were just talking about designating one side or the other to be authorities for different entities: SF is the authority for funds which means the most useful place to put a constraint is probably a Salesforce before update trigger. But then it's till worth having the check in matchbot as well for defense-in-depth.

NoelLH: I was thinking that we only want to consider changes for funds that aren't brand new, but I agree it's redundant.

I think that throwing on a change to champion fund, which should be impossible unless something very weird happened, would be OK. But I'm not sure about throwing if e.g. somebody unticks Is Topup? on a pledge in error. I think my order of preference of what could happen from best to worst is:
* error log it but continue (maybe a future optimisation?)
* ignore it (PR now)
* crash entirely – least desirable as it might block other fund updates we are relying upon, and the impact of drift for one record is likely to be less

bdsl: Is there or can there be a trigger to stop Is Topup being unticked so we don't end up with different data in the two systems? Or are you thinking that the window between changes being made in SF and pulled to matchbot is useful to allow undoes in case we accidentally change in the allowed direction?

NoelLH: Yeah, I think preventing those changes after campaigns launch could be good, but is a little involved. It came up in the comments on BG2-2838 but I think we left it at history tracking for now. Preventing it in general is too aggressive for some campaign types I think, it should be possible to correct mistakes before MatchBot knows about the pledges.

bdsl: IMHO we should at generate an alarm, I think it's quite a serious problem that needs to be fixed if the fund in matchbot won't behave the way the person who edited it in SF will expect it to because we refused to pull the update.

Agree we shouldn't block the other updates - if we throw then we'd probably want to catch and continue to the next iteration of this loop https://github.com/thebiggive/matchbot/blob/a6322932f419d2a4b231a22284bd0583ba1e26ad/src/Domain/FundRepository.php#L46

PR #1279, opened 2025-03-24
MAT-410: Update allocation order on all related campaign fundings whe…

bdsl:
…n changing fund type

Still need to look at how to eager load the campaignFundings for performance


bdsl:
Closed as agreed we're going to try a different approach

PR #1282, opened 2025-03-25
MAT-410 – store allocation order on Funds too

NoelLH:
Prepare for removing reliance on a per-CampaignFunding copy of the same information, shortly

### DIFF
--- a/integrationTests/IntegrationTest.php
+++ b/integrationTests/IntegrationTest.php
@@ -298,9 +298,9 @@ abstract class IntegrationTest extends TestCase
 
         $db->executeStatement(<<<SQL
             INSERT INTO Fund (name, salesforceId, salesforceLastPull, createdAt, updatedAt, fundType,
-                              currencyCode) VALUES 
+                              currencyCode, allocationOrder) VALUES 
                 ('Some test fund', '$fundSfID', '$nyd', '$nyd', '$nyd', '{$fundType->value}',
-                 'GBP')
+                 'GBP', 1)
         SQL
         );
 
--- a/src/Domain/CampaignFunding.php
+++ b/src/Domain/CampaignFunding.php
@@ -20,7 +20,7 @@ use Doctrine\ORM\Mapping as ORM;
  * See {@see FundType::allocationOrder()}
  */
 #[ORM\Table]
-#[ORM\Index(name: 'available_fundings', columns: ['amountAvailable', 'allocationOrder', 'id'])]
+#[ORM\Index(name: 'available_fundings', columns: ['amountAvailable', 'id'])]
 #[ORM\Entity(repositoryClass: CampaignFundingRepository::class)]
 #[ORM\HasLifecycleCallbacks]
 class CampaignFunding extends Model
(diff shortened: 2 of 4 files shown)


integrationTests/IntegrationTest.php
```diff
@@ -298,9 +298,9 @@ public function addFunding(
 
         $db->executeStatement(<<<SQL
             INSERT INTO Fund (name, salesforceId, salesforceLastPull, createdAt, updatedAt, fundType,
-                              currencyCode) VALUES 
+                              currencyCode, allocationOrder) VALUES 
                 ('Some test fund', '$fundSfID', '$nyd', '$nyd', '$nyd', '{$fundType->value}',
-                 'GBP')
+                 'GBP', 1)
```
bdsl, line 303: Probably worth having this a bit more realistic with 
```suggestion
                 'GBP', {$fundType->getAllocationOrder()})
```

src/Domain/CampaignFunding.php
```diff
@@ -156,6 +158,11 @@ public function getCampaigns(): array
 
     public function getAllocationOrder(): int
     {
+        if ($this->getFund()->getAllocationOrder() > 0) {
```
bdsl, line 161: 0 is the implicit default set for MySQL right? Makes sense.

bdsl (approved):

src/Migrations/Version20250325123754.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        return 'Prepare schema for new allocation order method';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('DROP INDEX available_fundings ON CampaignFunding');
+        $this->addSql('CREATE INDEX available_fundings ON CampaignFunding (amountAvailable, id)');
+        $this->addSql('ALTER TABLE Fund ADD allocationOrder INT NOT NULL');
+        $this->addSql('CREATE INDEX allocationOrder ON Fund (allocationOrder)');
```
bdsl, line 25: not tested but I think if you add this you don't need the expand/contract pattern:

```suggestion
        $this->addSql('CREATE INDEX allocationOrder ON Fund (allocationOrder)');
        
        $this->addSql(<<<'SQL'
            UPDATE Fund SET allocationOrder = CASE
                WHEN fundType = 'pledge' THEN 100
                WHEN fundType = 'championFund' THEN 200
                WHEN fundType = 'TopupPledge' THEN 300
            END
        SQL);
```

NoelLH: Thanks, yep I think you're right. Sorry I forgot this suggestion and tried to do it with a join in a new PR at https://github.com/thebiggive/matchbot/pull/1283/files#diff-f86833cd06ef4cd40ca911051395ea608ed85db2d86cd250c979872f91d78348R22-R28

I'm not opposed to keeping it all on `Fund` although I think the case above is not right for 'topupPledge'. Not sure which way is safer really, it's hard to test usefully and also do it all on one PR.

NoelLH:
Superseded by / folded into https://github.com/thebiggive/matchbot/pull/1283

PR #1285, opened 2025-03-26
Regression alarm fix: Always delay sending succesful donations to SF …

bdsl:
…by 3 seconds

Should make it very unlikely that SF tries to process the same donation in its sucessful state before it deals with its pending state. Pushing them in the wrong order generates an alarm.


bdsl:
After writing this I did think maybe it would have been better done as a middleware, but what I've got here should work.

PR #1288, opened 2025-03-26
Regression alarm fix: Clear entity manager in all cases before proces…

bdsl:
…sing each message queue message

I think even better than clearing the EM would be to replace it entirely, since afaik there's no gurantee that a previous message handler didn't close it. But that would pretty much require rebuilding our whole service container graph, so not going to try to do that now.


app/dependencies.php
```diff
@@ -127,13 +128,23 @@ function (ContainerInterface $c): Auth\DonationPublicAuthMiddleware {
             $messengerReceiverLocator->set($priorityKey, $c->get(Transports::TRANSPORT_HIGH_PRIORITY));
             $messengerReceiverLocator->set($salesforceKey, $c->get(Transports::TRANSPORT_LOW_PRIORITY));
 
+            $eventDispatcher = new EventDispatcher();
+
+            $eventDispatcher->addListener(WorkerMessageReceivedEvent::class, function () use ($c) {
+                // clear the entity manager before handling each message to make sure we get up-to-date copies
+                // of any entities - otherwise they could be up to an hour out of date as the consumer is a long-
```
NoelLH, line 135: I'm not sure I see where one hour comes from? Typical workers currently live up to 24 hours but maybe there's a cache lifetime default I'm missing?

NoelLH (approved):
Only hoping to clarify a comment, LGTM

app/dependencies.php
bdsl: You're right it is 24 hours, I somehow read saw 86400 here and thought of 3600 instead. https://github.com/thebiggive/matchbot/blob/95f59cc97c66c80dc7bc821eb536d2e9b962c6fb/composer.json#L217

bdsl, line 135:
```suggestion
                // of any entities - otherwise they could be up to a day out of date as the consumer is a long-
```

bdsl: Fixed in [64fd062](https://github.com/thebiggive/matchbot/pull/1288/commits/64fd0628a0f3c15ab658eb2c14ae981be87e4ad3) , thanks @NoelLH

PR #1290, opened 2025-03-27
BG2-2854 – remove Gift Aid from one donation where the box was ticked in error

NoelLH:


src/Migrations/Version20250327103803.php
```diff
@@ -0,0 +1,42 @@
...
+         * @var array{uuid: string, transactionId: string}
+         */
+        $idPair = [
+            'uuid' => '01f1edad-1f3b-4441-a21c-6c031eda9ee8',
+            'transactionId' => 'pi_3R5k5KKkGuKkxwBN0SIhyZre',
+        ];
+
+        $this->addSql(<<<EOT
+            UPDATE Donation
+            SET salesforcePushStatus = 'pending-update', giftAid = 0, tipGiftAid = 0
```
bdsl, line 32: not essential but probably worth doing from now on:
```suggestion
            SET salesforcePushStatus = 'pending-update', giftAid = 0, tipGiftAid = 0, giftAidRemovedAt = now(),
```

bdsl (approved):
One suggestion inline, not a blocker. I've double checked the transaction ID and UUID against the case and the prod DB.

src/Migrations/Version20250327103803.php
NoelLH: Thanks, done in c62c0c571ebd71d42934939ab6b156cbb4ce179c as I wasn't sure if the engine would deal with an extra comma

bdsl: ah yep no I don't think mysql does allow trailing commas.

PR #1292, opened 2025-03-28
Get `Donation.tbgGiftAidRequestQueuedAt` Doctrine & PHP types back in sync

NoelLH:


bdsl (approved):
LGTM, I'm not sure but I think maybe the problem is just the datetime type on 307 and all we needed is this - Doctrine can read the type from line 308

#[ORM\Column()]

I don't think the db migration will make any difference but it won't do any harm either.

PR #1297, opened 2025-04-09
 ID-47: Optionally recieve email verification tokens from Identity and…

bdsl:
… send to Mailer as part of donation confirmation emails

This doesn't do anything user visible yet but means when Idnentity starts sending these tokens,
and mailer starts outputting them, then the donors will see the them.

Before we're ready for donors to see them we need to build a place for
the donor to use them, and then very shortly after block creating an
account without one.

### DIFF
--- a/src/Domain/EmailVerificationTokenRepository.php
+++ b/src/Domain/EmailVerificationTokenRepository.php
@@ -17,10 +17,10 @@ class EmailVerificationTokenRepository
     {
         $emailVerificationToken = $this->em->createQuery(
             dql: <<<'DQL'
-                SELECT t from Matchbot\Domain\EmailVerificationToken t 
+                SELECT t FROM MatchBot\Domain\EmailVerificationToken t
                 WHERE t.emailAddress = :email
                 AND t.createdAt > :created_since
-                ORDER BY createdAt DESC 
+                ORDER BY t.createdAt DESC
                 DQL
         )->setParameters([
             'email' => $emailAddress->email,
--- a/integrationTests/EmailVerificationTokenRepositoryTest.php
+++ b/integrationTests/EmailVerificationTokenRepositoryTest.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace MatchBot\IntegrationTests;
+
+use MatchBot\Domain\EmailAddress;
+use MatchBot\Domain\EmailVerificationTokenRepository;
+
+class EmailVerificationTokenRepositoryTest extends IntegrationTest
+{
+    public function testYesWeHaveNoTokens(): void
+    {
+        $sut = $this->getService(EmailVerificationTokenRepository::class);
+
+        $token = $sut->findRecentTokenForEmailAddress(
+            EmailAddress::of('email@example.com'),
+            new \DateTimeImmutable()
+        );
+
+        $this->assertNull($token);
+    }
+}


src/Domain/EmailVerificationTokenRepository.php
```diff
@@ -0,0 +1,34 @@
...
+     */
+    public function __construct(private EntityManagerInterface $em)
+    {
+    }
+
+    public function findRecentTokenForEmailAddress(EmailAddress $emailAddress, \DateTimeImmutable $at): ?EmailVerificationToken
+    {
+        $emailVerificationToken = $this->em->createQuery(
+            dql: <<<'DQL'
+                SELECT t from Matchbot\Domain\EmailVerificationToken t
```
bdsl, line 20: I'm getting an error when I run this on my local (shown in docker logs when the callback happens on donation success from Stripe), but I haven't been able to work out what's wrong:

```
[2025-04-09T11:51:20.960548+00:00] matchbot.ERROR: Doctrine\ORM\Query\QueryException: 
[Semantical Error] line 0, col 14 near 'Matchbot\Domain\EmailVerificationToken': Error: Class 
'Matchbot\Domain\EmailVerificationToken' is not defined.  
#  /var/www/html/vendor/doctrine/orm/src/Query/QueryException.php:45   #0 
```

The class `Matchbot\Domain\EmailVerificationToken` does exist, added in this PR, so I don't know why Doctrine doesn't seem to be able to find it.

NoelLH (approved):
LGTM

src/Migrations/Version20191110162524.php
```diff
@@ -22,7 +22,7 @@ public function up(Schema $schema) : void
         // this up() migration is auto-generated, please modify it to your needs
         $this->abortIf(! $this->connection->getDatabasePlatform() instanceof \Doctrine\DBAL\Platforms\AbstractMySQLPlatform, 'Migration can only be executed safely on \'mysql\'.');
 
-        $this->addSql('CREATE TABLE CommandLockKeys (key_id VARCHAR(64) NOT NULL, key_token VARCHAR(44) NOT NULL, key_expiration INT UNSIGNED NOT NULL, PRIMARY KEY(key_id)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_bin` ENGINE = InnoDB');
+        $this->addSql('CREATE TABLE IF NOT EXISTS CommandLockKeys (key_id VARCHAR(64) NOT NULL, key_token VARCHAR(44) NOT NULL, key_expiration INT UNSIGNED NOT NULL, PRIMARY KEY(key_id)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_bin` ENGINE = InnoDB');
```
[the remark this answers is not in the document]
bdsl: It happens to me a lot on my local - I wipe the database, then run a command that automatically creates the CommandLockKeys table, then try to run migrations and they fail because the table already exists.

bdsl:
Thanks for reviewing @NoelLH . Do you have any ideas on this? https://github.com/thebiggive/matchbot/pull/1297#discussion_r2035208208

src/Domain/EmailVerificationTokenRepository.php
bdsl: Fixed in https://github.com/thebiggive/matchbot/pull/1297/commits/ab0ba2e04a9f901006332f4e3a3268e6c033a2bd, main issue was case of MatchBot in class name. Probably would have worked on Mac but not on Linux.

PR #1304, opened 2025-04-11
ID-47 – donor thanks mail: send create account URI

NoelLH:


.env.example
```diff
@@ -50,6 +50,8 @@ SALESFORCE_SECRET_KEY=topsecret
 MAILER_BASE_URI=https://mailer-staging.thebiggivetest.org.uk
 MAILER_SEND_SECRET=
 
+ACCOUNT_MANAGEMENT_BASE_URI=http://localhost:4200
```
NoelLH, line 53: Will need corresponding var in infrastructure

bdsl: I'm not sure it does - the base URI only changes rarely, I think the way we've managed these links in Donate or Wordpress seems to have worked well, i.e. having one hardcoded for each of three envs, local, staging, regression, and prod. 

How about adding a function to get this URI on the Environment enum https://github.com/thebiggive/matchbot/blob/4c2fff3f2bc830ac10c6a0afcc6f1482f395b59b/src/Application/Environment.php ?

src/Domain/DonorAccountRepository.php
```diff
@@ -25,6 +25,20 @@ public function findByStripeIdOrNull(StripeCustomerId $stripeAccountId): ?DonorA
         return $this->findOneBy(['stripeCustomerId.stripeCustomerId' => $stripeAccountId->stripeCustomerId]);
     }
 
+    public function shouldInviteRegistration(Donation $donation): bool
```
bdsl, line 28: this doesn't seem like a persistence focused function, I'm not sure a repository is the right place for it. Will have a look to see if I can find somewhere it might fit better.

src/Domain/DonorAccountRepository.php
```diff
@@ -25,6 +25,20 @@ public function findByStripeIdOrNull(StripeCustomerId $stripeAccountId): ?DonorA
...
+        $stripeCustomerId = $donation->getPspCustomerId();
+        if (!$stripeCustomerId) {
+            // We don't really expect missing Customer ID at this point for new donors, but seems
+            // marginally more logical to treat this like a missing DonorAccount(?)
+            return true;
+        }
+
+        // Identity sends key info for MB DonorAccount iff a password was set via \Messages\Person, so
+        // we can use record existence to decide whether to send a register link.
+        return $this->findByStripeIdOrNull($stripeCustomerId) === null;
```
bdsl, line 39: probably will be more straightfoward to look up by the donor id once that's recorded on the donation (pr to do that https://github.com/thebiggive/matchbot/pull/1302 was just merged)

bdsl (dismissed):
Added comments inline, just used request changes feature for the feature-flag, see https://github.com/thebiggive/matchbot/pull/1304/files#r2039846868

src/Domain/DonorAccountRepository.php
bdsl: I looked at the usages - one is in [ResendDonorThanksNotification.php](https://github.com/thebiggive/matchbot/pull/1304/files#diff-508d7d8a3ea3f8f2886b4dbf95aa0353c2bb8e21be7353b6e9a08297badae64b) but for now we won't have a fresh token to include in the resend email so it should just be false there. The other usage is in donation service so I'd suggest making a function inside donationService to replace this.

I haven't tested but think this would work:

```php
    private function shouldInviteRegistration(Donation $donation): bool
    {
        $donorId = $donation->getDonorId();
        if ($donorId === null) {
            return false;
        }
        
        $existingDonorAccount = $this->donorAccountRepository->findByPersonId($donorId);

        return (is_null($existingDonorAccount));
    }
```

Could make it a lot more optimal by just asking the repo to tell us whether the account exists or not instead of actually loading the whole thing from the db.

.env.example
bdsl: If you do think we need the env var could we make it something like `DONATE_BASE_URI` instead of `ACCOUNT_MANAGEMENT_BASE_URI` , so we can re-use it in future if we want to generate links to e.g. campaigns.

bdsl: Thinking again there is some value in having it in env vars as it would be shared with `donate_default_uri_prefix` used by identity. Maybe worth setting up vars for `experienceUriPrefix` and `blogUriPrefix` at the same time and passing them to all our main services, e.g. so we can use them for presenting links directly in mailer in future.

src/Domain/DonationNotifier.php
```diff
@@ -97,15 +111,21 @@ public static function emailMessageForCollectedDonation(
...
         $emailAddress = $donation->getDonorEmailAddress();
         Assertion::notNull($emailAddress);
-        $emailVerificationToken = $this->emailVerificationTokenRepository->findRecentTokenForEmailAddress(
-            $emailAddress,
-            $this->now,
-        );
-        $emailMessage = self::emailMessageForCollectedDonation($donation, $emailVerificationToken);
+
+        $emailVerificationToken = null;
+        if ($sendRegisterUri) {
```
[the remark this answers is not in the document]
NoelLH: Thanks, done as part of 06b4e91614c55e21dbb1063980ef1c2c251ce18e

src/Domain/DonorAccountRepository.php
bdsl: And if we do want to include a registration link in resent donation emails that's going to be a bit awkward to implement - we're auto deleting Person records in identity if they don't have passwords set after 32 hours, so we'd have to either keep them longer or recreate them from information on the donation record. We also need to treat the verification tokens as short lived so we'd need to have Identity generate a new one and then matchbot include that in the email.

Probably could be done if we put the glue code into SF but probably we'll just tell the donor to create an account manually, and we can implement an option to see non-logged in donations if you have an account with verified email address.

.env.example
bdsl: If you make an change in infrastructure that will take us to code in **7** repos modified for one story.

src/Domain/DonorAccountRepository.php
bdsl: I also realised we shouldn't invite registration if there's already any account registered with the same email address. We could test that here but better to just not generate the token in the first place in that case - so I've made a PR for that in identity: https://github.com/thebiggive/identity/pull/204

bdsl (approved):

PR #1310, opened 2025-04-15
MAT-415 – add current campaigns fund totals push to Salesforce

NoelLH:
Expected to be scheduled daily

### DIFF
--- a/src/Application/Commands/PushDailyFundTotals.php
+++ b/src/Application/Commands/PushDailyFundTotals.php
@@ -4,7 +4,6 @@ namespace MatchBot\Application\Commands;
 
 use MatchBot\Application\Messenger\FundTotalUpdated;
 use MatchBot\Domain\FundRepository;
-use Psr\Log\LoggerInterface;
 use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
@@ -22,7 +21,6 @@ class PushDailyFundTotals extends LockingCommand
 {
     public function __construct(
         private readonly FundRepository $fundRepository,
-        private readonly LoggerInterface $logger,
         private readonly RoutableMessageBus $bus,
     ) {
         parent::__construct();
@@ -30,7 +28,7 @@ class PushDailyFundTotals extends LockingCommand
 
     public function doExecute(InputInterface $input, OutputInterface $output): int
     {
-        $funds = $this->fundRepository->findForCampaignsOpenAt(new \DateTime('now'));
+        $funds = $this->fundRepository->findForCampaignsOpenAt(new \DateTimeImmutable('now'));
         foreach ($funds as $fund) {
             if ($fund->getFundType()->isPledge()) {
                 // Skip pledges from daily total push as Salesforce doesn't yet do anything with that info
@@ -41,7 +39,7 @@ class PushDailyFundTotals extends LockingCommand
             $this->bus->dispatch(new Envelope(FundTotalUpdated::fromFund($fund)));
         }
 
-        $this->logger->info(sprintf('Pushed %d fund totals to Salesforce for open campaigns', count($funds)));
+        $output->writeln(sprintf('Pushed %d fund totals to Salesforce for open campaigns', count($funds)));
 
         return 0;
     }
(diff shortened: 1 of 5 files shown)


src/Application/Commands/PushDailyFundTotals.php
```diff
@@ -0,0 +1,48 @@
...
+        private readonly FundRepository $fundRepository,
+        private readonly LoggerInterface $logger,
+        private readonly RoutableMessageBus $bus,
+    ) {
+        parent::__construct();
+    }
+
+    public function doExecute(InputInterface $input, OutputInterface $output): int
+    {
+        $funds = $this->fundRepository->findForCampaignsOpenAt(new \DateTime('now'));
```
bdsl, line 33:
```suggestion
        $funds = $this->fundRepository->findForCampaignsOpenAt(new \DateTimeImmutable('now'));
```

(will need the matching change in repository). Worth treating DateTime as deprecated IMHO, especially since it somewhat dangerously has methods with the same names as the DateTimeImmutable that mutate it, so if we don't remember which one we have we could mutate by mistake.

I'd also inject the timestamp via the constructor but unless you're going to write a test against this class I don't think that's so important.

src/Domain/FundRepository.php
```diff
@@ -214,6 +214,31 @@ public function findForCampaignsClosedSince(DateTime $closedBeforeDate, DateTime
...
+            WHERE
+                campaign.startDate < :openAtDate AND
+                campaign.endDate > :openAtDate
+            GROUP BY fund
+        EOT;
+
+        /** @var Fund[] $result */
+        $result = $this->getEntityManager()->createQuery($query)
+            ->setParameter('openAtDate', $openAtDate)
+            ->disableResultCache()
```
bdsl, line 236: Is there a particular reason for disableResultCache here? Does this query need up to date results more so than other queries?

src/Application/Commands/PushDailyFundTotals.php
```diff
@@ -0,0 +1,48 @@
...
+        foreach ($funds as $fund) {
+            if ($fund->getFundType()->isPledge()) {
+                // Skip pledges from daily total push as Salesforce doesn't yet do anything with that info
+                // and it could be a lot of requests against a volume-limited Site.
+                continue;
+            }
+
+            $this->bus->dispatch(new Envelope(FundTotalUpdated::fromFund($fund)));
```
bdsl, line 41: This is going to cause this function to run once for each non-pledge fund: https://github.com/thebiggive/matchbot/blob/12a8c1fe42d721dad72ee50a29f640266176ac19/src/Client/Fund.php#L49

which I think will then run the `setUsedAmount` function for each one in SF.

I'm not sure exactly what the limitation in SF is, but I wonder if we could make it a lot more efficient if we had a function in SF we could use that would set the usedAmount for many funds, and send it a map from fund id to amount. Probably something to think about for deploying after GMF, but if that works could that mean we update the funds much more often than once a day?

bdsl (commented):
Comments added inline, I think we should change the date type, otherwise all seems fine.

src/Domain/FundRepository.php
```diff
@@ -214,6 +214,31 @@ public function findForCampaignsClosedSince(DateTime $closedBeforeDate, DateTime
         return $result;
     }
 
+    /**
+     * @param DateTime $openAtDate Typically now
+     * @return Fund[]
+     */
+    public function findForCampaignsOpenAt(DateTime $openAtDate): array
```
bdsl, line 221:
```suggestion
    /**
     * @param \DateTimeImmutable $openAtDate Typically now
     * @return Fund[]
     */
    public function findForCampaignsOpenAt(\DateTimeImmutable $openAtDate): array
```

src/Domain/FundRepository.php
NoelLH: Just trying to maintain similarity with the closest existing analogue, until we pick up the task to look more closely at caches and whether it might actually be doing anything useful.

src/Application/Commands/PushDailyFundTotals.php
NoelLH: Certainly an option to build but I don't know if the total bandwidth would be that different or if there's a need to have this more often than daily and on close, nobody has mentioned a more frequent need.

src/Application/Commands/PushDailyFundTotals.php
NoelLH: Type change done in https://github.com/thebiggive/matchbot/pull/1310/commits/5fef1730299dba20e1d8b21c61d282a50260aed2

bdsl (approved):

PR #1317, opened 2025-04-17
Remove assertion about non-first charity address lines

NoelLH:
Allow unusual data to be treated as a fully null address without a crash, for now.

### DIFF
--- a/src/Domain/CampaignRepository.php
+++ b/src/Domain/CampaignRepository.php
@@ -361,9 +361,11 @@ class CampaignRepository extends SalesforceReadProxyRepository
             $postalAddress
         );
 
+        // For now, treat whole address as null if there's no `line1`. This can happen with allowed
+        // Salesforce addresses for now. For example, a charity may fill in just a country in the
+        // portal and save that as their own address. We're better off omitting it from donor emails
+        // if that happens.
         if (is_null($postalAddress['line1'])) {
-            Assertion::allNull($postalAddress);
-
             return PostalAddress::null();
         }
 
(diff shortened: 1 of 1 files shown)


src/Domain/CampaignRepository.php
```diff
@@ -361,9 +361,11 @@ static function (?string $string) {
             $postalAddress
         );
 
+        // For now, treat whole address as null if there's no `line1`. This can happen with allowed
+        // Salesforce addresses for now. For example, a charity may fill in just a country in the
+        // portal and save that as their own address. We're better off omitting it from donor emails
+        // if that happens.
```
bdsl, line 367: It's a bit making me wish we'd saved the address as one string or an list of strings instead of named fields. I don't really like ignoring data if our UI accepted it without complaint. Wondering if we should make this a validation rule on the charity object. Or we just allow line1 to be null in Matchbot. Any falsy fields already get filtered out when the address is displayed. 

We shouldn't assume this is just for donor emails, that's the current use but there's nothing to say that in the code nearby, it could equally be for display via frontend.

bdsl: @NoelLH what do you think about keeping the assert but catching the exception and making it only generate an alarm for now? Then if we get too many of those alarms we can put a validation rule on the campaign in SF?

NoelLH: Ah yep that is probably better, good idea

NoelLH: Done in 4371bee65e2c6510e39c0961c15f14629a1b8595

bdsl (approved):

PR #1321, opened 2025-04-22
Increase the maximum Money value to just under 10 billion major units

NoelLH:


bdsl (approved):
No objection because the new value is bigger but doesn't seem qualitatively different to what we had before, but looking at this assertion again does make me think it would be better to have a limit here based on the scale of charities we deal with in reality - e.g. we've raised about £350,000,000 total ever so maybe something like £1,000,000,000 as the next big round number.

NoelLH:
> No objection because the new value is bigger but doesn't seem qualitatively different to what we had before, but looking at this assertion again does make me think it would be better to have a limit here based on the scale of charities we deal with in reality - e.g. we've raised about £350,000,000 total ever so maybe something like £1,000,000,000 as the next big round number.

Yep, as we discussed I think if our tests are not getting close to using these biggest funds up – as seems likely – then we could refine our regression test sandbox data and maybe Salesforce validation rules as a subsequent improvement.

PR #1328, opened 2025-04-24
MAT-417: Correct details on one donation

bdsl:
I've tested this by manually copying both the source row from the DonorAccount table and the to-be-corrected uuid from the donation table into my local machine database, and running the migration does update the donation as I expected


NoelLH (approved):
Reviewed alongside Production data, LGTM

PR #1330, opened 2025-04-25
Log more quietly after 3DS fails, then there's an unsuccessful extra Confirm action call

NoelLH:
Seen occasionally during GMF25

### DIFF
--- a/src/Domain/DonationService.php
+++ b/src/Domain/DonationService.php
@@ -59,6 +59,9 @@ class DonationService
         'The provided PaymentMethod has failed authentication',
         'You must collect the security code (CVC) for this card from the cardholder before you can use it',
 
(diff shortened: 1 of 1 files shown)


src/Domain/DonationService.php
```diff
@@ -59,6 +59,9 @@ class DonationService
         'The provided PaymentMethod has failed authentication',
         'You must collect the security code (CVC) for this card from the cardholder before you can use it',
 
+        // Donor ignoring 3DS then reusing the donation seems to occasionally get this
```
bdsl, line 62:
```suggestion
        // Donor ignoring 3DS then reusing the donation seems to occasionally get this
        // see ticket DON-1140
```

bdsl (approved):
Added a comment suggestion, LGTM

PR #1332, opened 2025-04-25
MAT-418 – fix `PersonId` embeddable's UUID type

NoelLH:
Previously, the UUID library would work because the actual database value is a string, but would find needless changes on mandates every time they were in the unit of work

### DIFF
--- a/src/Domain/PersonId.php
+++ b/src/Domain/PersonId.php
@@ -4,7 +4,6 @@ namespace MatchBot\Domain;
 
 use Doctrine\ORM\Mapping\Column;
 use Doctrine\ORM\Mapping\Embeddable;
-use MatchBot\Application\Assertion;
 use Ramsey\Uuid\Uuid;
 use Ramsey\Uuid\UuidInterface;
 
@@ -16,13 +15,11 @@ use Ramsey\Uuid\UuidInterface;
 class PersonId
 {
     #[Column(type: 'uuid')]
-    public readonly string $id;
+    public readonly UuidInterface $id;
 
-    private function __construct(
-        string $personId
-    ) {
-        $this->id = $personId;
-        Assertion::uuid($personId);
+    private function __construct(string $personId)
+    {
+        $this->id = Uuid::fromString($personId);
     }
 
     public static function of(string $personId): self


src/Domain/PersonId.php
```diff
@@ -16,13 +15,11 @@
 class PersonId
 {
     #[Column(type: 'uuid')]
```
bdsl, line 17:
```suggestion
    #[Column]
```

bdsl (approved):
Thanks, looks like the original error was having different types specified for \MatchBot\Domain\PersonId::$id in the PHP type system and the Doctrine system. 

Added a suggestion to remove the doctrine type as it looks like Doctrine can read the type from the PHP type declaration to reduce the chance of having a misalignment again here or if we copy the code to use somewhere else.

bdsl:
I just checked, the `DonorAccount.uuid` is now non-null in all rows in prod, so we probably could make it non-null in the model which means we could use the embeddable directly instead of pulling the UUID out of it.

NoelLH:
> Thanks, looks like the original error was having different types specified for \MatchBot\Domain\PersonId::$id in the PHP type system and the Doctrine system.
> 
> Added a suggestion to remove the doctrine type as it looks like Doctrine can read the type from the PHP type declaration to reduce the chance of having a misalignment again here or if we copy the code to use somewhere else.

Yep, and this was maybe masked from psalm by the removed assertion in the constructor  – I'm not entirely clear why that wasn't failing at runtime actually.

It seems like removing the `type` from the attribute makes Doctrine want to define the column differently. I'll look into whether that seems like a good idea now.

NoelLH:
> It seems like removing the `type` from the attribute makes Doctrine want to define the column differently. I'll look into whether that seems like a good idea now.

It wants to do this:
`$this->addSql('ALTER TABLE RegularGivingMandate CHANGE personid personid VARCHAR(255) NOT NULL');`

whereas with `type` in the attribute it's a `CHAR(36)` with a specific comment showing that Doctrine knows it's a uuid.

I think that because we are adding a non-built-in type, inference from the class doesn't work.
`Type::addType('uuid', Ramsey\Uuid\Doctrine\UuidType::class);`

https://github.com/ramsey/uuid-doctrine?tab=readme-ov-file#mappings has an example that includes `type` and the docs for `addType()` suggest the `name` is supposed to match the type's name, so I don't think switching the name for a class name string would work.

bdsl:
> > It seems like removing the `type` from the attribute makes Doctrine want to define the column differently. I'll look into whether that seems like a good idea now.
> 
> It wants to do this: `$this->addSql('ALTER TABLE RegularGivingMandate CHANGE personid personid VARCHAR(255) NOT NULL');`
> 
> whereas with `type` in the attribute it's a `CHAR(36)` with a specific comment showing that Doctrine knows it's a uuid.
> 
> I think that because we are adding a non-built-in type, inference from the class doesn't work. `Type::addType('uuid', Ramsey\Uuid\Doctrine\UuidType::class);`
> 
> https://github.com/ramsey/uuid-doctrine?tab=readme-ov-file#mappings has an example that includes `type` and the docs for `addType()` suggest the `name` is supposed to match the type's name, so I don't think switching the name for a class name string would work.

Fair enough - I ran the migrations diff command on my local but didn't get as far as running the validate command.

PR #1183, opened 2025-01-20
DON-998: Misc regular giving improvments

bdsl:
…onation

Used in DonationService currently, to be re-used in RegularGivingService

PR #1333, opened 2025-04-25
DON-1140 – remove extraneous re-setting of `capture_method` on Payment Intent confirmation

NoelLH:


src/Domain/DonationService.php
```diff
@@ -74,12 +70,6 @@ class DonationService
         'The parameter application_fee_amount cannot be updated on a PaymentIntent after a capture has already been made.',
     ];
 
-
-    // for now we are opting out of async - when we go to async capture we will need to listen to the
-    // webhooks to update donations and also regular giving mandates. See
-    // https://docs.stripe.com/payments/payment-intents/asynchronous-capture#listen-webhooks
-    // and MAT-395
```
NoelLH, line 81: I just moved over the issue ref part of this comment, as I'm not convinced we need the extra stuff Stripe are talking about getting from webhooks there.

src/Domain/DonationService.php
```diff
@@ -322,7 +312,6 @@ public function confirmOnSessionDonation(
             $paymentIntentId,
             [
                 'confirmation_token' => $tokenId->stripeConfirmationTokenId,
-                ...self::ASYNC_CAPTURE_OPT_OUT
```
bdsl, line 325: so its only three months but I can't immediately remember why I added this here in https://github.com/thebiggive/matchbot/pull/1183/files#diff-bd27501321f38ab1ad31b1e5c4640c8f32b128dd8b282103ad0a6801476c87f8R69 . Trying to work that out to check if I misunderstood something then or if there's a reason we should keep it.

bdsl (approved):
LGTM, I think maybe the stripe docs at https://docs.stripe.com/api/payment_intents/update are wrongly showing the capture_method as if it was changeable on update and that mislead me 3 months ago. More comments on Jira.

PR #1336, opened 2025-04-26
Develop -> main

NoelLH:


bdsl (approved):
I don't think this code is actually used anywhere in production other than for regular giving use cases which aren't yet live, but deploying to prod now in case I'm wrong and since it is an important bug fix.

PR #1338, opened 2025-04-26
Add MYSQL schema to files with commands to update files from DB and c…

bdsl:
…heck they match

Intended to make it easier for us to see what our DB schema is by having it as files in the repo, and to see changes as diffs.

To update files after changing DB schema content:
```shell
	./matchbot matchbot:write-schema-files
```

To check if files match:
```shell
./matchbot matchbot:write-schema-files --check
```

The latter has been added to circleCI config.

This could be used in in almost any app with a DB - similar to how Ruby On Rails generates a schema.rb file. I'm not aware of any existing tool to do this in PHP, but there may well be one.


src/Application/Commands/WriteSchemaFile.php
```diff
@@ -0,0 +1,141 @@
...
+            }
+
+            $description = $connection->executeQuery("SHOW CREATE TABLE $table")->fetchAllAssociative();
+            $createTableStatement = $description[0]['Create Table'];
+            \assert(is_string($createTableStatement));
+
+            /* AUTO_INCREMENT varies depending on data (not metadata) in db, not suitable to commit) */
+            $createTableStatement = (string)\preg_replace('/AUTO_INCREMENT=\d+ /', '', $createTableStatement);
+
+            /* Charset and collation vary between local docker setup and circleci. Leave out of files for now to allow
```
NoelLH, line 85: Presumably we could make these match and save a bit of complexity by extending `cimg/mysql` to set params [as for RDS](https://github.com/thebiggive/infrastructure/blob/a0b30387532ddbe24ffc2ccb2a311f3155957a21/terraform/rds-matchbot.tf#L109-L149), but I think this is also fine.

NoelLH (approved):
I see how this could make looking at schema history a little quicker, but it also adds a dev step in cases where we already need to run a migrations diff; and it's also a step that doesn't exist in PHP dev more generally whereas working with ORM and migrations in this way is pretty standard. So probably worth having a readme note about that and a Composer script that does both things together, without wiping the local database as `doctrine:delete-and-recreate` does?

bdsl:
> I see how this could make looking at schema history a little quicker, but it also adds a dev step in cases where we already need to run a migrations diff; and it's also a step that doesn't exist in PHP dev more generally whereas working with ORM and migrations in this way is pretty standard. So probably worth having a readme note about that and a Composer script that does both things together, without wiping the local database as `doctrine:delete-and-recreate` does?

Yes sounds good. Probably the script can do the three things: migrations diff; migrations migrate; and and write schema files.

src/Application/Commands/WriteSchemaFile.php
bdsl: Yes. Having our own docker image just to make this script simpler seems like a bad trade-off, but there probably will be other reasons that we'd want to use the same collation and charset in circleci as in prod.

PR #1339, opened 2025-04-29
Fix marking Salesforce donations to retry after a push failure

NoelLH:


src/Domain/DoctrineDonationRepository.php
```diff
@@ -661,6 +678,7 @@ public function push(DonationUpserted $changeMessage): void
 
             return;
         } catch (BadResponseException $exception) {
+            $this->setSalesforceRePushNeeded($changeMessage->uuid);
```
NoelLH, line 681: I believe this fixes the very specific immediate problem, but it seems like it may be safer to also set the push status before every push attempt so we are protected from similar bugs on different types of crash or exception around here too. Not sure.

If we did want to do that it would maybe make sense to do it before the Messenger message is dispatched. Lots of things envelope these the same so a donation service helper that does both may work.

PR #1342, opened 2025-04-30
Fix ECS entrypoints trying to run against test DB

NoelLH:


composer.json
```diff
@@ -137,8 +140,9 @@
         ],
         "doctrine:ensure-prod": "doctrine orm:ensure-production-settings",
         "doctrine:generate-proxies": "doctrine orm:generate-proxies",
+        "doctrine:migrate:live": "vendor/bin/doctrine-migrations migrate --no-interaction --allow-no-migration",
         "doctrine:migrate": [
-            "vendor/bin/doctrine-migrations migrate --no-interaction --allow-no-migration",
+            "@doctrine:migrate:live",
```
bdsl, line 145: this looks like a necessary fix although afaik the bug comes from f78599c0701bd which was in Jan so not sure if/why we'd just start getting alerts about it now. 

Ideally I think we'd make either have the live and dev versions be equal complexity commands or have live be the default instead of having non-live be default, but I think let's get this out as-is for now to stop generating alarms.

bdsl (approved):

composer.json
NoelLH: @bdsl I'm also not clear on that, I missed that both things changed in Jan and not just the cache one. I guess something to do with the new `Settings` now being in a container alongside all other app dependencies, and the timing vs. Composer's `putenv`?

PR #1344, opened 2025-04-30
fix flaky integration tests

bdsl:
Think the basic problem was just that a query in the test was trying to select the mandate by amount, and as there are only 500 possible amounts this isn't enough to reliably select a given mandate, especially on developer machines where we may run the tests many times before clearing the database.

Replaced with selecting the most recent mandate based on numeric ID, which should work as long as we don't run tests in parallel on a shared database.


bdsl:
All changes are in test code so merging without need for review.

PR #1345, opened 2025-05-02
MAT-390: Adjust matchbot:delete-stale-payment-details to also clear non-usable cards from passworded accounts

bdsl:
…command

Once changes from
https://github.com/thebiggive/donate-frontend/pull/1968 are operating in each environment I'm planning to adjust this command so that it can also detatch all the payment methods we don't have a use for from donors who *do* have passwords set. Should only be needed for as long as takes to get through all the donor accounts / stripe customers from before the date of the front end change.

Dry run mode should help us see what it will do in advance, in dev env even if nowhere else.

### DIFF
--- a/src/Client/Stripe.php
+++ b/src/Client/Stripe.php
@@ -8,12 +8,14 @@ use MatchBot\Domain\StripeCustomerId;
 use MatchBot\Domain\StripePaymentMethodId;
 use Stripe\BalanceTransaction;
 use Stripe\Charge;
+use Stripe\Collection;
 use Stripe\ConfirmationToken;
 use Stripe\CustomerSession;
 use Stripe\Exception\ApiErrorException;
 use Stripe\Exception\InvalidRequestException;
 use Stripe\PaymentIntent;
 use Stripe\PaymentMethod;
+use Stripe\SearchResult;
 use Stripe\SetupIntent;
 
 /**
@@ -35,6 +37,11 @@ interface Stripe
      * @throws ApiErrorException
      */
     public function updatePaymentIntent(string $paymentIntentId, array $updateData): void;
+
+    /**
+     * @throws ApiErrorException
+     */
+    public function updateCustomer(string $customerId, array $updateData): void;
 
     /**
      * @throws ApiErrorException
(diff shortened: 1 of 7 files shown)


bdsl:
This needs to be held back until after https://github.com/thebiggive/donate-frontend/pull/1968 is in production, otherwise it will mark customers as having had their payment methods cleared and then FE will go and add new methods to them that need clearing.

src/Client/Stripe.php
```diff
@@ -73,4 +80,12 @@ public function createSetupIntent(StripeCustomerId $stripeCustomerId): SetupInte
...
 
     public function detatchPaymentMethod(StripePaymentMethodId $paymentMethodId): void;
+
+    public function searchCustomers(array $searchParams): SearchResult|Collection;
+
+    /**
+     * Returns a list of PaymentMethods for Treasury flows.
+     * @return \Stripe\Collection<\Stripe\PaymentMethod>
+     */
+    public function listAllPaymentMethodsForTreasury(array $array): Collection;
```
NoelLH, line 90: As discussed, we'll probably phase out this call and use 'list for customer' per recommendation at https://docs.stripe.com/api/payment_methods/list

tests/Application/Commands/DeleteStalePaymentDetailsTest.php
```diff
@@ -87,8 +84,68 @@ public function testDeletingOneStaleCard(): void
...
+    {
+        $stripeClientProphecy = $this->prophesize(Stripe::class);
+        $customer = new Customer('cus_some_cust_id');
+        $customer->metadata = new StripeObject();
+
+        $paymentMethod1 = new PaymentMethod('pm_1');
+        $paymentMethod1->allow_redisplay = 'always';
+
+        $paymentMethod2 = new PaymentMethod('pm_2');
+        $paymentMethod2->allow_redisplay = 'never';
```
NoelLH, line 97:
```suggestion
        $paymentMethod2->allow_redisplay = 'limited';
```

tests/Application/Commands/DeleteStalePaymentDetailsTest.php
```diff
@@ -87,8 +84,68 @@ public function testDeletingOneStaleCard(): void
...
+        $stripeClientProphecy = $this->prophesize(Stripe::class);
+        $customer = new Customer('cus_some_cust_id');
+        $customer->metadata = new StripeObject();
+        $customer->metadata['hasPasswordSince'] = 'any non null value';
+
+        $paymentMethod1 = new PaymentMethod('pm_1');
+        $paymentMethod1->allow_redisplay = 'always';
+
+        $paymentMethod2 = new PaymentMethod('pm_2');
+        $paymentMethod2->allow_redisplay = 'never';
```
NoelLH, line 128:
```suggestion
        $paymentMethod2->allow_redisplay = 'limited';
```

NoelLH (dismissed):

NoelLH (dismissed):
All LGTM!

I see we have some preexisting places still using Stripe `customers->allPaymentMethods` directly, so it might be nice eventually to replace these with the new helper too to help quickly see all similar calls together. But not essential and doesn't need to be on this PR.

NoelLH (approved):

PR #1347, opened 2025-05-06
Remove donation create retry logic that was never run during GMF2025 …

bdsl:
…and I believe would not have worked anyway

A UniqueConstraintViolationException afaik not be usefully retryable because the EM would be closed the finally block of \Doctrine\ORM\UnitOfWork::commit


NoelLH (approved):
Makes sense, as discussed we might hit this scenario in not-pre-warmed campaigns but the recovery likely doesn't work now in that case anyway. Expected to stop being a possible crash when we pre load all campaigns into MatchBot before launch.

PR #1348, opened 2025-05-06
Fix some PHPStan errors

bdsl:


integrationTests/DonationPersistenceTest.php
```diff
@@ -32,20 +32,6 @@ public function testItSavesADonationToDB(): void
         $this->assertRowsSimilar($this->donationRow(), $donationRow);
     }
 
-
```
bdsl, line 35: the test above checks that it can save and then load from the DB. IMHO it's not especially useful to just check loading in isolation, and now that campaign_id is non-nullable this is more awkward to set up as we have to insert the charity and campaign into db before we can manually insert a donation.

phpstan-baseline.neon
bdsl, line 1: all deletions no added lines in this file

src/Application/Handlers/HttpErrorHandler.php
```diff
@@ -73,7 +73,6 @@ protected function respond(): Response
 
         if (
             !($exception instanceof HttpException)
-            && ($exception instanceof Exception || $exception instanceof Throwable)
```
bdsl, line 76: this could never have been false

src/Domain/Campaign.php
```diff
@@ -26,10 +26,11 @@ class Campaign extends SalesforceReadProxy
      * @var Charity
      */
     #[ORM\ManyToOne(targetEntity: Charity::class, cascade: ['persist'])]
-    #[ORM\JoinColumn(name: 'charity_id', referencedColumnName: 'id')]
+    #[ORM\JoinColumn(name: 'charity_id', referencedColumnName: 'id', nullable: false)]
```
bdsl, line 29: see https://github.com/thebiggive/matchbot/pull/1348/files#r2075960211

schema/FundingWithdrawal.sql
```diff
@@ -7,7 +7,7 @@
 
 CREATE TABLE `FundingWithdrawal` (
   `id` int unsigned NOT NULL AUTO_INCREMENT,
-  `donation_id` int unsigned DEFAULT NULL,
+  `donation_id` int unsigned NOT NULL,
```
bdsl, line 10: `SELECT * from FundingWithdrawal where donation_id is null; ` returns empty in prod

src/Monolog/Handler/SlackHandler.php
```diff
@@ -44,7 +44,7 @@ public function handle(array|\ArrayAccess $record): bool
 
         $lines = explode(PHP_EOL, $message);
 
-        $messageFirstLine = $lines[0] ?? '';
+        $messageFirstLine = $lines[0];
```
bdsl, line 47: explode cannot return an empty array

src/Domain/CampaignFunding.php
```diff
@@ -37,6 +37,7 @@ class CampaignFunding extends Model
     protected Collection $campaigns;
 
     #[ORM\ManyToOne(targetEntity: Fund::class, cascade: ['persist'], inversedBy: 'campaignFundings')]
+    #[ORm\JoinColumn(nullable: false)]
```
NoelLH, line 40:
```suggestion
    #[ORM\JoinColumn(nullable: false)]
```

src/Application/Handlers/HttpErrorHandler.php
NoelLH: Yep, this is the one I tweaked while porting to Identtiy

NoelLH (dismissed):

PR #1350, opened 2025-05-07
Turn off foreign key checks for remaining new migration

NoelLH:


bdsl (approved):
Thanks, managed to forget that I did three migrations not two. Makes sense that since I didn't fix the oldest none of them will have run yet in staging.

PR #1351, opened 2025-05-07
MAT-390 part 3: Display all payment cards on payment methods page, us…

bdsl:
…e stripe default to show only cards with redisplay="always" on in payment element

Should not be merged until part 2 is completed running in production.

See https://docs.stripe.com/api/customer_sessions/create#create_customer_session-components-payment_element-features-payment_method_allow_redisplay_filters

> components.payment_element.features.payment_method_allow_redisplay_filtersarray of enums

> A list of allow_redisplay values that controls which saved payment methods the Payment
> Element displays by filtering to only show payment methods with an allow_redisplay value
> that is present in this list.
>
> If not specified, defaults to [“always”]. In order to display all saved payment methods,
> specify [“always”, “limited”, “unspecified”].

### DIFF
--- a/src/Application/Actions/GetPaymentMethods.php
+++ b/src/Application/Actions/GetPaymentMethods.php
@@ -41,22 +41,22 @@ class GetPaymentMethods extends Action
 
         $regularGivingPaymentMethod = null;
 
-        // exclude payment methods with 'allow_redisplay' set to limited:
-        $displayableMethods = array_values(array_filter(
+        $nonRegularGivingMethods = array_values(array_filter(
             $paymentMethodArray,
             static function (array $paymentMethod) use ($donor, &$regularGivingPaymentMethod) {
                 if ($paymentMethod['id'] === $donor->getRegularGivingPaymentMethod()?->stripePaymentMethodId) {
                     $regularGivingPaymentMethod = $paymentMethod;
-                } else {
-                    return in_array($paymentMethod['allow_redisplay'], ['always', 'unspecified']);
+                    return false;
                 }
+
+                return true;
             }
         ));
 
         return $this->respondWithData(
             $response,
             [
-                'data' => $displayableMethods,
+                'data' => $nonRegularGivingMethods,
                 'regularGivingPaymentMethod' => $regularGivingPaymentMethod
             ]
         );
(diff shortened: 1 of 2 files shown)


src/Application/Actions/GetPaymentMethods.php
```diff
@@ -41,22 +41,22 @@ protected function action(Request $request, Response $response, array $args): Re
 
         $regularGivingPaymentMethod = null;
 
-        // exclude payment methods with 'allow_redisplay' set to limited:
-        $displayableMethods = array_values(array_filter(
+        $nonRegularGivingMethods = array_values(array_filter(
             $paymentMethodArray,
             static function (array $paymentMethod) use ($donor, &$regularGivingPaymentMethod) {
                 if ($paymentMethod['id'] === $donor->getRegularGivingPaymentMethod()?->stripePaymentMethodId) {
                     $regularGivingPaymentMethod = $paymentMethod;
```
bdsl, line 48: this had an implicit return null before - replaced it with explicit return false.

NoelLH: Does this mean we're removing the Regular Giving method from the cards you manage in the account page?

NoelLH (approved):

src/Application/Actions/GetPaymentMethods.php
bdsl: It's on the page, its just shown by itself instead as part of the main list. That's why line 49 assigns $regularGivingPaymentMethod before returning null. This PR isn't intending to change that.

bdsl: I can't 100% remember why we decided to only list it separately, but maybe we thought it would be confusing having it twice, and there are special rules for deleting it - I think you can't delete it, you can only replace it with another method, unless you have zero active mandates.

NoelLH: Right, yep that does now ring a bell. I guess the mutually exclusive on- or off-session choice for the different purposes also makes it potentially clearer to keep them distinct.

bdsl:
Merging this slightly in advance, as I expect MAT-390 part 3 to be completed within the next hour in prod. Merging to be ready to deploy in the morning.

PR #1353, opened 2025-05-07
MAT-390 / ID-52: Delete DeleteStalePaymentDetails command

bdsl:
When MAT-390 part 2 is complete in production, and Identity has taken over detaching payment methods for customers without passwords, and the call to this command in identity is not needed, the command and its supporting code will be redundant.


src/Client/Stripe.php
```diff
@@ -80,8 +75,4 @@ public function createSetupIntent(StripeCustomerId $stripeCustomerId): SetupInte
     public function retrievePaymentMethod(StripeCustomerId $customerId, StripePaymentMethodId $methodId): PaymentMethod;
 
     public function detatchPaymentMethod(StripePaymentMethodId $paymentMethodId): void;
-
-    public function searchCustomers(array $searchParams): SearchResult|Collection;
-
-    public function listAllPaymentMethodsForCustomer(StripeCustomerId $id, array $params): Collection;
```
bdsl, line 86: I might not have deleted these if Stripe hadn't made me, but I think there is value in keeping only the functions we actually use here so we can somewhat at a glance which Stripe features matchbot uses.

NoelLH (dismissed):

PR #1354, opened 2025-05-08
MAT-390: Catch ApiErrorException when deleting stale payment method

bdsl:


bdsl:
Yes this code should just run once in prod which I think will take around 4 hours then have its remaining responsibility taken over by identity and then be deleted.

PR #1356, opened 2025-05-08
Phpstan extra strict

bdsl:
Enable stricter checking with phpstan and https://github.com/phpstan/phpstan-strict-rules

### DIFF
--- a/src/Application/Actions/Hooks/Stripe.php
+++ b/src/Application/Actions/Hooks/Stripe.php
@@ -60,16 +60,16 @@ abstract class Stripe extends Action
         }
 
         if (!$this->event->livemode && getenv('APP_ENV') === 'production') {
-            $message = 'Skipping non-live %s webhook in Production';
+            $logMessage = sprintf('Skipping non-live %s webhook in Production', $this->event->type);
             /**
              * This is normal for Connect events so just `info()` log it in that case.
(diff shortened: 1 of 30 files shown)


src/Application/Actions/Hooks/Stripe.php
```diff
@@ -60,13 +60,18 @@ protected function prepareEvent(
...
+            $logMessage = sprintf('Skipping non-live %s webhook in Production', $this->event->type);
             /**
              * This is normal for Connect events so just `info()` log it in that case.
              * "...both live and test webhooks will be sent to your production webhook URLs."
              * @link https://stripe.com/docs/connect/webhooks
              */
-            $method = $connect ? 'info' : 'warning';
-            $this->logger->$method(sprintf('Skipping non-live %s webhook in Production', $this->event->type));
+            if ($connect) {
+                $this->logger->info($logMessage);
```
NoelLH, line 70: I guess we could also call `log()` with the level as an arg? More concise but maybe not clearer, not a blocker.

NoelLH (approved):

src/Application/Actions/Hooks/Stripe.php
bdsl: yeah I was going to do that but the first argument to `log` on the psr interface is `mixed` so I didn't like it.

PR #1361, opened 2025-05-12
BG2-2784: Upgrade stripe library v15.10.0 => v17.2.0 / for API 2025-0…

bdsl:
…4-30.basil

https://github.com/stripe/stripe-php/releases/tag/v17.2.0

The stripe library has suddenly got a lot more strongly typed - which I
like, but they way they've done it with long complicated docblock
defined types that don't have aliases seems quite difficult to work
with.

A lot of the errors are just indiscriminately suppressed for now -
planning to go through them one at a time before merging to check which
need more substantial fixes, add clearer explanations for why any
suppressions are needed, and reduce repettive typing code.

PHPStan seems to give slightly more useful errors here than Psalm, as it
says which key is missing in a complicated array-shape type


bdsl:
For breaking changes see
https://github.com/stripe/stripe-php/releases/tag/v16.0.0
https://github.com/stripe/stripe-php/releases/tag/v17.0.0

I don't believe any of those affect us.

PR #1363, opened 2025-05-13
Upgrade PHP platform requirement to 8.3.21 and Psalm to latest 6.10.3

bdsl:


integrationTests/CreateDonationTest.php
```diff
@@ -8,6 +8,7 @@
 
 class CreateDonationTest extends IntegrationTest
 {
+    #[\Override]
```
bdsl, line 11: we could turn off the requirement for `Override` in test directories, although probably worth keeping on. Much more important to have them in non-test.

NoelLH (approved):

PR #1365, opened 2025-05-14
MAT-421 – matching recovery

NoelLH:
Edit: now pretty much ready, with test reinstated and temporary stuff removed.

* Recovery from allocation DB errors now reverses fund allocation the right number of times
* Retries of donation create are higher up the stack
* Donation Create only tries up to 2 times, not 3
* Retries don't happen for Regular Giving

Nice for after this PR: test the new retry approach and delete the newly skipped test.

---

WIP, quite a bit still to tidy of course.

* Stripe cancelling with lock test issue is one first faced on https://github.com/thebiggive/matchbot/pull/893 – skipping that test temporarily
* psalm errors are side effects of skipping that test
* I'll check tomorrow if I still think the existing Redis reversal plus clearing the EM is equivalent to the new method I started drafting

### DIFF
--- a/integrationTests/DonationMatchingTest.php
+++ b/integrationTests/DonationMatchingTest.php
@@ -55,9 +55,6 @@ class DonationMatchingTest extends IntegrationTest
         // arrange
         $this->matchingAdapater = $this->makeAdapterThatThrowsAfterSubtractingFunds($this->matchingAdapater);
         $this->setInContainer(Adapter::class, $this->matchingAdapater);
-        $donationRepository = $this->getService(\MatchBot\Domain\DonationRepository::class);
-        Assertion::isInstanceOf($donationRepository, DoctrineDonationRepository::class);
-        $donationRepository->setMatchingAdapter($this->matchingAdapater);
 
         $campaignInfo = $this->addFundedCampaignAndCharityToDB(
             campaignSfId: $this->randomString(),
--- a/integrationTests/DonationMatchingTest.php
+++ b/integrationTests/DonationMatchingTest.php
@@ -76,7 +76,7 @@ class DonationMatchingTest extends IntegrationTest
             $this->createDonation(
                 withPremadeCampaign: false,
                 campaignSfID: $campaign->getSalesforceId(),
-                amountInPounds: 10
+                amountInPounds: 10,
             );
         } catch (\Exception $e) {
             $this->assertEquals(
--- a/integrationTests/StripeCancelsDonationTest.php
+++ b/integrationTests/StripeCancelsDonationTest.php
@@ -2,6 +2,7 @@
 
 namespace MatchBot\IntegrationTests;
 
+use Doctrine\ORM\EntityManagerInterface;
 use GuzzleHttp\Psr7\ServerRequest;
 use MatchBot\Application\Settings;
 use MatchBot\Domain\DonationStatus;
@@ -11,8 +12,6 @@ class StripeCancelsDonationTest extends IntegrationTest
 {
     public function testStripeCanCancelDonation(): void
     {
-        $this->markTestSkipped('See comment in StripePaymentsUpdate::handlePaymentIntentCancelled()');
-
         /**
          * @psalm-suppress MixedArrayAccess
          * @var array<string,string> $donation
@@ -22,6 +21,13 @@ class StripeCancelsDonationTest extends IntegrationTest
             true,
             flags: \JSON_THROW_ON_ERROR
         )['donation'];
+
+        // Clear the previous copy of the donation, otherwise loading the donation with a lock fails, probably
+        // because of https://github.com/doctrine/orm/issues/9505 combined with the test creating and
+        // then patching the donation in one thread. Doctrine doesn't recognise the donation
+        // to cancel as the same and gets an error trying to set the readonly $amount property.
+        // Having a clear EM at the start of each web request is closer to how the real app runs anyway.
+        $this->getService(EntityManagerInterface::class)->clear();
 
         $this->sendCancellationWebhookFromStripe($donation['transactionId']);
 
--- a/integrationTests/StripeCancelsDonationTest.php
+++ b/integrationTests/StripeCancelsDonationTest.php
@@ -11,6 +11,8 @@ class StripeCancelsDonationTest extends IntegrationTest
 {
     public function testStripeCanCancelDonation(): void
     {
+        $this->markTestSkipped('See comment in StripePaymentsUpdate::handlePaymentIntentCancelled()');
+
         /**
          * @psalm-suppress MixedArrayAccess
          * @var array<string,string> $donation
@@ -32,6 +34,9 @@ class StripeCancelsDonationTest extends IntegrationTest
         );
     }
 
+    /**
+     * @psalm-suppress UnusedMethod will be used if we fix test
+     */
     private function sendCancellationWebhookFromStripe(string $transactionId): void
     {
         $paymentIntentId = $transactionId;
--- a/integrationTests/StripeCancelsDonationTest.php
+++ b/integrationTests/StripeCancelsDonationTest.php
@@ -40,9 +40,6 @@ class StripeCancelsDonationTest extends IntegrationTest
         );
     }
 
-    /**
-     * @psalm-suppress UnusedMethod will be used if we fix test
-     */
     private function sendCancellationWebhookFromStripe(string $transactionId): void
     {
         $paymentIntentId = $transactionId;
--- a/src/Application/Actions/Hooks/StripePaymentsUpdate.php
+++ b/src/Application/Actions/Hooks/StripePaymentsUpdate.php
@@ -10,8 +10,6 @@ use Doctrine\ORM\EntityManagerInterface;
 use MatchBot\Application\Actions\ActionPayload;
 use MatchBot\Application\Messenger\DonationUpserted;
 use MatchBot\Application\Notifier\StripeChatterInterface;
-use MatchBot\Domain\CardBrand;
-use MatchBot\Domain\Country;
 use MatchBot\Domain\Currency;
 use MatchBot\Domain\Donation;
 use MatchBot\Domain\DonationFundsNotifier;
@@ -27,19 +25,11 @@ use Psr\Container\ContainerInterface;
 use Psr\Http\Message\ResponseInterface as Response;
 use Psr\Http\Message\ServerRequestInterface as Request;
 use Psr\Log\LoggerInterface;
-use Ramsey\Uuid\Uuid;
-use Stripe\Card;
 use Stripe\Charge;
 use Stripe\Dispute;
 use Stripe\Event;
 use Stripe\PaymentIntent;
-use Stripe\StripeClient;
-use Symfony\Component\Messenger\Envelope;
 use Symfony\Component\Messenger\RoutableMessageBus;
-use Symfony\Component\Messenger\Stamp\DelayStamp;
-use Symfony\Component\Notifier\Bridge\Slack\Block\SlackHeaderBlock;
-use Symfony\Component\Notifier\Bridge\Slack\Block\SlackSectionBlock;
-use Symfony\Component\Notifier\Bridge\Slack\SlackOptions;
 use Symfony\Component\Notifier\ChatterInterface;
 use Symfony\Component\Notifier\Message\ChatMessage;
 
@@ -284,7 +274,7 @@ class StripePaymentsUpdate extends Stripe
 
         $this->entityManager->beginTransaction();
 
-        $donation = $this->donationRepository->findOneBy(['chargeId' => $charge->id]);
+        $donation = $this->donationRepository->findAndLockOneBy(['chargeId' => $charge->id]);
 
         if (!$donation) {
             $this->logger->notice(sprintf('Donation not found with Charge ID %s', $charge->id));
@@ -364,7 +354,12 @@ class StripePaymentsUpdate extends Stripe
         $paymentIntent = $event->data->object;
         \assert($paymentIntent instanceof PaymentIntent);
 
-        $donation = $this->donationRepository->findOneBy(['transactionId' => $paymentIntent->id]);
+        $this->entityManager->beginTransaction();
+        // Locking this fails StripeCancelsDonationTest currently, probably because of
+        // https://github.com/doctrine/orm/issues/9505 combined with the test creating and
+        // then patching the donation in one thread? Doctrine doesn't recognise the donation
+        // to cancel as the same and gets an error trying to set the readonly $amount property.
+        $donation = $this->donationRepository->findAndLockOneBy(['transactionId' => $paymentIntent->id]);
 
         if ($donation === null) {
             if (getenv('APP_ENV') !== 'production') {
@@ -393,6 +388,7 @@ class StripePaymentsUpdate extends Stripe
         }
 
         $this->entityManager->flush();
+        $this->entityManager->commit();
         $this->queueSalesforceUpdate($donation);
 
         return $this->respond($response, new ActionPayload(200));
--- a/src/Application/Actions/Hooks/StripePaymentsUpdate.php
+++ b/src/Application/Actions/Hooks/StripePaymentsUpdate.php
@@ -355,10 +355,6 @@ class StripePaymentsUpdate extends Stripe
         \assert($paymentIntent instanceof PaymentIntent);
 
         $this->entityManager->beginTransaction();
-        // Locking this fails StripeCancelsDonationTest currently, probably because of
-        // https://github.com/doctrine/orm/issues/9505 combined with the test creating and
-        // then patching the donation in one thread? Doctrine doesn't recognise the donation
-        // to cancel as the same and gets an error trying to set the readonly $amount property.
         $donation = $this->donationRepository->findAndLockOneBy(['transactionId' => $paymentIntent->id]);
 
         if ($donation === null) {
--- a/src/Application/Matching/Allocator.php
+++ b/src/Application/Matching/Allocator.php
@@ -6,6 +6,7 @@ namespace MatchBot\Application\Matching;
 
 use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\ORM\EntityManagerInterface;
+use MatchBot\Application\Assertion;
 use MatchBot\Domain\CampaignFunding;
 use MatchBot\Domain\CampaignFundingRepository;
 use MatchBot\Domain\Donation;
@@ -24,7 +25,9 @@ class Allocator
         private EntityManagerInterface $entityManager,
         private LoggerInterface $logger,
     ) {
-        $this->campaignFundingRepository = $entityManager->getRepository(CampaignFunding::class);
+        $campaignFundingRepo = $entityManager->getRepository(CampaignFunding::class);
+        \assert($campaignFundingRepo instanceof CampaignFundingRepository);
+        $this->campaignFundingRepository = $campaignFundingRepo;
     }
 
     /**
@@ -39,6 +42,8 @@ class Allocator
      */
     public function allocateMatchFunds(Donation $donation): string
     {
+        Assertion::notNull($this->campaignFundingRepository);
+
         // We look up matching withdrawals to allow for the case where retrospective matching was required
         // and the donation is not new, and *some* (or full) matching already occurred. The collection of withdrawals
         // is most often empty (for new donations) so this will frequently be 0.00.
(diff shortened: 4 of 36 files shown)


src/Domain/DoctrineDonationRepository.php
```diff
@@ -90,10 +87,28 @@ public function allocateMatchFunds(Donation $donation): string
...
+            ));
+
+            // todo pick one
+            $this->matchingAdapter->releaseNewlyAllocatedFunds();
+            //$this->reverseMatchingAllocations($donation, $newWithdrawals, $likelyAvailableFunds);
+
+            // Ensure nothing later tries to persist the pending Donation or CampaignFunding changes.
+            $this->getEntityManager()->clear();
+
+            return '0.0';
```
bdsl, line 107: I think if we're going to clear the EM then we should throw effectively straight back to the UI. There's no reason for a caller to assume that this might clear the EM so I think it might be hard to discover that or understand why changes aren't getting persisted if we don't remember that this clear line is here.

Also probably better to close the em instead of just clearing it, so any attempt to flush after this would throw instead of letting us think we'd successfully persisted changes.

src/Domain/DoctrineDonationRepository.php
```diff
@@ -90,10 +87,28 @@ public function allocateMatchFunds(Donation $donation): string
...
+        } catch (\Throwable $exception) {
+            $this->getLogger()->error(sprintf(
+                'Failed to flush DB for new withdrawals so donation UUID %s did not get matching: %s',
+                $donation->getUuid()->toString(),
+                $exception->getMessage(),
+            ));
+
+            // todo pick one
+            $this->matchingAdapter->releaseNewlyAllocatedFunds();
+            //$this->reverseMatchingAllocations($donation, $newWithdrawals, $likelyAvailableFunds);
```
bdsl, line 102: noel thinks probably pick `releaseNewlyAllocatedFunds` (I haven't looked enough yet to have my own view)

src/Domain/DoctrineDonationRepository.php
```diff
@@ -887,4 +905,20 @@ public function findOneByUUID(UuidInterface $donationUUID): ?Donation
...
+//            $this->matchingAdapter->addAmount($newWithdrawal->getCampaignFunding(), $newWithdrawal->getAmount());
+//        }
+//
+////        $em = $this->getEntityManager();
+////        $em->refresh($donation);
+////
+////        foreach ($likelyAvailableFundings as $campaignFunding) {
+////            $em->refresh($campaignFunding);
+////        }
+//    }
```
bdsl, line 923: agreed with noel this will probably be deleted

integrationTests/DonationMatchingTest.php
```diff
@@ -76,7 +76,7 @@ public function testACrashedDonationDoesNotReduceAvailableMatchFunds(): void
             $this->createDonation(
                 withPremadeCampaign: false,
                 campaignSfID: $campaign->getSalesforceId(),
-                amountInPounds: 10
```
bdsl, line 79: :+1:

src/Application/Actions/Hooks/StripePaymentsUpdate.php
```diff
@@ -364,7 +354,12 @@ private function handlePaymentIntentCancelled(Event $event, Response $response):
         $paymentIntent = $event->data->object;
         \assert($paymentIntent instanceof PaymentIntent);
 
-        $donation = $this->donationRepository->findOneBy(['transactionId' => $paymentIntent->id]);
+        $this->entityManager->beginTransaction();
+        // Locking this fails StripeCancelsDonationTest currently, probably because of
+        // https://github.com/doctrine/orm/issues/9505 combined with the test creating and
+        // then patching the donation in one thread? Doctrine doesn't recognise the donation
+        // to cancel as the same and gets an error trying to set the readonly $amount property.
```
bdsl, line 361: I see there's a PR 3 days ago to fix that issue in the ORM https://github.com/doctrine/orm/pull/11935/files . 

We could workaround by making amount only readonly for static analysis and not actually runtime readonly, although having runtime readonly is reassuring.

Probably a better option is to change the implementation of `findAndLockOneBy` to avoid using `refresh`. I think it only needs refresh because we don't know the ID of the donation that we want to lock before we fetch it. We could do a query without using the ORM to get the ID and then lock by ID. Or maybe we don't need to know the ID first if we use a DQL query, something like

```php
        $query = $this->getEntityManager()->createQuery(
            'SELECT d FROM Matchbot\Domain\Donation d where uuid = ?'
        );
        $query->setParameter(0, '00000000-0000-0000-0000-000000000000');
        $query->setLockMode(LockMode::PESSIMISTIC_WRITE);
        $result = $query->execute();
```

src/Application/Messenger/Handler/GiftAidResultHandler.php
```diff
@@ -28,8 +28,10 @@ public function __invoke(Messages\Donation $donationMessage): void
             $donationMessage->id,
         ));
 
+        $this->entityManager->beginTransaction();
```
bdsl, line 31: What do you think about replacing this (and similar) with
```php
$this->entityManager->wrapInTransaction(function () use ($donationMessage) {
```

I think makes it a bit easier to see at a glance where the transaction ends and impossible to forget to end the transaction. Sometimes the deep nesting can be awkward but I think OK in this case and usually.

src/Domain/DoctrineDonationRepository.php
bdsl, line 107: definitely think we shouldn't return after catching any Throwable - we don't know what went wrong so safest thing we can do is re-throw.

src/Domain/DoctrineDonationRepository.php
```diff
@@ -52,7 +51,7 @@ public function allocateMatchFunds(Donation $donation): string
         // is most often empty (for new donations) so this will frequently be 0.00.
```
bdsl, line 51: as the allocateMatchFunds function is getting longer maybe this would be a good time to move it out of the repository into a new class just for this function and the ones it calls (`allocateFundsAndPrepareDBChanges` and  `releaseNewlyAllocatedFunds`)

src/Domain/DonationService.php
```diff
@@ -846,26 +861,20 @@ public function attemptFundingAllocation(Donation $donation): void
...
-                    // warning indicates that we *may* retry, as it depends on whether this is in the last retry or
-                    // not.
-                    $this->logger->warning(sprintf('Allocation got error, may retry: %s', $t->getMessage()));
-
-                    $this->matchingAdapter->releaseNewlyAllocatedFunds();
+            retryable: fn (Donation $donation) => $this->donationRepository->allocateMatchFunds($donation),
+            actionName: 'allocate match funds',
+            donation: $donation,
+            onErrorTidy: function (Donation $donation) {
+                $this->logger->info(sprintf('Resetting after match error for UUID %s', $donation->getUuid()));
```
bdsl, line 868: I think it's probably clearer if we don't pass donation to runWithPossibleRetry when all that will do is pass it to the closures - and then it has to be documented as only taking closures that accept donations. Instead we can capture the donation to the closures before passing them. The short closure will auto-capture, the long one can use `use`.

```suggestion
            retryable: fn () => $this->donationRepository->allocateMatchFunds($donation),
            actionName: 'allocate match funds',
            onErrorTidy: function () use ($donation) {
                $this->logger->info(sprintf('Resetting after match error for UUID %s', $donation->getUuid()));
```

src/Domain/DoctrineDonationRepository.php
NoelLH: Yep, agreed – this should be done in e9d51a4b734c01357d31a7fa537cda242124f5dc

src/Application/Actions/Hooks/StripePaymentsUpdate.php
NoelLH: Interesting it might be getting fixed upstream!

I realised that an EM clear in the test itself also fixes it and is actually quite realistic, so I think it's a decent workaround for this test. Did that in d734eb3f440ba89176802875ed2a25156dd4f66b

I'm not necessarily opposed to changing how we find and lock, but given this I probably won't try on this PR. I'm not sure about putting the `WHERE` in DQL because we find by multiple different fields. I think we're deferring a reasonably amount of logic to Doctrine the current way, at the expense of a slightly confusing pattern where we `refresh()`.

src/Domain/DoctrineDonationRepository.php
```diff
@@ -90,10 +87,22 @@ public function allocateMatchFunds(Donation $donation): string
...
-        $this->getEntityManager()->flush(); // Flush `$newWithdrawals` if any.
+        try {
+            // Flush `$newWithdrawals` if any and updates to DB copies of CampaignFundings.
+            $this->getEntityManager()->flush();
+        } catch (\Throwable $exception) {
+            $this->matchingAdapter->releaseNewlyAllocatedFunds();
+
+            // Ensure nothing later tries to persist the pending Donation or CampaignFunding changes.
+            $this->getEntityManager()->close();
+            throw new \RuntimeException(
```
bdsl, line 98: I think use of `RuntimeException` implies something not worth catching except at the very top level. But might be worth throwing something more specific (and not derived from this class) here so we can catch it in e.g. the action and tell the donor that matching failed and they can try again.

src/Domain/DonationService.php
NoelLH: Yes, I tried this and got confused when unit tests had issues. It looks like the new issue there was actually from me moving `$retryable` to the short syntax, and relying on the automatic capture you mentioned. Not sure why the tests weren't happy that way. I was getting e.g.
> ArgumentCountError: Too few arguments to function MatchBot\Domain\DonationService::MatchBot\Domain\{closure}(), 0 passed in /var/www/html/src/Domain/DonationService.php on line 239 and exactly 1 expecte

Reversed that change for now and removed the extraneous new arg, in 17b3e482df1a99f5d3ccee324ea0134b7457f5bc

src/Application/Messenger/Handler/GiftAidResultHandler.php
bdsl: Thanks. Also I just noticed that Doctrine's implementation of wrapInTransaction does a bit more than begin and commit - if the closure throws it rolls back the transaction and closes the entity manager, both of which are good.

integrationTests/StripeCancelsDonationTest.php
```diff
@@ -32,6 +40,9 @@ public function testStripeCanCancelDonation(): void
         );
     }
 
+    /**
+     * @psalm-suppress UnusedMethod will be used if we fix test
+     */
     private function sendCancellationWebhookFromStripe(string $transactionId): void
```
bdsl, line 46:
```suggestion
    private function sendCancellationWebhookFromStripe(string $transactionId): void
```

src/Domain/DonationService.php
bdsl: Thanks. Not sure what the issue with the short syntax was, tests pass for me when I make this edit:

```diff
         $this->runWithPossibleRetry(
-            retryable: function () use ($donation) {
-                $this->donationRepository->allocateMatchFunds($donation);
-            },
+            retryable: fn() => $this->donationRepository->allocateMatchFunds($donation),
             actionName: 'allocate match funds',
```

src/Domain/DonationService.php
```diff
@@ -846,26 +863,22 @@ public function attemptFundingAllocation(Donation $donation): void
...
+                // DB copies get out of sync with the Redis source of truth.
+                $this->entityManager->clear();
 
-                    throw $t;
-                }
+                $this->logger->info(sprintf('Reset tasks done after match error for UUID %s', $donation->getUuid()));
             },
-            'allocate match funds'
         );
     }
```
bdsl, line 883: I'm not sure passing `onErrorTidy` as a param is doing anything useful? Looks like we can achieve the same thing more simply like this:

```php
    public function attemptFundingAllocation(Donation $donation): void
    {
        // @todo-MAT-388: remove runWithPossibleRetry if we determine its not useful and unwrap body of function below
        try {
            $this->runWithPossibleRetry(
                retryable: fn() => $this->donationRepository->allocateMatchFunds($donation),
                actionName: 'allocate match funds',
            );
        } catch (\Throwable $t) {
            $this->logger->info(sprintf('Resetting after match error for UUID %s', $donation->getUuid()));

            $this->matchingAdapter->releaseNewlyAllocatedFunds();

            // `DoctrineDonationRepository` *should* have already `clear()`ed the EM in this case, but
            // do it again just in case. We want ot ensure that no FundingWithdrawals of CampaignFunding
            // DB copies get out of sync with the Redis source of truth.
            $this->entityManager->clear();

            $this->logger->info(sprintf('Reset tasks done after match error for UUID %s', $donation->getUuid()));

            throw $t;
        }
    }
```

bdsl: but of course we can never gurantee that we'll catch and tidy up after any error, since it could be a fatal error instead of a throw, or something like a lost connection to redis.

bdsl:
@NoelLH I think I've reviewed this as much as I can for now, I'm still struggling a bit to break the changes down in in dependant units small enough to fit in my head to understand one at a time.

I'm wondering if there's a 'kernel' of the diff you can point to, the minimal possible change that would have prevent the particular bug we saw? I think that it's something that reduces the number of times we call `\MatchBot\Application\Matching\Adapter::addAmount` as we called that two or three times when it should have been once. So the PR consists of that kernal plus several related improvements that prevent similar bugs or make the code easier to work with.

And maybe better to come back to this after we've continued the discussion on Jira and/or pair instead of doing async code review.

src/Application/Matching/Allocator.php
```diff
@@ -0,0 +1,229 @@
...
+    private ?CampaignFundingRepository $campaignFundingRepository = null;
+
+    public function __construct(
+        private Adapter $adapter,
+        private EntityManagerInterface $entityManager,
+        private LoggerInterface $logger,
+    ) {
+        // For now, just type this as a few tests don't set it up. We assert before use in allocateMatchFunds() where
+        // it's really needed.
+        /** @var CampaignFundingRepository $campaignFundingRepo */ // @phpstan-ignore varTag.type
```
NoelLH, line 30: Not sure why phpstan doesn't like this type. Adding the `<>` makes it complain about too many template parts..

tests/Application/Matching/AllocatorTest.php
```diff
@@ -310,4 +304,30 @@ public function testItRejectsFundingInWrongCurrency(): void
         // act
         $this->sut->allocateMatchFunds($donation);
     }
+
+    public function testReleaseMatchFundsSuccess(): void
```
NoelLH, line 308: This is moved from the general donation repo test. Everything else came from the allocation-specific repo test, as git's picked up.

src/Application/Matching/Allocator.php
bdsl: phpstan wants us to do `\assert($campaignFundingRepo instanceof CampaignFundingRepository);` on the next line to make sure the type is what we say it is instead of just using the var tag to suggest it. I'm not sure where you were adding `<>`, CampaignFundingRepository is not generic.

src/Application/Matching/Allocator.php
```diff
@@ -0,0 +1,229 @@
...
+use MatchBot\Domain\CampaignFunding;
+use MatchBot\Domain\CampaignFundingRepository;
+use MatchBot\Domain\Donation;
+use MatchBot\Domain\FundingWithdrawal;
+use Psr\Log\LoggerInterface;
+
+class Allocator
+{
+    /**
+     * Seems we must do this dynamically to avoid circular dependency. Remains null for some tests for now.
```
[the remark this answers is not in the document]
NoelLH: Oh, maybe! I saw an opaquely null service at runtime which is exactly what I remembered from the last circular dependency chain involving EM repo's, so assumed and fixed it that way.

src/Application/Matching/Allocator.php
NoelLH: Yes, that's what I wanted to do but it would have involved more test work as the comment says. Hopefully not needed in any case if we can successfully change the constructor DI.

src/Domain/DonationService.php
```diff
@@ -846,26 +863,22 @@ public function attemptFundingAllocation(Donation $donation): void
...
-                    // we have to also remove the FundingWithdrawls from MySQL - otherwise the redis amount
-                    // would be reduced again when the donation expires.
-                    $this->donationRepository->removeAllFundingWithdrawalsForDonation($donation);
+                $this->matchingAdapter->releaseNewlyAllocatedFunds();
 
-                    $this->entityManager->flush();
+                // `DoctrineDonationRepository` *should* have already `clear()`ed the EM in this case, but
+                // do it again just in case. We want ot ensure that no FundingWithdrawals of CampaignFunding
+                // DB copies get out of sync with the Redis source of truth.
+                $this->entityManager->clear();
```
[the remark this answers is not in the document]
NoelLH: In this context I don't agree, as the fn is called in a much more narrow context and closing the EM would stop us handling retryable cases, leading to avoidable breaks for donors.

Picking up the point about safety for future code, we could explicit checking the EM unit of work is empty at the start of `attemptFundingAllocation()` and error otherwise.

src/Application/Matching/Allocator.php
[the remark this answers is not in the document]
bdsl: Right, will you merge the PR 1367 linked above into this one if you think it's good? Not sure if by runtime you mean manual tests or auto tests, but auto tests are passing on that that PR

src/Domain/DonationService.php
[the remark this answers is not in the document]
NoelLH: The map need not be empty but there should be no changes waiting.

I've tried to cover it in c6603004fa34aadfbd8d2c4c9783340442b76070. It seems like quite a lot of runtime cost compared to relying on the fn doc block and future code review, but I'm not totally against keeping it.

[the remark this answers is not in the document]
bdsl: I'm not sure why you say the map need not be empty. Calling `\Doctrine\Persistence\ObjectManager::clear` detaches all managed entities. Even if we don't have changes we a probably relying on the entities being attached to the manager so that it will detect and persist any changes we make to them later.

src/Application/Matching/Allocator.php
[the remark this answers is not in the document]
NoelLH: Thanks, yep the PR looks good! It was automated tests and I must have got confused making other changes simultaneously.

src/Domain/DonationService.php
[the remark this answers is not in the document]
NoelLH: The runtime code before this flushes but does not clear. I think that's OK, so my new commit aims to check that there aren't pending changes rather than that the UOW is entirely empty.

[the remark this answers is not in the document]
bdsl: I'm still confused, probably best to discuss on a call later

src/Domain/DonationService.php
NoelLH: I think this might reverse the main fix?

Let's imagine `runWithPossibleRetry()` runs 3 times: twice with retryable exceptions and then once with a deadlock.

As it stands it makes 3 attempts at its `$retryable`. If it did matching adapter allocation 3 times, we want a recovery fn to reverse all 3. A single catch around the whole thing would only reverse 1.

src/Application/Messenger/Handler/GiftAidResultHandler.php
```diff
@@ -28,28 +28,30 @@ public function __invoke(Messages\Donation $donationMessage): void
             $donationMessage->id,
         ));
 
-        /** @var Donation $donation */
-        $donation = $this->donationRepository->findOneByUUID(Uuid::fromString($donationMessage->id));
+        $this->entityManager->wrapInTransaction(function () use ($donationMessage) {
+            /** @var Donation $donation */
+            $donation = $this->donationRepository->findAndLockOneByUUID(Uuid::fromString($donationMessage->id));
```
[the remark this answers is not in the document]
NoelLH: I'm not sure we could straightforwardly make the locking private, as there are things like payouts that are already quite expensive which share a transaction. Locks live inside a loop but we wouldn't want a transaction for every iteration.

src/Domain/DonationService.php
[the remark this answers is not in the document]
bdsl: agreed clear is a problem unless we add some extra stuff as we have code like this:

```
$em->persist($something);
$em->flush;
callOurFunctionWithClearInside();
$something->mutate();
$em->flush();
```

If we don't know that `callOurFunctionWithClearInside` will clear the EM then the last line will unexpectedly fail to persist changes made on previous line.

bdsl (approved):
Paring with @NoelLH  - LGTM although some stuff to look at after merge including why we now seem to need process isolation in integration tests to enable deleting donation records left over from previous tests.

PR #1366, opened 2025-05-14
MAT-395: Change stripe capture method to automatic_async

bdsl:
As there are quite a lot of changes already in https://github.com/thebiggive/matchbot/pull/1362 that I'd like to get into prod without more delay than necessary I think I'd wait until that's deployed before merging and testing this on staging.

Changes here may also end up conflicting with https://github.com/thebiggive/matchbot/pull/1365

### DIFF
--- a/src/Application/Actions/Hooks/StripePaymentsUpdate.php
+++ b/src/Application/Actions/Hooks/StripePaymentsUpdate.php
@@ -102,14 +102,19 @@ class StripePaymentsUpdate extends Stripe
                 return $this->handleChargeDisputeClosed($event, $response);
             case Event::CHARGE_REFUNDED:
                 return $this->handleChargeRefunded($event, $response);
+                // we have to listen for both CHARGE_SUCCEEDED and CHARGE_UPDATED - succeded comes much quicker and
+               // we need to thank the donor for their donation quickly so they know it worked. Updated comes later and
+            // gives us further information that we'll need for charity payouts, e.g. transfer ID.
             case Event::CHARGE_SUCCEEDED:
                 return $this->handleChargeSucceeded($event, $response);
+            case Event::CHARGE_UPDATED:
+                return $this->handleChargeUpdated($event, $response);
             case Event::PAYMENT_INTENT_CANCELED:
                 return $this->handlePaymentIntentCancelled($event, $response);
             case Event::CUSTOMER_CASH_BALANCE_TRANSACTION_CREATED:
                 return $this->handleCashBalanceUpdate($event, $response);
             default:
-                $this->logger->warning(sprintf('Unsupported event type "%s"', $type));
+                $this->logger->info(sprintf('Unsupported event type "%s"', $type));
                 return $this->respond($response, new ActionPayload(204));
         }
     }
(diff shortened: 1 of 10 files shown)


src/Application/Actions/Hooks/StripePaymentsUpdate.php
```diff
@@ -181,6 +187,63 @@ private function handleChargeSucceeded(Event $event, Response $response): Respon
...
+        ));
+
+        $this->entityManager->persist($donation);
+        $this->entityManager->commit();
+        $this->entityManager->flush();
+        $this->queueSalesforceUpdate($donation);
+
+        return $this->respondWithData($response, $charge);
+    }
+
```
bdsl, line 246: There's quite a bit of duplicated code in this file (and was already). I'd like to tidy that up but if I do it in this PR now I think it will create an awkward merge conflict with the other open matchbot PR.

src/Application/Actions/Hooks/StripePaymentsUpdate.php
```diff
@@ -97,21 +97,27 @@ protected function action(Request $request, Response $response, array $args): Re
...
-                return $this->handleChargeRefunded($event, $response);
-            case Event::CHARGE_SUCCEEDED:
-                return $this->handleChargeSucceeded($event, $response);
-            case Event::PAYMENT_INTENT_CANCELED:
-                return $this->handlePaymentIntentCancelled($event, $response);
-            case Event::CUSTOMER_CASH_BALANCE_TRANSACTION_CREATED:
-                return $this->handleCashBalanceUpdate($event, $response);
-            default:
-                $this->logger->warning(sprintf('Unsupported event type "%s"', $type));
+        return match ($type) {
```
bdsl, line 100: I thought match was slightly simpler and more robust than switch, although its marginal in this case.

src/Application/Actions/Hooks/StripePaymentsUpdate.php
```diff
@@ -97,21 +97,27 @@ protected function action(Request $request, Response $response, array $args): Re
...
+            Event::CHARGE_SUCCEEDED =>
+                 $this->handleChargeSucceeded($event, $response),
+            Event::CHARGE_UPDATED =>
+                 $this->handleChargeUpdated($event, $response),
+            Event::PAYMENT_INTENT_CANCELED =>
+                 $this->handlePaymentIntentCancelled($event, $response),
+            Event::CUSTOMER_CASH_BALANCE_TRANSACTION_CREATED =>
+                 $this->handleCashBalanceUpdate($event, $response),
+            default => (function () use ($type, $response) {
+                $this->logger->info(sprintf('Unsupported event type "%s"', $type));
```
bdsl, line 117: We're not attempting to handle every event type, so I think it makes sense to downgrade the warning here to an info if not a debug or to nothing at all.

NoelLH: I think that as we have none now, it might be helpful to see warnings if we ever add more webhooks but then don't use them. Not a blocker though.

NoelLH (commented):
I guess our test mocks like `ch_succeeded` should probably be updated to reflect the new expectation ideally, and a counterpart added for update when we have the extra info.

Approach LGTM so far.

src/Application/Actions/Hooks/StripePaymentsUpdate.php
bdsl: Ah ok. I was seeing warnings in my local, maybe we have warnings in local dev envs but not in other environments. I'll put it back to warning in that case, and might look at stopping the local warnings separately.

src/Application/Actions/Hooks/StripePaymentsUpdate.php
```diff
@@ -97,27 +97,32 @@ protected function action(Request $request, Response $response, array $args): Re
...
```
bdsl, line 117:
```suggestion
                $this->logger->warning(sprintf('Unsupported event type "%s"', $type));
```

bdsl:
> I guess our test mocks like `ch_succeeded` should probably be updated to reflect the new expectation ideally, and a counterpart added for update when we have the extra info.

Right, I'll have a look at those.

src/Application/Actions/Hooks/StripePaymentsUpdate.php
bdsl: back to warning in 1f63a1a280435071a90a056c24616c7f709590a5

NoelLH (commented):
Mostly LGTM.

I think if we want to still share `updateDonationStatusFromSuccessfulCharge()` between the 2 events and keep things clear, we might ideally want to also do similarly for `$charge->balance_transaction` and make explicit that it may be null. We're probably getting away with casting it to blank string because of the existing handling for simulated hooks, but ideally the param type and the comment about that would be updated so it's obvious at a glance that this is now a real runtime scenario too.

bdsl:
> Mostly LGTM.
> 
> I think if we want to still share `updateDonationStatusFromSuccessfulCharge()` between the 2 events and keep things clear, we might ideally want to also do similarly for `$charge->balance_transaction` and make explicit that it may be null. We're probably getting away with casting it to blank string because of the existing handling for simulated hooks, but ideally the param type and the comment about that would be updated so it's obvious at a glance that this is now a real runtime scenario too.

Thanks I pushed https://github.com/thebiggive/matchbot/pull/1366/commits/f583b744e95b77cf41575231d6a409a2168dcfa8 hopefully making this clear enough.

NoelLH (approved):

PR #1367, opened 2025-05-15
MAT-421: Refactor use standard constructor injection for CampaignFund…

bdsl:
…ingRepositor in Allocator

### DIFF
--- a/tests/Application/Actions/Donations/CancelAllTest.php
+++ b/tests/Application/Actions/Donations/CancelAllTest.php
@@ -8,7 +8,10 @@ use DI\Container;
 use Doctrine\ORM\EntityManagerInterface;
 use MatchBot\Application\Matching\Allocator;
 use MatchBot\Client\BadRequestException;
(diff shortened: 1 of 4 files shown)


tests/Application/Actions/Donations/CancelAllTest.php
```diff
@@ -183,6 +186,7 @@ private function setDoublesInContainer(
         ObjectProphecy $stripeProphecy = null,
         ObjectProphecy $allocatorProphecy = null,
     ): void {
+        $container->set(CampaignFundingRepository::class, $this->prophesize(CampaignFundingRepository::class)->reveal());
```
bdsl, line 189: or
```suggestion
$container->set(Allocator::class, $this->prophesize(Allocator::class)->reveal());
```

which would be more similar to what we have now. DonationRepository is splitting into DonationRepository and Allocator so we would now fake both of them instead of just the former.

NoelLH (approved):

PR #1368, opened 2025-05-17
More stanning for php

bdsl:
The biggest change at this level is that whenever using a generic or iteratable type we have to say what it's of - we can't just have an `array` or  a `Salesforce18Id`, we have to specify what type it an array or Salesforce18Id *of*.


NoelLH (approved):
Skimmed each commit somewhat, given the description overview & the number of small changes. I agree it seems safe enough.

PR #1370, opened 2025-05-20
MAT-421: Run integration tests without process isolation

bdsl:
We need to make sure that any db transactions opened during an integration test are committed or rolled back by the end so that it doesn't affect subsequent tests.

This change cuts about 40 seconds off the integration test runtime, which feels like a big saving to me given how often I'm pushing matchbot changes and waiting for a result.


integrationTests/CreateRegularGivingMandateTest.php
```diff
@@ -95,6 +95,8 @@ public function setUp(): void
     public function tearDown(): void
     {
         $this->getContainer()->set(MessageBusInterface::class, $this->originalMessageBus);
+
+        parent::tearDown();
```
bdsl, line 99: not sure exactly why PHPStan suddenly started complaining that this is missing, but it's right that it should be there.

PR #1371, opened 2025-05-20
MAT-421: Restore skipped test

bdsl:
This is now passing but may need a bit more work to understand why we're logging the same things twice and adding comments to test


NoelLH (commented):
Looks good so far. I'd agree it seems worth understanding more and probably not logging the '1 of 3' part if the inner retry can tell that it won't really retry the particular exception type it hit

PR #1373, opened 2025-05-20
BG2-2890: Delete unwanted donor account and redact email address on d…

bdsl:
…onations

### DIFF
--- a/src/Migrations/Version20250520161231.php
+++ b/src/Migrations/Version20250520161231.php
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MatchBot\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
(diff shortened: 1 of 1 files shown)


src/Migrations/Version20250520161231.php
```diff
@@ -0,0 +1,47 @@
...
+        );
+
+        if (Environment::current() === Environment::Staging) {
+            return; // in case of collisions between donation ids we're wanting to adjust in prod and any in staging.
+        }
+
+        $this->addSql(
+            <<<SQL
+            UPDATE Donation SET 
+                                Donation.donorEmailAddress = 'email-redacted',
```
NoelLH, line 34:
```suggestion
                               donorEmailAddress = 'email-redacted',
```
When I simulate the update in pMA it seems to be unable to work out which field this is without this change. Not totally sure why but seems safer to do this just in case that's accurate.

NoelLH (approved):

PR #1374, opened 2025-05-20
MAT-422: Stop GA claims

bdsl:


src/Migrations/Version20250520170950.php
```diff
@@ -0,0 +1,36 @@
...
+    public function getDescription(): string
+    {
+        return 'Remove plan to claim GA for one charity';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql(<<<'SQL'
+            UPDATE Donation 
+            SET matchbot.Donation.tbgShouldProcessGiftAid = 0
```
NoelLH, line 21:
```suggestion
            SET tbgShouldProcessGiftAid = 0
```
Possibly needed as with other recent migration – pMA simulate certainly gets confused with the prefix.

src/Migrations/Version20250520170950.php
```diff
@@ -0,0 +1,36 @@
...
+            WHERE Donation.campaign_id = 8963 and Donation.donationStatus in ('paid', 'collected')
```
NoelLH, line 22:
```suggestion
            WHERE Donation.campaign_id = 8963 and Donation.donationStatus in ('Paid', 'Collected')
```
Marginally clearer for future code searchers? But collation is `_ci`, not a blocker.

src/Migrations/Version20250520170950.php
```diff
@@ -0,0 +1,36 @@
...
+            WHERE Donation.campaign_id = 8963 and Donation.donationStatus in ('Paid', 'Collected')
+            LIMIT 500; -- this is more than the actual number of donations, not wishing to publish the exact figure.
+            SQL
+        );
+    }
+
+    public function down(Schema $schema): void
+    {        $this->addSql(<<<'SQL'
+            UPDATE Donation 
+            SET matchbot.Donation.tbgShouldProcessGiftAid = 1 -- was 1 in all cases before running up migration
```
NoelLH, line 31:
```suggestion
            SET tbgShouldProcessGiftAid = 1 -- was 1 in all cases before running up migration
```
As above

NoelLH (dismissed):
I think `down()` needs the parallel fixes ideally, not a blocker

PR #1375, opened 2025-05-20
MAT-419: Store donation payout details and send to Salesforce

bdsl:
Adds new fields on donation table and object:

```
`stripePayoutId` varchar(255) DEFAULT NULL,
`paidOutAt` datetime DEFAULT NULL COMMENT '(DC2Type:datetime_immutable)',
`payoutSuccessful` tinyint(1) DEFAULT NULL,
```
  
And similarly adds to message we send to salesforce:

```
'stripePayoutId' => $this->stripePayoutId,
'paidOutAt' => $this->paidOutAt?->format(DateTimeInterface::ATOM),
'payoutSuccessful' => $this->payoutSuccessful,
```
  
also creates an abbreviated view of donations for use with phpmyadmin

### DIFF
--- a/integrationTests/StripePayoutHandlerTest.php
+++ b/integrationTests/StripePayoutHandlerTest.php
@@ -86,6 +86,13 @@ class StripePayoutHandlerTest extends IntegrationTest
         $this->assertSame(SalesforceWriteProxy::PUSH_STATUS_PENDING_UPDATE, $donationFetchedFromDB->getSalesforcePushStatus());
 
         $this->assertSame(
+            1598535656, // timestamp from po.json
+            $donationFetchedFromDB->getPaidOutAt()?->getTimestamp()
+        );
+
+        $this->assertSame('po_externalId_123', $donationFetchedFromDB->getStripePayoutId());
+
+        $this->assertSame(
             <<<"LOGS"
                 info: Payout: Getting all charges related to Payout ID po_externalId_123 for Connect account ID acct_unitTest123
                 info: Payout: Getting all Connect account paid Charge IDs for Payout ID po_externalId_123 complete, found 1
--- a/integrationTests/StripePayoutHandlerTest.php
+++ b/integrationTests/StripePayoutHandlerTest.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace MatchBot\Tests\Application\Messenger\Handler;
+namespace MatchBot\IntegrationTests;
 
 use DI\Container;
 use Doctrine\ORM\EntityManagerInterface;
--- a/integrationTests/StripePayoutHandlerTest.php
+++ b/integrationTests/StripePayoutHandlerTest.php
@@ -35,141 +35,21 @@ class StripePayoutHandlerTest extends IntegrationTest
     private const string CONNECTED_ACCOUNT_ID = 'acct_unitTest123';
     private const string DEFAULT_PAYOUT_ID = 'po_externalId_123';
     private const string RETRIED_PAYOUT_ID = 'po_retrySuccess_234';
+    private TestLogger $testLogger;
 
-    public function testUnrecognisedChargeId(): void
+    #[\Override]
+    public function setUp(): void
     {
-        $this->markTestSkipped();
-        $this->getApp();
-        $container = $this->getContainer();
+        parent::setUp();
+        $this->testLogger = new TestLogger();
 
-        $balanceTxnResponse = $this->getStripeHookMock('ApiResponse/bt_invalid');
-        $chargeResponse = $this->getStripeHookMock('ApiResponse/ch_list_with_invalid');
-
-        $entityManagerProphecy = $this->prophesize(EntityManagerInterface::class);
-        $entityManagerProphecy->beginTransaction()->shouldBeCalledOnce();
-
-        // We call these once per whole loop, now we use txns, to ensure it's closed cleanly.
-        $entityManagerProphecy->commit()->shouldBeCalledOnce();
-        $entityManagerProphecy->flush()->shouldBeCalledOnce();
-
-        $stripeBalanceTransactionProphecy = $this->prophesize(BalanceTransactionService::class);
-        $stripeBalanceTransactionProphecy->all(
-            [
-                'limit' => 100,
-                'payout' => self::DEFAULT_PAYOUT_ID,
-            ],
-            ['stripe_account' => self::CONNECTED_ACCOUNT_ID],
-        )
-            ->shouldBeCalledOnce()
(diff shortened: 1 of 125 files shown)


schema/donation-summary-no-orm.sql
```diff
@@ -0,0 +1,8 @@
+-- Auto generated file, do not edit.
+-- run ./matchbot matchbot:write-schema-files to update
+
+-- Note that CHARSET and COLLATE details have been removed to allow sucessful comparisons between schemas
+-- on dev machines and CircleCI. Do not use these files to create a database. They are provided for
+-- information only.
+
+CREATE ALGORITHM=UNDEFINED DEFINER=`root`@`%` SQL SECURITY DEFINER VIEW `donation-summary-no-orm` AS select `Donation`.`id` AS `id`,`Donation`.`salesforceId` AS `salesforceId`,`Donation`.`createdAt` AS `createdAt`,`Donation`.`amount` AS `amount`,`Donation`.`donationStatus` AS `donationStatus`,`Charity`.`name` AS `CharityName`,`Campaign`.`name` AS `CampaignName` from ((`Donation` join `Campaign` on((`Campaign`.`id` = `Donation`.`campaign_id`))) join `Charity` on((`Charity`.`id` = `Campaign`.`charity_id`)))
```
bdsl, line 8: `no-orm` in the name was intended to allow us to have doctrine filter it out when checking for migrations etc, but it turns that's not necessary - the schema introspection ignores everything other than 'base tables', i.e. ignores views by default. See implementation: https://github.com/doctrine/dbal/blob/3.9.x/src/Platforms/AbstractMySQLPlatform.php#L384

If we did need to filter stuff ourselves we can do it by passing a predicate that takes a table name to `\Doctrine\DBAL\Configuration::setSchemaAssetsFilter`, or by using `--filter-expression=FILTER-EXPRESSION` CLI argument for vendor/bin/doctrine-migrations migrations:diff

bdsl: slightly annoying that mysql seems to always output view create table statements one line like this, but not that worthwhile to fix.

src/Domain/Donation.php
```diff
@@ -367,6 +368,20 @@ class Donation extends SalesforceWriteProxy
     #[ORM\Column(nullable: true)]
     private ?DateTimeImmutable $preAuthorizationDate = null;
 
+    /**
+     * Stripe payout used to pay this donation out to recipient charity. Null on
+     *  all donations from before May 2025.
+     */
+    #[ORM\Column(nullable: true)]
+    private ?string $stripePayoutId = null;
```
NoelLH, line 376: What do you think about also adding a payout status, to help with `payout.failed` scenarios and help us see consistent info across systems if a payout bounces?

As https://docs.stripe.com/api/payouts/object#payout_object-status says,
> Some payouts that fail might initially show as `paid`, then change to `failed`.

It's reasonably rare but we have seen this happen.

integrationTests/StripePayoutHandlerTest.php
```diff
@@ -0,0 +1,204 @@
...
+    private function invokePayoutHandler(
+        Container $container,
+        string $payoutId = self::DEFAULT_PAYOUT_ID,
+    ): void {
+        $payoutHandler = $container->get(StripePayoutHandler::class);
+        $payoutMessage = new StripePayout(connectAccountId: self::CONNECTED_ACCOUNT_ID, payoutId: $payoutId);
+        $payoutHandler($payoutMessage);
+    }
+
+    public function getPersistedCollectedDonation(string $transferId): Donation
```
NoelLH, line 194:
```suggestion
    private function getPersistedCollectedDonation(string $transferId): Donation
```

NoelLH (approved):
LGTM. Some things inline, no blockers.

src/Application/Messenger/Handler/StripePayoutHandler.php
```diff
@@ -119,7 +119,7 @@ public function __invoke(StripePayout $payout): void
             if ($donation->getDonationStatus() === DonationStatus::Collected) {
                 // We're confident to set donation status to paid because this
                 // method is called only when Stripe event `payout.paid` is received.
-                $donation->setDonationStatus(DonationStatus::Paid);
+                $donation->recordPayout($payoutId, $payoutInfo['created']);
```
NoelLH, line 122: I had a closer look at some real payout data and I'm not sure this is the best value to use. It's sort of become less important now that may accounts get accelerated payouts but I don't think this is guaranteed.

`arrival_date` is sort of a date rather than a datetime (although it's still provided as a unix timestamp) but it might still be the better value to use. It seems like payouts which take 5 days have the later ETA as arrival date while those expected to land same day show the beginning of the day they were `created`.

bdsl: Thanks, will have a look at that in the morning. Being a date rather than a datetime reminds me I'm still interested in using https://github.com/brick/date-time especially as temporal progresses through tc39, but not just for this change with no real business logic and slighly less now I see that it's almost a year since brick/date-time has had a release.

src/Domain/Donation.php
NoelLH: We already handle `payout.failed` so I'm not sure I see why we can't keep it up to date.

bdsl: I'm sure we can keep it up to date if we listen to all the events that can change a payout status, I guess I just wasn't sure how valuable that is if it just means clicking through to see our own payout record instead of the stripe one.

If we have the status then it seems like it'd make more sense to have a db table for payouts to join to so we update it in one place instead of on all donations. I don't mind making that, and then I might also think about making a JSON column and just put a copy of everything stripe sends us in there in case we want it.

And probably not but I'm then wondering if there's any concept that's more internal to Big Give that would be better to use that we might translate the stripe payout into, partly so we're not tying our system to stripe's way of thinking more than we have to. If not then I'll just think of it as keeping a copy of some data from stripe.

NoelLH: It seems likely to me that this will go in the charity portal at some point, at which point I think there's definitely some value in being able to show them if we know a payout failed.

I don't know if it's that important to optimise the update itself as it's a rare case. If that's the only reason then I'm not sure we should make it a separate object – the reasoning doesn't seem that different to the payout date[time] field.

bdsl: Thanks, will consider in the morning, maybe we'll pair on this (and the related code for SF)

src/Application/Messenger/Handler/StripePayoutHandler.php
bdsl: Thanks, switched to arrival date.

NoelLH (dismissed):

PR #1376, opened 2025-05-21
MAT-421 – patch regression environment's out of sync fund ID 2

NoelLH:


src/Migrations/Version20250521091607.php
```diff
@@ -0,0 +1,46 @@
...
+
+    public function up(Schema $schema): void
+    {
+        if (getenv('APP_ENV') !== 'regression') {
+            return;
+        }
+
+        $donationId = $this->donationId;
+        $this->addSql("UPDATE Donation SET amount = 1 WHERE id = $donationId LIMIT 1");
+        $this->addSql("UPDATE FundingWithdrawal SET amount = 1 WHERE donation_id = $donationId LIMIT 1");
```
bdsl, line 33: this can only run in regression so doesn't need very careful review - if we merge it and find the alarms there stop that's good.

I checked the donation and the FW in the db both currently have values of £5. I think it's only adjusting the FW that matters for the alarms we're getting but good to do both for consistency. Reducing the FW by £4 does look like it will bring things back into sync.

I think the fact that this is needed probably means the message output by the HandleOutOfSyncFunds could be clearer, but I don't think that's essential to do as part of this ticket.

bdsl (approved):

PR #1381, opened 2025-05-22
MAT-411: Start serving charity campaign to FE from database if available

bdsl:
Not perfect yet but a lot of the way there. The new Campaign and Charity HTTP model classes are fairly direct ports of the equivilent classes that we have in Apex code.


src/Application/Actions/Campaigns/Get.php
```diff
@@ -38,6 +44,12 @@ protected function action(Request $request, Response $response, array $args): Re
 
         $sfId = $args['salesforceId'] ?? throw new HttpNotFoundException($request);
 
+        $campaignFromMatchbotDB = $this->campaignRepository->findOneBySalesforceId(Salesforce18Id::ofCampaign($sfId));
+
+        if ($campaignFromMatchbotDB) {
```
bdsl, line 49: the ticket I'm working on right now is only concerned with the true branch here, but of course we can't break the site in the false branch so then we keep the existing behaviour by proxying from SF.

PR #1382, opened 2025-05-23
MAT-411: Check campaign rendered from matchbot against what SF sent us.

bdsl:
In future we will probably want to generate an alarm and/or throw if this doesn't match. Right now we know it doesn't match so we just include the list of errors in the response, and FE can display it on the donation page (at least outside prod) until we've fixed them all.

To test make a donation on local to force matchbot to pull a campaign from SF, then view the API response at e.g.
http://localhost:30030/v1/campaigns/a05S900000IdO9lIAF

You should see quite a long list of errors something like this. (one of them is because I edited the campaign title in my my matchbot db)

image

My plan next is to make FE display these all near the top of the donate page.

### DIFF
--- a/src/Domain/CampaignRenderCompatibilityChecker.php
+++ b/src/Domain/CampaignRenderCompatibilityChecker.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace MatchBot\Domain;
+
+use Assert\Assert;
+use Assert\LazyAssertion;
+use Assert\LazyAssertionException;
+
+/**
+ * Checks that a campaign as rendered to an array by new matchbot code is compatible with how it was
+ * rendered by the SF API.
+ */
+class CampaignRenderCompatibilityChecker
+{
+    /**
+     * @param array<array-key,mixed> $actual
+     * @param array<array-key,mixed> $expected
+     *
+     * @throws LazyAssertionException
+     */
+    public static function checkCampaignHttpModelMatchesModelFromSF(
+        array $actual,
+        array $expected
+    ): void {
+        $lazyAsert = Assert::lazy();
+
+        self::recursiveCompare(
+            $actual,
+            $expected,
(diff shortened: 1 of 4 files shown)


src/Domain/CampaignRenderCompatibilityChecker.php
```diff
@@ -0,0 +1,57 @@
...
+namespace MatchBot\Domain;
+
+use Assert\Assert;
+use Assert\LazyAssertionException;
+
+/**
+ * Checks that a campaign as rendered to an array by new matchbot code is compatible with how it was
+ * rendered by the SF API.
+ */
+class CampaignRenderCompatibilityChecker
```
bdsl, line 12: this is all very general code, its just checking if everything in one array matches another, so maybe could come from a library but not sure which library and writing from scratch was probably as quick as choosing a library.

Also it may get more complex soon as we might need to include allowances for some differences, e.g. for the campaign values that can change inside matchbot after the campaign is pulled from SF like number of donations.

src/Domain/CampaignRenderCompatibilityChecker.php
```diff
@@ -0,0 +1,57 @@
...
+     * @param array<array-key,mixed> $modelFromMatcbhotDBAsArray
+     * @param array<array-key,mixed> $modelGeneratedFromSF
+     *
+     * @throws LazyAssertionException
+     */
+    public static function checkCampaignHttpModelMatchesModelFromSF(
+        array $modelFromMatcbhotDBAsArray,
+        array $modelGeneratedFromSF
+    ): void {
+        $lazyAsert = Assert::lazy();
```
NoelLH, line 24:
```suggestion
        $lazyAssert = Assert::lazy();
```

NoelLH (approved):

PR #1383, opened 2025-05-23
MAT-395 – fix double donation confirmation

NoelLH:


tests/Application/Actions/Hooks/StripePaymentsUpdateTest.php
```diff
@@ -644,4 +630,33 @@ public function setChargeIdOnDonation(Donation $donation, string $chargeId): voi
...
+        $mock = json_decode(
+            $this->getStripeHookMock($mockName),
+            associative: true,
+            flags: \JSON_THROW_ON_ERROR
+        );
+
+        return $mock;
+    }
+
+    private function sendWebhook(array $data): ResponseInterface
```
NoelLH, line 649: We could definitely use this for a few additional tests too, but I ran out of time today.

tests/Application/Actions/Hooks/StripePaymentsUpdateTest.php
```diff
@@ -134,35 +135,20 @@ public function testMissingSignature(): void
 
     public function testSuccessfulPayment(): void
     {
-        $app = $this->getAppInstance();
         /** @var Container $container */
         $container = $this->getContainer();
 
-        $donation = $this->getTestDonation();
+        // Amounts set to match Stripe mocks' current values
+        $donation = $this->getTestDonation(amount: '6.00', tipAmount: '0.00', collected: false);
```
NoelLH, line 142: I slightly lost the thread of why the non-test changes specifically make this necessary; but it seems like it makes sense for our initial test data and mock webhook responses to align anyway.

NoelLH: Thanks – sorry, my comment / waiting to commit made it unclear, but I think I started with `collected: false` only. But I wasn't initially expecting my changes in particular to make `amount`s matching to the mock necessary. But the tiny mystery was more why the mismatch was allowed before, so I'm not very concerned that they now have to match.

PR #1391, opened 2025-05-28
MAT-405: Calculate campaign amount raised using matchbot db data instead of SF data

bdsl:
…directly

### DIFF
--- a/src/Domain/Donation.php
+++ b/src/Domain/Donation.php
@@ -607,11 +607,9 @@ class Donation extends SalesforceWriteProxy
         $fundingWithdrawalsByType = $this->getWithdrawalTotalByFundType();
 
         $data = [
-            // want total of all three fundingwithdrawls +
             'amountMatchedByChampionFunds' => (float) $fundingWithdrawalsByType['amountMatchedByChampionFunds'],
             'amountMatchedByPledges' => (float) $fundingWithdrawalsByType['amountMatchedByPledges'],
             'amountPreauthorizedFromChampionFunds' => (float) $fundingWithdrawalsByType['amountPreauthorizedFromChampionFunds'],
-
             'billingPostalAddress' => $this->donorBillingPostcode,
             'charityFee' => (float) $this->getCharityFee(),
             'charityFeeVat' => (float) $this->getCharityFeeVat(),


src/Domain/CampaignService.php
```diff
@@ -248,4 +244,15 @@ public function checkCampaignCanBeHandledByMatchbotDB(array $campaignData, Sales
...
+    /**
+     * Gets the amount raised for a given charity campaign, based on donations in the Matchbot DB
+     *
+     * @param Salesforce18Id<Campaign> $campaignId
+     * @return Money
+     */
+    public function amountRaised(Salesforce18Id $campaignId): Money
+    {
+        return $this->campaignRepository->totalAmountRaised($campaignId);
+    }
```
bdsl, line 257: can inline this later and make test be a test for repository

bdsl (commented):
LGTM, just finished pairing with @smalizaidi 

Also looks good to @smalizaidi

src/Domain/CampaignRepository.php
```diff
@@ -289,6 +289,51 @@ public function setFundRepository(FundRepository $fundRepository): void
...
+        Assertion::numeric($donationResult);
+
+        $matchedFundQuery = $this->getEntityManager()->createQuery(
+            <<<'DQL'
+            SELECT COALESCE(SUM(fw.amount), 0) as sum FROM MatchBot\Domain\FundingWithdrawal fw
+            JOIN MatchBot\Domain\Donation donation
+            JOIN MatchBot\Domain\Campaign campaign
+            WHERE campaign.salesforceId = :campaignId
+        DQL
+        );
```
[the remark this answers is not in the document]
NoelLH: Looks like this is an old comment and we do this. LGTM.

src/Domain/CampaignRepository.php
```diff
@@ -289,6 +289,68 @@ public function setFundRepository(FundRepository $fundRepository): void
...
 
+    /**
+     * @param Salesforce18Id<Campaign> $campaignId
+     */
+    public function totalAmountRaised(Salesforce18Id $campaignId): Money
+    {
+        $donationQuery = $this->getEntityManager()->createQuery(
+            <<<'DQL'
+            SELECT donation.currencyCode, COALESCE(SUM(donation.amount), 0) as sum FROM MatchBot\Domain\Donation donation
+            JOIN MatchBot\Domain\Campaign campaign
```
NoelLH, line 300: This feels like it might be worth a small hit in value-object safety to avoid a DB join. If callers passed campaigns' numeric ID, which I think they can get just as easily from memory, we could avoid joining to Campaign. Not a blocker.

src/Domain/CampaignRepository.php
```diff
@@ -289,6 +289,68 @@ public function setFundRepository(FundRepository $fundRepository): void
...
+
+        $donationSum = $donationResult[0]['sum'];
+
+        Assertion::numeric($donationSum);
+
+        $matchedFundQuery = $this->getEntityManager()->createQuery(
+            <<<'DQL'
+            SELECT COALESCE(SUM(fw.amount), 0) as sum FROM MatchBot\Domain\FundingWithdrawal fw
+            JOIN MatchBot\Domain\Donation donation
+            JOIN MatchBot\Domain\Campaign campaign
```
NoelLH, line 333: As above, we could check 2 tables rather than 3 if passing numeric campaign ID.

NoelLH (approved):
LGTM! One non-blocking suggestion.

src/Domain/CampaignRepository.php
bdsl: Yep I think that's a worthwhile optimisation, will do adjust this.

And I don't think I'll do it yet but we could get the similar safety making a value object that wraps an int, could be generic the same way that Salesforce18Id is.

src/Domain/CampaignRepository.php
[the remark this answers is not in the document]
bdsl: right, sorry I should have resolved it or replied to myself before asking you to review.

src/Domain/CampaignService.php
bdsl: or not - followup up adds caching logic to this function.

PR #1400, opened 2025-05-30
MAT-405: Fixes for non-emergency errors seen recently

bdsl:


src/Domain/CampaignRenderCompatibilityChecker.php
```diff
@@ -68,6 +68,22 @@ private static function recursiveCompare(
                 continue;
             }
 
+            if ($key === 'postalAddress') {
+                // postalAddress is required by FE, so not output by matchbot.
```
bdsl, line 72:
```suggestion
                // postalAddress is not required by FE, so not output by matchbot.
```

PR #1407, opened 2025-06-03
Mat 405 stop more warnings

bdsl:


src/Domain/CampaignService.php
```diff
@@ -60,6 +60,13 @@ public function renderCampaign(CampaignDomainModel $campaign): array
         $sfCampaignData = $campaign->getSalesforceData();
         $sfCharityData = $sfCampaignData['charity'];
 
+        try {
+            $websiteURi = $charity->getWebsiteUri()?->__toString();
```
NoelLH, line 64:
```suggestion
            $websiteUri = $charity->getWebsiteUri()?->__toString();
```
Slightly less confusing case, not a blocker.

NoelLH (approved):
LGTM, one optional var case correction

PR #1414, opened 2025-06-05
MAT-405: Stop warning on null title converting to Untitled

bdsl:
Forgot to include this in last change pushed to prod, so still seeing warnings in logs


bdsl:
merged as not worth review

PR #1416, opened 2025-06-05
MAT-405: Link charity campaigns to associated metacampaigns

bdsl:
Kernel of this PR is change from

````
$parentRef = $sfCampaignData['parentRef'];
parentRef: $parentRef,
````

to 

````
parentRef: $campaign->getMetaCampaignSfId()?->value,
````

I think needed since MetaCampaignSfId is such a core concept, and the association between campaign and metacampaign will be needed to implement search that it should be fully represented in the matchbot model and not just use data direct from SF without interpreting it, and it needs to be indexed.

### DIFF
--- a/src/Domain/Campaign.php
+++ b/src/Domain/Campaign.php
@@ -19,7 +19,7 @@ use MatchBot\Client\Campaign as CampaignClient;
  */
 #[ORM\Table]
 #[ORM\Index(name: 'end_date_and_is_matched', columns: ['endDate', 'isMatched'])]
-#[ORM\Index(name: 'metaCampaignSfId', columns: ['metaCampaignSfId'])]
+#[ORM\Index(name: 'metaCampaignSlug', columns: ['metaCampaignSlug'])]
 #[ORM\Entity(repositoryClass: CampaignRepository::class)]
 #[ORM\HasLifecycleCallbacks]
 class Campaign extends SalesforceReadProxy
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: src/Domain/Campaign.php [hunk 1]
```diff
@@ -134,14 +146,16 @@ class Campaign extends SalesforceReadProxy
 
 
     /**
-     * @param \DateTimeImmutable|null $regularGivingCollectionEnd
      * @param Salesforce18Id<Campaign> $sfId
+     * @param Salesforce18Id<MetaCampaign>|null $metaCampaignSfId
```
bdsl, line 150: good use case for the generic phantom type here - SA tools will make sure don't get these two mixed up even though they've got the same type at runtime.

src/Domain/SalesforceProxy.php
```diff
@@ -10,6 +10,7 @@
  * @see SalesforceReadProxy
  * @see SalesforceWriteProxy
  */
+#[ORM\Index(name: 'salesforceId', columns: ['salesforceId'])]
```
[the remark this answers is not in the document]
NoelLH: `unique: true` on the field already creates a unique index for it I believe. Maybe that takes precedence so Doctrine doesn't complain but determines there's nothing additional to do?

bdsl:
> I think needed since MetaCampaignSfId is such a core concept, and the association between campaign and metacampaign will be needed to implement search that it should be fully represented in the matchbot model and not just use data direct from SF without interpreting it, and it needs to be indexed.

And its more immediatly needed in order to render charity campaign api fields parentAmountRaised, parentDonationCount, parentMatchFundsRemaining, parentTarget and parentUsesSharedFunds. Parent means metacampaign.

bdsl:
@bdsl looking at apex implementation parentRef can actually be either an 18 digit ID *or* a slug, just for extra fun.

https://github.com/thebiggive/salesforce/blob/e107a3c131270d9d9058c508a66af61001da407f/force-app/main/default/classes/CampaignService.cls#L381

src/Domain/Campaign.php
```diff
@@ -409,6 +429,10 @@ final public function updateFromSfPull(
         Assertion::nullOrRegex($status, "/^[A-Za-z]{2,30}$/");
         Assertion::betweenLength($name, 2, 255);
         Assertion::nullOrMaxLength($thankYouMessage, 500);
+        Assertion::nullOrBetweenLength($metaCampaignSlug, 1, 64);
+        Assertion::nullOrRegex($metaCampaignSlug, '/^[-a-z50-9]+$/');
```
NoelLH, line 433:
```suggestion
        Assertion::nullOrRegex($metaCampaignSlug, '/^[-a-z0-9]+$/');
```
I think in this case the `5` is harmless, just a little confusing.

I wonder if we should allow uppercase A-Z – I think it may be the only range that is [allowed on the SF field](https://github.com/thebiggive/salesforce/blob/cb006cb0251ffa8856cdea172157029e4422b9a3/force-app/main/default/objects/ChampionFunding__c/validationRules/Slug_Uses_Valid_Characters.validationRule-meta.xml#L7-L8) but not here, but in practice I can't remember seeing an uppercase character used.

NoelLH (dismissed):
Possible tweak inline but no blockers, LGTM

src/Domain/Campaign.php
bdsl: Thanks, 5 removed and uppercase allowed in https://github.com/thebiggive/matchbot/pull/1416/commits/4cf2b4f471176792abc5f179dfa3c5b49e1a66b7

I actually discovered after writing this regex that I'd already created a MetaCampaignSlug VO class and forgoton about it - that has a regex `^[A-Za-z0-9-]+$` so changed to the same thing here, although IMHO it's better that we never use capitals in a slug just in case they ever get matched in a case sensitive way.

Should really DRY up and re-use the regex and slightly different max length from the VO, but not worth changing right now to me. Not using the VO as a property type since embeddables are a bit too awkward with doctrine.

PR #1417, opened 2025-06-05
MAT-427 – re-submit Gift Aid for one campaign

NoelLH:


bdsl (commented):
LGTM. I ran `SELECT * from Campaign where id = 7774;` in prod and can confirm that that is the campaign we're interested in.

PR #1426, opened 2025-06-06
MAT-431: accept new or updated campaign put from salesforce

bdsl:
The salesforce side of the API yet to be built, but the body format is the same as the existing campaign API response, (except wrapped in  `{campaign: }` object to give space for adding extra things into the request body in future that are not part of the campaign)


src/Application/Auth/SalesforceAuthMiddleware.php
```diff
@@ -36,11 +37,25 @@ public function __construct(
...
+        // https://evertpot.com/222/
+        //
+        // This wasn't an issue up till now because all the requests from SF happened to have empty bodies.
+
+        $stream = fopen('php://memory', 'rb+');
+        \assert(\is_resource($stream));
+        fwrite($stream, $content);
+        rewind($stream);
+
+        return $handler->handle($request->withBody(new Stream($stream)));
```
bdsl, line 58: took me a while to work this out. Second time I've been bitten by `$request->getBody()->getContents()` not just taking a copy of the content but removing them from the stream.

Anyway good that I wrote a test that covered this instead of waiting to discover the issue when we actually try to integrate with SF.

src/Application/Actions/Campaigns/GetSummariesForCharity.php
```diff
@@ -39,7 +39,7 @@ public function __construct(
             $args['charitySalesforceId'] ?? throw new HttpNotFoundException($request)
         );
 
-        $charity = $this->charityRepository->findOneBySfIDOrThrow($sfId);
+        $charity = $this->charityRepository->findOneBySalesforceIdOrThrow($sfId);
```
bdsl, line 42: function renamed for consistency with `CampaignRepository::findOneBySalesforceId`

src/Client/Campaign.php
```diff
@@ -33,14 +33,12 @@
...
  *      stripeAccountId: string,
- *      hmrcReferenceNumber: string,
+ *      hmrcReferenceNumber: string|null,
  *      giftAidOnboardingStatus: string,
- *      regulator: string,
  *      regulatorRegion: string,
  *      regulatorNumber: string,
  *     },
  *     endDate: ?string,
- *     feePercentage: ?float,
```
bdsl, line 43: I added type to a campaign constant array in test so SA tools made me fix this. Was going to give up on Psalm's unreadable error message but then PHPStan told me what was wrong one line at a time.

src/Client/Campaign.php
```diff
@@ -33,14 +33,12 @@
  *      instagram: ?string,
  *      optInStatement: ?string,
  *      stripeAccountId: string,
- *      hmrcReferenceNumber: string,
+ *      hmrcReferenceNumber: string|null,
  *      giftAidOnboardingStatus: string,
- *      regulator: string,
```
bdsl, line 38: this was here by mistake, SF does not send regulator

src/Client/Campaign.php
bdsl, line 43: `feePercentage` was removed from the API last year.

integrationTests/AcceptCampaignPushFromSFTest.php
bdsl, line 1: glad that I decided to write an integration test here, otherwise I wouldn't have discovered that SalesforceAuthMiddleware needed changes.

NoelLH (dismissed):
I think `rewind()` solves the stream thing more cleanly. Otherwise LGTM.

smalizaidi (approved):
LGTM

PR #1429, opened 2025-06-09
MAT-405: Calculate metacampaign amount raised without using donation …

bdsl:
…records from SF

We do however need to pull in any adjustment amount from SF on the metacampaign object.

Also need to check if it's correct that we're including gift aid here and not including gift aid on the amount raised for an individual charity campaign.

Need to add tests before releasing this


NoelLH:
> Also need to check if it's correct that we're including gift aid here and not including gift aid on the amount raised for an individual charity campaign.

I think this is at least deliberate, and probably the most correct thing we can do relative to how we present the numbers to donors.

AFAIK this is what is desired because:
* For charity campaigns, we want to be able to show progress towards the target and usage of match funds very clearly, and Gift Aid makes this harder for charities and donors to predict and follow.
* That's not a factor for the meta-campaign amount raised, where we don't compare to a target and would rather show the entire amount going to the campaign's charities; we also explicitly label this "inc. Gift Aid" in the most prominent figure at the top of meta-campaigns, for the avoidance of doubt and to help understand any difference from the sum of charity campaigns' totals.

schema/MetaCampaign.sql
```diff
@@ -20,6 +20,8 @@ CREATE TABLE `MetaCampaign` (
   `isEmergencyIMF` tinyint(1) NOT NULL,
   `salesforceLastPull` datetime DEFAULT NULL,
   `salesforceId` varchar(18) DEFAULT NULL,
+  `total_adjustment_amountInPence` int NOT NULL,
+  `total_adjustment_currency` varchar(255) NOT NULL,
```
NoelLH, line 24: I missed this when we first used the `Money` embeddable, but what do you think about giving it `length: 3`?

src/Domain/MetaCampaignRepository.php
```diff
@@ -45,4 +46,74 @@ public function countCompleteDonationsToMetaCampaign(MetaCampaign $metaCampaign)
...
+     * Returns the total of all the complete donations to this metacampaign, including matching and including gift aid,
+     * and any "offline" donations or adjustments.
+     */
+    public function totalAmountRaised(MetaCampaign $metaCampaign): Money
+    {
+        $donationQuery = $this->em->createQuery(
+            <<<'DQL'
+            SELECT donation.currencyCode, COALESCE(SUM(
+            donation.amount + 
+            if(donation.giftAid, donation.amount * 0.25, 0)
```
NoelLH, line 60: Could use `Donation::GIFT_AID_PERCENTAGE`?

And/or, wondering if there's possibly an argument that a database level view with a `giftAidAmount` could be helpful here, to:
1. reduce on-the-fly calculation generally, and
2. make it less painful to denormalise differently should the Gift Aid rate change

src/Domain/MetaCampaignRepository.php
```diff
@@ -45,4 +46,74 @@ public function countCompleteDonationsToMetaCampaign(MetaCampaign $metaCampaign)
...
+            $donationSum = $donationResult[0]['sum'];
+        }
+
+        Assertion::numeric($donationSum);
+
+        $matchedFundQuery = $this->em->createQuery(
+            <<<'DQL'
+            SELECT COALESCE(SUM(fw.amount), 0) as sum
+            FROM MatchBot\Domain\FundingWithdrawal fw
+            JOIN fw.donation donation JOIN d.campaign c
```
NoelLH, line 95:
```suggestion
            JOIN fw.donation donation JOIN donation.campaign c
```

NoelLH (commented):
LGTM so far apart from what I think is a DQL typo. Commented on a couple of other minor things inline.

bdsl:
> > 
> 
> I think this is at least deliberate, and probably the most correct thing we can do relative to how we present the numbers to donors.

Right, and certainly we shouldn't deviate from what SF does during the process of switching over from SF to MB backend. I think there is probably still quite a bit of scope for confusion here particularly when using shared funds and (as will often go together) when we have a one-campaign meta-campaign (i.e. raising money for DEC to distribute), where we do show the `parentTarget` etc.

Makes me think again maybe need to find or create a place to document semantics like this - I'll do briefly within matchbot docblocks for now on this PR but since like a lot of things this is used across multiple repos that might not be the best place. Could be OpenAPI docs but not everything we'd want to document is part of an API and those feel like quite a lot of work to create and edit if done manually, so might be better as a simple markdown table or something.

schema/MetaCampaign.sql
bdsl: Yep good idea, done in 9aed4cd8

src/Domain/MetaCampaignRepository.php
```diff
@@ -45,4 +46,77 @@ public function countCompleteDonationsToMetaCampaign(MetaCampaign $metaCampaign)
...
+            SELECT donation.currencyCode, COALESCE(SUM(
+            donation.amount + 
+            (CASE WHEN donation.giftAid = 1 THEN donation.amount * :giftAidPercent / 100 ELSE 0 END)))
+             as sum
+            FROM MatchBot\Domain\Donation donation JOIN donation.campaign c
+            WHERE c.metaCampaignSlug = :slug 
+            AND donation.donationStatus IN (:succcessStatus)
+            GROUP BY donation.currencyCode
+        DQL
+        );
```
bdsl, line 69: dumping the SQL version of above gives:

```sql
SELECT d0_.currencyCode AS currencyCode_0, COALESCE(SUM(d0_.amount + (CASE WHEN d0_.giftAid = 1 THEN d0_.amount * ? / 100 ELSE 0 END))) AS sclr_1 
FROM Donation d0_ 
INNER JOIN Campaign c1_ ON d0_.campaign_id = c1_.id 
WHERE c1_.metaCampaignSlug = ? 
AND d0_.donationStatus IN (?) GROUP BY d0_.currencyCode
```

integrationTests/MetaCampaignRepositoryTest.php
```diff
@@ -0,0 +1,119 @@
...
+        $this->metaCampaign = TestCase::someMetaCampaign(false, false);
+        $this->campaign = TestCase::someCampaign(metaCampaignSlug: $this->metaCampaign->getSlug());
+
+        $this->fund = new Fund(
+            currencyCode: 'GBP',
+            name: 'Test Match Fund',
+            salesforceId: Salesforce18Id::ofFund(TestCase::randomString()),
+            fundType: FundType::ChampionFund
+        );
+
```
NoelLH:
```suggestion
```
I think we can maybe lose the 2nd of these?

NoelLH (dismissed):
LGTM. One possible bit of test code that I think can be tidied, no blockers

PR #1436, opened 2025-06-11
Add logs to help diagnose intermittent missing register links in donation thanks email

NoelLH:
And a little minor tidying that's not strictly related

### DIFF
--- a/src/Domain/DonationNotifier.php
+++ b/src/Domain/DonationNotifier.php
@@ -4,8 +4,8 @@ namespace MatchBot\Domain;
 
 use MatchBot\Application\Assertion;
 use MatchBot\Application\Email\EmailMessage;
-use MatchBot\Application\Environment;
 use MatchBot\Client\Mailer;
+use Psr\Log\LoggerInterface;
 
 class DonationNotifier
 {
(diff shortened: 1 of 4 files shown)


src/Domain/DonationNotifier.php
```diff
@@ -123,6 +124,21 @@ public function notifyDonorOfDonationSuccess(
...
+                $this->logger->info(sprintf(
+                    'No email verification token found for donation %s for email %s using cutoff %s',
+                    $donation->getId() ?? '[unknown]',
+                    $emailAddress->email,
+                    $this->now->format('c'),
+                ));
+            }
+        } else {
+            $this->logger->info(sprintf(
+                'No $sendRegisterUri so not checking for email verification token for donation %s for email %s',
```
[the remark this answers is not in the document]
bdsl: just to clarify that it's a bool not a URI or something. Not critical of course.

src/Domain/DonationService.php
```diff
@@ -878,7 +883,7 @@ private function shouldInviteRegistration(Donation $donation): bool
     {
         $donorId = $donation->getDonorId();
         if (!$donorId) {
-            // must be an old donation
+            // must be an already persisted donation
```
bdsl, line 886: I think old did mean old here - a donation from before we started writing to the `donorUUID` for every donation.

bdsl (commented):
I'm not sure about the change to the `must be an old donation` comment, everything else here LGTM.

src/Domain/DonationService.php
NoelLH: Thanks, put that back how it was

PR #1440, opened 2025-06-12
Add guidelines for Junie AI agent

bdsl:
I made Junie write its own guidelines by coyping the prompt suggested at https://www.jetbrains.com/guide/ai/article/junie/intellij-idea/

> Analyze the project structure and tech stack, and create a .junie/guidelines.md file with concise, well-structured information to help new developers.
> Include guidance on organizing the structure, running tests, executing scripts, and following best practices. Keep the content short, clear, and practical.

### DIFF
--- a/.junie/guidelines.md
+++ b/.junie/guidelines.md
@@ -1,0 +1,75 @@
+# MatchBot Developer Guidelines
+
+## Project Overview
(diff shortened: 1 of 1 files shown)


bdsl:
This looks like it should be useful, although my I wonder if it's a bit unhelpfully duplicating content from the readme file etc and will get outdated in the end.

Since the new generated file is very much human readable I'm not sure if there's actually a good reason for it to be a distinct file just meant for Junie instead of something that's intended for both Junie and humans to read so that both can get updates together.

PR #1449, opened 2025-06-23
MAT-405: Start pulling needed fields for calculating targets from SF + calculate and render `parentTarget`

bdsl:
Calculating and rendering the target on the individual charity campaign is still to do, but should now be straightforward to port the logic from that part of `Campaign_Target__c` in SF.

To make this useful we need to start sending the new fields from SF, but that doesn't have to be before this is deployed:

```
 *     imfCampaignTargetOverride?: float, // for meta campaign
 *     totalFundingAllocation?: float, // for charity campaign
 *     amountPledged?: float, // for charity campaign
 *     totalMatchedFundsAvailable?: float, // for meta campaign
 ```
 
 So far the meta campaign figures are used to calculate an output, the charity campaign figures are just saved to the DB but can be used for a calculation in future.


bdsl:
I did have some doubt after coding the change here about whether it was actually necessary to re-implement the logic from Campaign_Target__c or maybe the ticket MAT-435 was wrong and it would have actually been fine to just save the output of that field as fetched from SF, on the basis that although its a formula its only referring to fields within the same object.

But I think now it maybe was right to re-implement, just because of the case where Master_Campaign__r.Is_Emergency_IMF__c is true - in that case a change to the meta campaign would need to be reflected in the output for the related charity campaign, so if we didn't re-implement the formula in matchbot we'd need to make SF send an update for the charity campaign when the meta campaign was edited.

PR #1458, opened 2025-06-24
MAT-436: Make Greggs Foundation champion name fix more permanent

bdsl:


bdsl:
thanks will merge in the morning.

PR #1461, opened 2025-06-25
MAT-413 – add CampaignStatistics for search support etc.

NoelLH:
Remaining todo for these to be in full service, end of 26/6 (plus one added 3/7):
- [x] Use the stats when rendering a campaign, rather than recalculating on demand
- [x] Support ordering by the stats when searching
- [x] Account for refunds – either recalculate when one happened recently or have a less frequent fallback task
- [x] Ideally, tidy up the test helper that has a `@todo` on it
- [x] Require app statuses accepted/approved for inclusion, if there's a metacampaign slug
- [x] Add a 'mop up' task to find and fill in stats for all campaigns which either (a) have none, or (b) have donations after their last stats update and have no recent stats updates


app/repositories.php
```diff
@@ -46,6 +47,8 @@
             return $repo;
         },
 
+        CampaignStatisticsRepository::class => static fn(ContainerInterface $c): CampaignStatisticsRepository => new CampaignStatisticsRepository($c->get(EntityManagerInterface::class)),
```
bdsl, line 50: Not a problem to have this but I think it's not necessary - the DI container will do this automatically with reflection if we don't have this line. We just need to list things where the way to instantiate is less obvious.

bdsl (commented):
Added comments inline - I think the only blocker is that we're including match funds used for pending donations in the total but we're not updating when donations either start being pending or have their funds expire, so its going to be misleading especially if there's a big pending donation at the time stats are calculated and then that is abandoned or cancelled and there are no donations for a long time. 

I guess once we have these stats we're likely to use them not just for search but also for display, replacing the cached stats we have now in Redis.

src/Application/Commands/UpdateCampaignDonationStats.php
```diff
@@ -0,0 +1,46 @@
...
+
+    #[\Override]
+    protected function doExecute(InputInterface $input, OutputInterface $output): int
+    {
+        $campaigns = $this->campaignRepository->findAllWithRecentDonations();
+
+        foreach ($campaigns as $campaign) {
+            $campaignId = $campaign->getId();
+            \assert($campaignId !== null);
+            $matchFundsUsed = $this->campaignRepository->totalMatchFundsUsed($campaignId);
```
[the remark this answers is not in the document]
NoelLH: Shouldn't `donation.donationStatus IN (:succcessStatus)` mean that it's not including funds for pendings? If so, I think it also removes the need to consider routine matching expiry – although it's true that really we should be factoring in refunds reasonably soon instead of only after the next Collected donation.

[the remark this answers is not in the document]
bdsl: you're right, I missed that we were only including successful donations there. We could add a check for a recent `refundedAt` date to the query but I think the extra cost of just looking at `updatedAt` is probably low enough that it might be worth doing it that way for simplicity.

src/Domain/CampaignRepository.php
```diff
@@ -335,28 +338,46 @@ public function totalAmountRaised(int $campaignId): Money
...
-        $matchedFundQuery = $this->getEntityManager()->createQuery(
+        if ($matchFundsUsed === null) {
+            $matchFundsUsed = $this->totalMatchFundsUsed($campaignId);
+        }
+
+        $total = \bcadd($donationSum, $matchFundsUsed->toNumericString(), 2);
+
+        $currency = Currency::fromIsoCode($donationResult[0]['currencyCode']);
+
+        return Money::fromNumericString($total, $currency);
```
bdsl, line 349: Can make this slightly shorter (and DRY up '2' and use of bcadd, which should increase the mutation coverage metric) by converting to money object before adding:


```suggestion
        $donationSum = Money::fromNumericString($donationSum, Currency::fromIsoCode($donationResult[0]['currencyCode']));
        
        return $donationSum->plus($matchFundsUsed);
```

bdsl: I realise that the use of bcadd here is just mobed from elsewhere in the file. I think it's probably worth trying to keep money arithmetic within the money class.

src/Domain/CampaignStatisticsRepository.php
```diff
@@ -0,0 +1,54 @@
...
+            $this->em->persist($statistics);
+        }
+    }
+
+    /**
+     * @return array{amountRaised: Money, matchFundsUsed: Money}
+     */
+    public function getStatistics(Campaign $campaign): array
+    {
+        $statistics = $this->doctrineRepository->findOneBy(['campaign' => $campaign]);
```
bdsl, line 40: I don't really get why this function is doing anything more than
`return $this->doctrineRepository->findOneBy(['campaign' => $campaign]);`

The array shape return doesn't seem particularly easier to use than the object would be.

If we return the stats directly as given by the doctrineRepository it would have to be nullable (unless we add an assert that its not null) but I think that's correct. If there's no stats in the db it doesn't mean there's been no donations it just means we haven't calculated stats yet. Might be an old campaign from before this code was deployed.

I guess for new campaigns we do mostly know the stats will be (temporarily) close enough to zero if non existent since we create stats within a minute after each donation is confirmed - so maybe inside here it would make more sense to construct, and return a stats object with zeros instead of returning an array?

src/Domain/CampaignStatisticsRepository.php
```diff
@@ -0,0 +1,54 @@
...
+
+        if (!$statistics) {
+            return [
+                'amountRaised' => Money::zero(),
+                'matchFundsUsed' => Money::zero(),
```
bdsl, line 45:
```suggestion
                'amountRaised' => Money::zero($campaign->getCurrency()),
                'matchFundsUsed' => Money::zero($campaign->getCurrency()),
```

(Campaign::getCurrency will need implementing but should be trivial)

src/Domain/Donation.php
```diff
@@ -89,7 +89,8 @@ class Donation extends SalesforceWriteProxy
     /**
      * @var string|null PSP's transaction ID assigned on their processing.
      *
-     * In the case of stripe (which is the only thing we support at present, this is the payment intent ID)
+     * In the case of stripe (which is the only thing we support at present, this is the payment intent ID).
+     * May change after initial assignment until donation's Collected, in some edge cases.
```
bdsl, line 93: :+1:  Useful comment about stripe.

Not suggesting doing it on this PR but I think I'd support going further and renaming to `stripePaymentIntentId`. If we support any other PSP in future then the transactionIds of that PSP won't really be usable interchangeably with the stripe one so might as well be on a different field.

app/repositories.php
NoelLH: Yep true, removed in 2c8bc50e37e03aa82d2266fff210d0044a5e0942

src/Domain/CampaignStatisticsRepository.php
NoelLH: Good point about the array shape, replaced with returning the model in 35160bfef360d6572142b756b74ec18798eb8aaa

Yeah, the stats are non-realtime anyway so I think it's more helpful to start with zeroes of the specific Money values than a null object. I guess I'm working on the assumption that we will fill in stats for old campaigns at some point for viewing historical campaign too.

src/Domain/CampaignStatisticsRepository.php
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * With figures all zero if no donations or match funds used yet *or* if first stats calculation
+     * has not yet run.
+     */
+    public function getStatistics(Campaign $campaign): CampaignStatistics
+    {
+        $statistics = $this->doctrineRepository->findOneBy(['campaign' => $campaign]);
+
+        if (!$statistics) {
+            return new CampaignStatistics($campaign, Money::zero(), Money::zero());
```
bdsl, line 47: can we return zeroes of the currency from the campaign instead of zero pounds? Maybe we shouldn't have a default currency on the zero function since its too easy to forget to specify a currency here.

src/Domain/CampaignStatisticsRepository.php
NoelLH: Done in the new code counterpart

src/Domain/CampaignRepository.php
NoelLH: Thanks, done in 6e5f2e022c73d026c1a8d6c670a1340a287d75f8

src/Domain/CampaignStatistics.php
```diff
@@ -0,0 +1,67 @@
...
+
+    #[ORM\OneToOne]
+    #[ORM\Id]
+    private Campaign $campaign;
+
+    #[ORM\Column(length: 18, unique: true)]
+    protected string $campaignSalesforceId;
+
+    #[ORM\Embedded(columnPrefix: 'amount_raised_')]
+    private Money $amountRaised;
```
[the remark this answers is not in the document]
NoelLH: Thanks, I did this alongside starting to use the model for real in https://github.com/thebiggive/matchbot/pull/1461/commits/15ae63574b4b7c83550db19bed667774210d662f

I also added total match funds for similar reasons, so we can phase out the separate Redis cache for charity campaigns.

src/Domain/CampaignRepository.php
bdsl: 6e5f2e022c73d026c1a8d6c670a1340a287d75f8 LGTM.

app/repositories.php
```diff
@@ -10,6 +10,7 @@
 use MatchBot\Domain\CampaignFunding;
 use MatchBot\Domain\CampaignFundingRepository;
 use MatchBot\Domain\CampaignRepository;
+use MatchBot\Domain\CampaignStatisticsRepository;
```
bdsl, line 13:
```suggestion
```

src/Domain/CampaignStatistics.php
```diff
@@ -0,0 +1,125 @@
...
+
+    /** Set on construct and updated when donations change */
+    #[ORM\Embedded(columnPrefix: 'match_funds_used_')]
+    private Money $matchFundsUsed;
+
+    /**
+     * @param Campaign $campaign
+     * @param Money $amountRaised
+     * @param Money $matchFundsUsed
+     * @param Money $matchFundsTotal
```
bdsl, line 60:
```suggestion
     * @param Money $matchFundsTotal
     *
     * $amountRaised must be equal to $matchFundsUsed + $donationSum`
```

src/Domain/CampaignStatistics.php
```diff
@@ -0,0 +1,125 @@
...
+        Money $donationSum,
+        Money $amountRaised,
+        Money $matchFundsUsed,
+        Money $matchFundsTotal,
+    ) {
+        Assertion::greaterOrEqualThan($matchFundsTotal->toNumericString(), $matchFundsUsed->toNumericString());
+        Assertion::eq(
+            $amountRaised->toNumericString(),
+            $donationSum->plus($matchFundsUsed)->toNumericString(),
+        );
```
bdsl, line 73: this eq is good - IMHO it would be worth making this object enforce that invariant consistently at all times. That means merging three setters (setAmountRaised, setMatchFundsUsed, and setDonationSum) into one function that sets all three together, and throws if invariant doesn't hold. Also I see setDonationSum doesn't exist which I think is an oversight and a bug where if not fixed the donationSum will always be left at zero or whatever it was set when the stats object were first generated for each campaign.

src/Domain/CampaignStatisticsRepository.php
```diff
@@ -0,0 +1,65 @@
...
+        Campaign $campaign,
+        Money $donationSum,
+        Money $amountRaised,
+        Money $matchFundsUsed,
+    ): void {
+        $statistics = $this->doctrineRepository->findOneBy(['campaign' => $campaign]);
+
+        if ($statistics) {
+            $statistics->setAmountRaised($amountRaised);
+            $statistics->setMatchFundsUsed($matchFundsUsed);
```
bdsl, line 37: this looks like its missing a line to set the `donationSum` . Also I would use a single function on stats to set three stats at once because then it can enforce the invariant internally. Something like this:


```suggestion
            $statistics->setDonationTotals(amountRaised: $amountRaised, matchFundsUsed: $matchFundsUsed, donationSum: $donationSum);
```

src/Domain/CampaignRepository.php
```diff
@@ -569,7 +606,8 @@ public function search(
         $qb = $this->getEntityManager()->createQueryBuilder();
 
         $safeSortField = match ($sortField) {
-            'matchFundsRemaining' => throw new \Exception('Sorting by matchFundsRemaining not yet implemented'),
```
bdsl, line 572: what's the reason for deleting `matchFundsRemaining` here? Don't we still want that as a form of to-do comment to remember that we have to implement that as its expected by the FE?

bdsl (commented):
Review done, all LGTM other than where I've commented inline.

The only blockers now are about `donationSum` which it looks like we're never updating ( https://github.com/thebiggive/matchbot/pull/1461#discussion_r2189668953 ) and matchFundsRemaining ( https://github.com/thebiggive/matchbot/pull/1461#discussion_r2189678680 )

src/Domain/CampaignRepository.php
NoelLH: I think I was thinking that `matchFundsUsed` with both sort directions covers the same thing, which is not quite right. I'll think a bit more about how to deal with that – it feels a bit messy to add another field to the stats but perhaps it's a false economy to avoid that redundancy, when setters can guarantee that things add up.

src/Domain/CampaignStatisticsRepository.php
NoelLH: Good catch, thanks! Liking the idea of the combined setter, will do that shortly.

bdsl (dismissed):

PR #1470, opened 2025-06-27
MAT-405: Calculate status of a metacampaign according to the time its…

bdsl:
… viewed

### DIFF
--- a/src/Domain/MetaCampaign.php
+++ b/src/Domain/MetaCampaign.php
@@ -25,6 +25,7 @@ use Psr\Http\Message\UriInterface;
 #[ORM\Index(name: 'hidden', columns: ['hidden'])]
 class MetaCampaign extends SalesforceReadProxy
 {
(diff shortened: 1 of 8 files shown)


src/Domain/MetaCampaign.php
```diff
@@ -251,6 +266,23 @@ public function getStatus(): ?string
     {
         return $this->status;
     }
+    
+    public function getStatusAt(\DateTimeImmutable $date): ?string
+    {
```
bdsl, line 271: port of Apex code from https://github.com/thebiggive/salesforce/blob/eef73c195e06f7365988b65b28095e4d23138c63/force-app/main/default/classes/CampaignService.cls#L217

smalizaidi (approved):
LGTM

smalizaidi:
There are some checks that are failing at the moment.

bdsl:
> There are some checks that are failing at the moment.

I think that was only the code coverage checks, which we've never really consistently tried to keep passing. It isn't required for merge.

PR #1471, opened 2025-06-27
MAT-405: Reject messages from SF if timestamped and time mismatches b…

bdsl:
…y over 2 minutes

Could indicate bad clock skew between Salesforce and Matchbot, or an attacker somehow getting hold of an old message and replaying it


smalizaidi (approved):
LGTM

PR #1472, opened 2025-06-30
MAT-433: Accept a list of charities from to update instead of just a …

bdsl:
…single charity

Code copied from \MatchBot\Application\Actions\Charities\Put to new \MatchBot\Application\Actions\Charities\UpsertMany

We may DRY up in future by deleting one or the other controller, probably the older one that only accepts a single charity at a time.


smalizaidi (approved):
LGTM, only there is an error in Circle CI.

PR #1478, opened 2025-07-03
MAT-439 – make single Campaign get by ID from SF more quiet & resilient

NoelLH:


src/Client/Campaign.php
NoelLH, line 1: This class largely reworked by Copilot agent to my instructions. I added more `@throws` info and slightly tweaked its suggested logging.

src/Client/Campaign.php
```diff
@@ -107,29 +108,51 @@ class Campaign extends Common
      * @param string $id
      * @return SFCampaignApiResponse Single Campaign response object as associative array
      * @throws NotFoundException if Campaign with given ID not found
+     * @throws TransferException if connection or request fails twice
+     * @throws \Exception if our loop logic or Guzzle has an unexpexcted case we missed
```
bdsl, line 112: not sure its worth declaring that we throw \Exception if there's a bug - I think that's always true and not something we'd ever catch except in a top level, and makes PhpStorm think we should catch it:

image

src/Client/Campaign.php
```diff
@@ -107,29 +108,51 @@ class Campaign extends Common
...
      */
     public function getById(string $id, bool $withCache): array
```
bdsl, line 114: retry loop looks good, although afaik at when MAT-405 is done the only reason we'll have to ever fetch a campaign from SF is to show the early preview, which is not that critical and might not need a retry system. No harm in having it though.

bdsl (approved):
LGTM, a couple of comments inline, no blockers.

src/Client/Campaign.php
NoelLH: Yeah not critical, but I think it's probably good to keep charity previews as stable as we can.

bdsl (approved):

src/Client/Campaign.php
NoelLH: Removed in d8bc2962a99f8d0a71494b29571f27925eb00a16

PR #1479, opened 2025-07-03
MAT-405: Implement counting campaigns in a metacampaign, using logic …

bdsl:
…ported from SOQL to DQL

Will require sending two new fields as part of the campaign from SF, relatedApplicationStatus and relatedApplicationCharityResponseToOffer


integrationTests/CampaignRepositoryTest.php
```diff
@@ -35,11 +35,13 @@ public function testItFindsANineMonthOldCampaignForACharityAwaitingGiftAidApprov
...
+            relatedApplicationStatus: null,
+            relatedApplicationCharityResponseToOffer: null,
             regularGivingCollectionEnd: null,
+            totalFundraisingTarget: Money::zero(),
             thankYouMessage: null,
             rawData: [],
-            hidden: false,
-            totalFundraisingTarget: Money::zero(),
+            hidden: false
         );
```
bdsl, line 45: while adding the new fields also re-ordering arguments to match formal param order.

PR #1488, opened 2025-07-08
BG2-2909, BG2-2906: Resubmit GA claims for two charities

bdsl:


NoelLH (approved):
Cross checked data, all LGTM

PR #1489, opened 2025-07-08
MAT-405: In non-prod always and only serve campaigns based on matchbo…

bdsl:
…t DB content, not SF

Plan is to do the same in prod in the near future.


smalizaidi (approved):
LGTM

PR #1503, opened 2025-07-10
MAT-420 – retry setting Mandate SF push status fields when necessary

NoelLH:


src/Application/Messenger/Handler/MandateUpsertedHandler.php
```diff
@@ -87,6 +88,55 @@ public function __invoke(MandateUpserted $message): void
...
+
+        do {
+            try {
+                if ($tries > 0) {
+                    /** @psalm-suppress InvalidCast There's a bug analysing do/while w.r.t. $tries */
+                    $this->logger->info("Retrying setting Salesforce fields for mandate $uuid after $tries tries");
+                }
+                $this->setSalesforceFields($uuid, $salesforceId);
+                return;
+            } catch (DBALException\RetryableException $exception) {
```
bdsl, line 113: yep makes sense to catch any RetryableException - I guess the message really only applies if its LockWaitTimeoutException rather than a DeadlockException but not really an issue.

PR #1510, opened 2025-07-15
MAT-413 campaign search: various sort & stats generation fixes

NoelLH:


src/Application/Actions/Campaigns/Search.php
```diff
@@ -13,6 +13,10 @@
 use Psr\Log\LoggerInterface;
 use Slim\Exception\HttpBadRequestException;
 
+/**
+ * Frontend will typically default searches with a `?term` to 'relevance' order and others to 'distanceToTarget'.
+ * No defaults are set server-side any more.
```
bdsl, line 18:
```suggestion
```

src/Domain/Campaign.php
```diff
@@ -175,6 +175,18 @@ class Campaign extends SalesforceReadProxy
     #[ORM\Embedded(columnPrefix: 'total_fundraising_target_')]
     private Money $totalFundraisingTarget;
 
+    /**
+     * Optional BG-defined default sort override for the metacampaign grid.
```
bdsl, line 179: worth adding a bit more docs? Does higher make it go up or down the grid? What does null do?

integrationTests/CampaignStatsAfterDonationCollectedTest.php
```diff
@@ -140,8 +141,8 @@ private function runStatsUpdateCommand(): void
         $application = $this->buildMinimalApp($lockFactory);
         $command = new UpdateCampaignDonationStats(
             $this->getContainer()->get(CampaignRepository::class),
-            $this->getContainer()->get(CampaignStatisticsRepository::class),
             $this->getContainer()->get(EntityManagerInterface::class),
+            $this->getContainer()->get(MatchFundsService::class),
```
bdsl, line 145: not critical but I think we could just ask the container  for a `UpdateCampaignDonationStats` and let it reflect on how to construct it.

src/Application/Commands/UpdateCampaignDonationStats.php
```diff
@@ -58,27 +65,46 @@ private function updateOldMissedCampaigns(OutputInterface $output): void
...
+            'Updated statistics for %d of %d campaigns with no recent stats',
+            $numChanged,
+            count($campaigns),
+        ));
     }
 
-    private function handleCampaign(Campaign $campaign, OutputInterface $output): void
+    /**
+     * Creates or finds + updates a {@see CampaignStatistics} record, via eager loading from $campaign.
+     * Doesn't flush, so callers need to when done building stats.
```
bdsl, line 85:
```suggestion
     * Doesn't flush, so callers need to when done building stats.
     *
     * @return bool whether or not the statistics have changed
```

bdsl (dismissed):
LGTM. I added a few suggestions inline.

src/Domain/Campaign.php
NoelLH: Done in 53d593d1e4221bbff163abd65722377c941075a0

PR #1511, opened 2025-07-16
MAT-442 – set Fund slugs and filter search results by them when appropriate

NoelLH:
And a couple of other minor typo & tech debt fixes

### DIFF
--- a/src/Domain/DonationService.php
+++ b/src/Domain/DonationService.php
@@ -183,10 +183,7 @@ class DonationService
             // it is the one place outside of `UpdateCampaigns` where we clear the whole
             // result cache. It's currently the only user-invoked or single item place where
             // we do so.
-            $resultCache = $this->entityManager->getConfiguration()->getResultCache();
-            if ($resultCache) {
-                $resultCache->clear();
-            }
+            $this->entityManager->getConfiguration()->getResultCache()?->clear();
         }
 
         if ($donationData->currencyCode !== $campaign->getCurrencyCode()) {
--- a/src/Domain/DonationService.php
+++ b/src/Domain/DonationService.php
@@ -183,12 +183,10 @@ class DonationService
             // it is the one place outside of `UpdateCampaigns` where we clear the whole
             // result cache. It's currently the only user-invoked or single item place where
             // we do so.
-            /**
-             * @psalm-suppress DeprecatedMethod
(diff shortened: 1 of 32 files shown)


src/Domain/DonationService.php
```diff
@@ -183,12 +183,10 @@ public function buildFromAPIRequest(DonationCreate $donationData, PersonId $dono
...
-            /**
-             * @psalm-suppress DeprecatedMethod
-             * @var CacheProvider $cacheDriver
-             */
-            $cacheDriver = $this->entityManager->getConfiguration()->getResultCacheImpl();
-            $cacheDriver->deleteAll();
+            $resultCache = $this->entityManager->getConfiguration()->getResultCache();
+            if ($resultCache) {
+                $resultCache->clear();
+            }
```
bdsl, line 189:
```suggestion
            $this->entityManager->getConfiguration()->getResultCache()?->clear();
```

src/Domain/FundRepository.php
```diff
@@ -274,5 +281,8 @@ protected function doUpdateFromSf(SalesforceReadProxy $proxy, bool $withCache):
 
         $fundData = $this->getClient()->getById($fundId, $withCache);
         $proxy->setName($fundData['name'] ?? '');
+        // Not sure I'm entirely convinced anything but tests is using this for Funds, but mirroring what we do with
+        // 'name' for now just in case. Maybe we can make the fn throw instead? Needs more attention later ideally.
+        $proxy->setSlug($fundData['slug'] ?? '');
```
bdsl, line 286: yes it doesn't look like any prod code calls `\MatchBot\Domain\SalesforceReadProxyRepository::updateFromSf` except where it knows that that the `SalesforceReadProxyRepository` is actually a `CampaignRepository`. Might be worth just making this function generate an error if called since it'd be alarming to be wrong about that. And then after that we could make it throw or better just move updateFromSf down from `SalesforceReadProxyRepository` to `CampaignRepository` and delete this function.

bdsl: We use `\MatchBot\Domain\FundRepository::pullForCampaign` instead to update funds.

bdsl (dismissed):
LGTM, added a couple of suggestions but not blockers.

bdsl (approved):

PR #1513, opened 2025-07-16
Phase out `x_` prefix for isMetaCampaign

NoelLH:
Should be only in the list of ignored fields (where I think it needs to stay until SF switches over) after this


bdsl (approved):
LGTM, I think since yesterday sf in prod is sending this with and without the `x_` so makes sense to switch to using the without version now.

PR #1514, opened 2025-07-17
BG2-2919: Don't generate alarm when failing to load a campaign preview

bdsl:
This may have been a long standing issue that some campaign previews are not available in SF when people expect them, but now generates more noise since matchbot is involved.


src/Application/Actions/Campaigns/GetEarlyPreview.php
```diff
@@ -9,7 +9,6 @@
 use MatchBot\Application\Actions\Action;
 use MatchBot\Application\Environment;
 use MatchBot\Client\Campaign as SfCampaignClient;
-use MatchBot\Client\NotFoundException;
```
bdsl, line 12: deleted since it was not in use. I do now use this class but seemed clearer to give the fully qualified name when its used.

PR #1517, opened 2025-07-18
MAT-441 – slightly delay pushing active Mandates to Salesforce

NoelLH:
To reduce the likelihood of record contention issues on the receiving side, similarly to what we do with Donations


bdsl (approved):
LGTM. I wonder if we should maybe not push mandates at all that have never been created, they are probably worth very little.

NoelLH:
> LGTM. I wonder if we should maybe not push mandates at all that have never been created, they are probably worth very little.

Maybe, I think the fewer exceptions there are to when we push the less confusing it is, but if we see continuing problems that could be another option in addition to those I put on Jira.

PR #1519, opened 2025-07-18
DON-1127 – enable Regular Giving in Production

NoelLH:


bdsl (approved):
LGTM - there are a couple of todo-regular giving comments in matchbot but IMHO they shouldn't be blockers, and in any case turning this feature flag on doesn't really do anything that matters until there is an open regular giving campaign.

PR #1521, opened 2025-07-18
DON-1120: WIP adding fields for metacampaign colours to metacampaign

bdsl:


bdsl:
Closing for now, I think as we have relatively few metacampaigns having them updatable in SF should be low priority - I'd rather keep this hard coded in PHP for now and then when we know we've got something we're happy with we can release it for real use first, and then later do this work to stop having to do manual code edits.

PR #1522, opened 2025-07-22
MAT-441 – fix & expand Mandate-pushing queue delay stamps

NoelLH:
* Fix comparison to our Active status
* Add a delay for Cancelled status too, to fix a similar but distinct record contention issue


bdsl (approved):
LGTM

PR #1525, opened 2025-07-24
MAT-443 – fix saved card donations, including Regular Giving, not setting application fees

NoelLH:


src/Domain/Donation.php
```diff
@@ -1619,6 +1624,7 @@ public function assertIsReadyToConfirm(\DateTimeImmutable $at): true
         $this->assertionsForConfirmOrPreAuth()
             ->that($this->transactionId)->notNull('Missing Transaction ID')
             ->that($this->getCampaign()->isOpenForFinalising($at))
+            ->that($this->hasCharityFee())
```
bdsl, line 1627: I think this is good but maybe it's worth changing the default for charity fee from `'0.00'` to `null`, and then this would just assert that it's not null.

That way if we want to introduce zero fee donations in future this won't get in the way, and more importantly we're not relying quite so much on this check, if we forget to set the fee again then we'd get an error either at runtime or statically. Except that lots of things including bcadd cooerce null to zero so we'd have to wrap the non-null value in a money object to stop that happening.

I'm also wondering again if it's possible to make `\MatchBot\Domain\Donation::deriveFees` private and have the donation class call it whenever necessary. Might mean we'd have to save the card brand and country on the donation so that the information is there to enable recalculating the fee if/when other values change (really just `hasGiftAid` since amount etc is immutable)

src/Domain/DonationService.php
```diff
@@ -710,14 +719,21 @@ public function findAllCompleteForCustomerAsAPIModels(StripeCustomerId $stripeCu
     /**
      * @throws PaymentIntentNotSucceeded
      * */
-    public function confirmDonationWithSavedPaymentMethod(Donation $donation, StripePaymentMethodId $paymentMethod): void
+    public function confirmDonationWithSavedPaymentMethod(Donation $donation, StripePaymentMethodId $paymentMethodId): void
     {
         $paymentIntentId = $donation->getTransactionId();
+
+        $this->updateDonationFeesFromPaymentMethodId($donation, $paymentMethodId);
```
bdsl, line 726: maybe better to make this line something like `$donation->setPaymentMethod($paymentMethod)`, and have the implementation of that set the fee as a side effect? Maybe also move that into the two callers of this method (`\MatchBot\Domain\RegularGivingService::setupNewMandate` and `\MatchBot\Domain\DonationService::confirmPreAuthorized`)

And then we'd remove the $paymentMethodId param from this method, and either call a new function `$donation->getPaymentMethod()` (or `$donation->getPaymentMethodId()` ) or possibly `$donation->getStripeConfirmationPayload()`. The latter of course would need to account for the more common case of having a confirmation token instead of a payment method.

src/Domain/DonationService.php
```diff
@@ -969,4 +985,24 @@ public function doCreateDonation(string $pspCustomerId, DonationCreate $donation
 
         return $donation;
     }
+
+    private function updateDonationFeesFromPaymentMethodId(Donation $donation, StripePaymentMethodId $paymentMethodId): void
```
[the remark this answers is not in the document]
NoelLH: It does feel like relatively common post-campaign corrections, around charity Gift Aid claim enrolment and donor Gift Aid eligibility, might make setters that always want to change fees problematic. But perhaps there aren't many cases that use the ORM over migrations in practice.

NoelLH:
Barney and Noel merging as pair

PR #1536, opened 2025-07-30
Make `UpdateCampaignDonationStats` ignore old campaigns with known discrepancies

NoelLH:


bdsl (approved):
LGTM, I think any statistics we generate for very old closed campaigns are not really going to matter. 

Although in principle it might be better to go ahead and generate the statistics anyway and change the assertion at https://github.com/thebiggive/matchbot/blob/7b87a74294927b3cbd72bff33e4fd70395f83537/src/Domain/CampaignStatistics.php#L158 so that we don't assert its true in 100% of cases if its not (and don't throw an exception if its not) but instead just log an error if its false and we're not looking at one of the campaigns with known overmatch.

But probably not worth doing that while we don't have a use case for the stats on those very old campaigns.

PR #1542, opened 2025-08-01
Support regularGiving campaign family

NoelLH:


bdsl (approved):
LGTM. Not sure why I didn't include this in the first place, looks like I must have copied the list of families from an outdated copy of SF code or something.

PR #1553, opened 2025-08-07
MAT-445: Stop non-match-funded campaigns from show up in any search r…

bdsl:
…esults}

Before:

image

After:

image


NoelLH (commented):
LGTM

PR #1555, opened 2025-08-07
MAT-445: Exclude campaigns that ended in the past from explore page

bdsl:


NoelLH (commented):
@bdsl Is the thinking that historic SF data isn't reliable enough and that it's simpler to work around that here? AFAIK we are expecting new campaigns to get status changes on their SF fields within a minute or two.

bdsl:
> @bdsl Is the thinking that historic SF data isn't reliable enough and that it's simpler to work around that here? AFAIK we are expecting new campaigns to get status changes on their SF fields within a minute or two.

Assuming this is about the "we can't rely on the status being updated when "  . I guess I was just thinking that the version of status that takes into account the time was a formula field so wouldn't be sent as an update.

Sounds like its probably not accurate to say we can't rely on the status being updated, but still IMHO it would be nice not to have to rely on that. Given that we've had to put a lot of time into understanding how statuses get updated in SF and the semantics of the various status fields it seems like it would be better to just rely on data that's saved directly on user input.

And IMHO worth having the campaigns disappear here exactly at the moment the campaign ends instead of a minute or two later.

NoelLH:
> > @bdsl Is the thinking that historic SF data isn't reliable enough and that it's simpler to work around that here? AFAIK we are expecting new campaigns to get status changes on their SF fields within a minute or two.
> 
> Assuming this is about the "we can't rely on the status being updated when " . I guess I was just thinking that the version of status that takes into account the time was a formula field so wouldn't be sent as an update.
> 
> Sounds like its probably not accurate to say we can't rely on the status being updated, but still IMHO it would be nice not to have to rely on that. Given that we've had to put a lot of time into understanding how statuses get updated in SF and the semantics of the various status fields it seems like it would be better to just rely on data that's saved directly on user input.
> 
> And IMHO worth having the campaigns disappear here exactly at the moment the campaign ends instead of a minute or two later.

I think it's a time-based Flow trigger which updates `Campaign_Status__c`. That field has `true` in the map which says if things are relevant to MB.

And with this PR there is still a 1 minute offset, so I'm not really clear what we gain functionally for the complexity.

bdsl:
> I think it's a time-based Flow trigger which updates Campaign_Status__c. That field has true in the map which says if things are relevant to MB.
> 
> And with this PR there is still a 1 minute offset, so I'm not really clear what we gain functionally for the complexity.

Right, probably not very much then. Maybe I was thinking there was redundancy between the status and the enddate fields and trying to reduce that by using the enddate field instead, which I think is also closer to the behavour in frontend.

Let me check in prod what campaigns if any have now that are not closed after the end date.

PR #1565, opened 2025-08-13
Don 984 sitemap

bdsl:
image


src/Application/Actions/Sitemap.php
```diff
@@ -0,0 +1,116 @@
...
+    protected function action(Request $request, Response $response, array $args): Response
+    {
+        // initial implementation of sitemap - following spec from https://www.sitemaps.org/protocol.html#prioritydef
+        // which is what Google uses.
+
+        // For now just listing up to 100k campaign and (if open) donate pages.
+
+        // @todo:
+        // - add charity pages as well as campaigns.
+        // - add metacampaigns
```
NoelLH, line 44: To my mind this is probably the highest priority/ maybe the only essential item on this list.

src/Application/Actions/Sitemap.php
```diff
@@ -0,0 +1,116 @@
...
+        // - add some of the most important other pages, e.g. static content. But afaik its not essential for
+        //   a sitemap to be comprehensive to be useful
+        // - consider if the 100k limit is appropriate.
+        // - increase cache duration, maybe explicitly expire cache when content changes. Consider persisting
+        // - something that would be cheaper to access than this to make additions to when we add a campaign.
+        // - add integration test
+        // - submit sitemap to search engines and/or list in robots.txt
+        $sitemapXml = $this->cache->get('sitemap', function (ItemInterface $cacheItem) {
+            $cacheItem->expiresAfter(60 * 10); // ten minutes
```
NoelLH, line 53: Makes sense for now. We could perhaps go shorter or reconsider whether the app itself needs any cache for this once it's behind CloudFront.

src/Application/Actions/Sitemap.php
```diff
@@ -0,0 +1,116 @@
...
+
+            $xml = new SimpleXMLElement('<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"></urlset>');
+            $campaigns = $this->campaignRepository->search(
+                sortField: 'amountRaised',
+                sortDirection:  'asc',
```
NoelLH, line 58: If we thought we might hit the 100k I'd suggest prioritising campaigns with later dates and/or setting an earliest date cutoff. But probably not going to make a real difference to the results for the foreseeable.

NoelLH (approved):
LGTM

PR #1567, opened 2025-08-14
MAT-444 Tell clients to cache responses from appropriate routes

NoelLH:


src/Application/CacheableResponseMiddleware.php
```diff
@@ -0,0 +1,26 @@
...
+    private const int CACHE_SECONDS = 60; // 1 minute
+
+    #[\Override]
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $response = $handler->handle($request);
+
+        return $response
+            ->withHeader('Cache-Control', 'public, max-age=' . self::CACHE_SECONDS)
+            ->withHeader('Expires', gmdate('D, d M Y H:i:s', time() + self::CACHE_SECONDS) . ' GMT');
```
NoelLH, line 24: Should be redundant with the above `Cache-Control` according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Expires but I think it's fairly common to do both, perhaps because for very old HTTP/1.0 clients `Expires` could be the only effective one? https://stackoverflow.com/questions/5799906/what-s-the-difference-between-expires-and-cache-control-headers

But the only client whose caching is very important here is CloudFront itself. Happy to simplify if we think it's better to only keep `Cache-Control`.

bdsl: I don't think it matters having both.

bdsl (approved):
LGTM but what would you think about refactoring it like below, i.e. having the middleware class return a closure. The advantage would be that then we could have each route cached for a different amount of time, e.g. early previews for one or two seconds, and we can see the duration directly in the routes file.

But maybe that's adding complexity for no real reason. Although for the sitemap I think we could cache for a lot longer, I've done 30 minutes for now but it probably could be cached for an hour or so.

Image

If you think this would be useful I can push it into this PR branch or develop after this is merged.

bdsl:
> LGTM but what would you think about refactoring it like below

Disadvantage of that is that there's nothing to stop us forgetting we have to pass the closure and just passing the class-string as normal, since the docblock for `\Slim\Interfaces\RouteInterface::add` says it accepts `string` among other things.

PR #1574, opened 2025-08-15
MAT-429 – support Stripe Pay By Bank

NoelLH:


bdsl (approved):
LGTM

src/Domain/DonationService.php
```diff
@@ -456,21 +455,31 @@ private function updateDonationFeesFromConfirmationToken(
...
+                $cardCountry,
+            ));
 
-        $donation->setPaymentCard(new PaymentCard($cardBrand, $cardCountry));
+            $donation->setPaymentCard(new PaymentCard($cardBrand, $cardCountry));
+        } else {
+            // if we had a ctoken at all we're not using Donation Funds so must be using Pay By Bank, which
+            // has no card / default fees.
+            \assert($paymentMethodPreview->pay_by_bank !== null);
+            $donation->setPaymentCard(null);
```
bdsl, line 481: I think this was my mistake a while ago but might be worth updating the docblock on setPaymentCard to clarify that it does now internally update the fees.

bdsl (dismissed):
LGTM

src/Domain/DonationService.php
NoelLH: Thanks, good spot. Made that hopefully clearer in a1a6dba541d188678dd69a0fbf715d32cd8c72a8

PR #1592, opened 2025-08-22
Remove PHPStan baseline, fix all reported static analysis issues

bdsl:


src/Application/Actions/Donations/Update.php
```diff
@@ -164,7 +164,7 @@ protected function action(Request $request, Response $response, array $args): Re
                     $this->entityManager->rollback();
 
                     // Log a warning to more easily spot occurrences in dashboards.
-                    $methodSummary = $donation->getPaymentMethodType()?->value ?? '[null]';
+                    $methodSummary = $donation->getPaymentMethodType()->value ?? '[null]';
```
bdsl, line 167: (no need for null-safe navigation to the left of `??`, as that does the same thing)

PR #1595, opened 2025-08-26
Refactor: Simplify SalesforceReadProxyRepository by moving updateFrom…

bdsl:
…Sf down to CampaignRepository, delete unused code

This leaves SalesforceReadProxyRepository as an empty class which potentially could also be deleted.


NoelLH:
Probably don't want to review or merge this while there are more urgent changes on the same classes pending review

NoelLH (approved):

PR #1611, opened 2025-09-03
Fix regression not actually cancelling stale mandates

NoelLH:
Should cancel donations too, aligning the logic with other cancellation scenarios


bdsl:
LGTM

PR #1613, opened 2025-09-05
DON-1155 Regular Giving: fix timezone bug determining 2nd date; tell Stripe when we're off session

NoelLH:


src/Domain/DonationService.php
```diff
@@ -711,8 +711,11 @@ public function findAllCompleteForCustomerAsAPIModels(StripeCustomerId $stripeCu
     /**
      * @throws PaymentIntentNotSucceeded
      * */
-    public function confirmDonationWithSavedPaymentMethod(Donation $donation, StripePaymentMethodId $paymentMethodId): void
-    {
+    public function confirmDonationWithSavedPaymentMethod(
+        Donation $donation,
+        StripePaymentMethodId $paymentMethodId,
+        bool $offSession = false,
```
bdsl, line 717: not a blocker but maybe worth removing the default value here so we can't forget to set it if we add any more calls to this function. There are only a few calls to it at the moment.

src/Domain/DonationService.php
bdsl, line 1: 👍

bdsl (commented):
Added one comment inline but overall LGTM.

src/Domain/DonationService.php
NoelLH: Done in 6a67ac1ce30c6d3dd543ff086341e5d03f202c36

PR #1620, opened 2025-09-09
Fix supression of campaign that we don't have enough data to render

bdsl:
Clearly we tried to do this some time ago but forgot that at the point we test the array it will always have a copy of the charity's data in it.


bdsl:
> Possibly worth giving `Campaign` itself a helper method to DRY up repetition vs. `Search::action()`? Not a blocker.

Good call, done in cf8c563a98dabdc967d665571f5d8d8022c290f2

PR #1639, opened 2025-09-24
MAT-449: Add Rate limit donating with new cards per user

bdsl:
Thinking we will call this to make sure users aren't testing different cards at too high a rate.

### DIFF
--- a/app/dependencies.php
+++ b/app/dependencies.php
@@ -334,8 +334,6 @@ return function (ContainerBuilder $containerBuilder) {
                     'policy' => 'token_bucket',
                     // how many distinct card confirm attempts are allowed in the time window
                     'limit' => 5,
-
-                    // how often they can create new donations once the initial allowance is used up.
                     'rate' => ['interval' => '30 minutes'],
                 ],
                 storage: $c->get(RateLimiterStorage::class)


src/Domain/DonationService.php
```diff
@@ -455,12 +458,41 @@ private function updateDonationFeesFromConfirmationToken(
...
+                $stripeCustomerId = $donation->getPspCustomerId()?->stripeCustomerId;
+                \assert(\is_string($stripeCustomerId));
+
+                $rateLimitId = 'confirm-donation-' . $stripeCustomerId;
+                $rateLImiter = (new RateLimiterFactory([
+                    'id' => $rateLimitId,
+                    'policy' => 'token_bucket',
+                    'limit' => 5,
+                    'rate' => ['interval' => '30 minutes'],
+                ], $this->rateLimiterstorage))->create();
```
NoelLH, line 480: I'm not sure I see why we need to work with the underlying Redis directly.

Could we pass `key` to `create()` here and indeed inject a (differently aliased) `RateLimiterFactory`, so that the unchanging config lives in dependencies and everything works more like `DonationService::doCreateDonation()`?

NoelLH (commented):
I think we could maybe make the way we use the rate limiter more similar to what we do for the separate `$pspCustomerId` token bucket, and that might make things easier to follow and avoid needing custom Redis key management.

We'll also want tests as discussed. Otherwise LGTM so far.

src/Domain/DonationService.php
bdsl: ah yes we could - I didn't see that `create` takes an argument. Yes I agree in that case it's better to put the rate limiter in the dependencies and inject it. Will do.

Also hadn't realised on Wednesday that we are already using the Symfony Rate Limiter.

But do you think using redis directly to check whether the card is a new one or not makes sense, since we only want to limit use of new cards, not retries of the same card they've tried before?

NoelLH: Ah yes, I think I was temporarily confused about the card fingerprints when I first reviewed. That probably does make sense since it's sort of the opposite of how rate limiting strategies normally work (more tries are allowed if in the _same_ value bucket) so is unlikely to be covered out the box. And then the rate limiting bucket itself is based on customer ID.

NoelLH (dismissed):
Apart from rate limit dependency docs needing an update, all LGTM now

PR #1663, opened 2025-10-08
MAT-460 – patch WGMF25 funding

NoelLH:


src/Application/Commands/PatchFunds8Oct.php
```diff
@@ -0,0 +1,129 @@
...
+
+        return Command::SUCCESS;
+    }
+
+    /**
+     * @return Campaign[]
+     */
+    private function getCampaignsAffected(): array
+    {
+        // Raw pMA export
```
bdsl, line 70: may be useful for ease of double checking to add to this comment what query was run in pMA ?

src/Application/Commands/PatchFunds8Oct.php
```diff
@@ -0,0 +1,129 @@
...
+
+        return $campaigns;
+    }
+
+    private function switchMatchingToCorrectFunding(Donation $donation): void
+    {
+        $originallyMatched = $donation->getFundingWithdrawalTotal();
+        /** @psalm-suppress InternalMethod */
+        $this->allocator->releaseMatchFunds($donation);
+        $nowMatched = $this->allocator->allocateMatchFunds(donation: $donation, forceNotBigGive: true);
```
bdsl, line 114: should these two be wrapped in a transaction?

src/Domain/DoctrineDonationRepository.php
```diff
@@ -85,6 +85,38 @@ public function findWithExpiredMatching(\DateTimeImmutable $now): array
         return array_map(static fn(array $row): UuidInterface => $row['uuid'], $rows);
     }
 
+    /**
+     * @return Donation[]
+     */
+    public function findWithBigGiveWgmf25Matching(Campaign $campaign): array
```
bdsl, line 91: the implementation here doesn't seem to have anything specific to WGMF 25, so slightly confusing name, but not worth checking since it looks like this is code to run once and then delete.

src/Application/Commands/PatchFunds8Oct.php
bdsl: Assuming this of from the "better query" from Jira. No need to add the comment to the code, only important that we're clear between us in the short term.

src/Application/Commands/PatchFunds8Oct.php
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    #[\Override]
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $campaigns = $this->getCampaignsAffected();
+        $output->writeln('Campaigns affected: ' . count($campaigns));
+
+        foreach ($campaigns as $campaign) {
+            $campaignId = $campaign->getId() ?? 0;
```
bdsl, line 49: the 0 isn't meaningful is it?
```suggestion
            $campaignId = $campaign->getId() ?? throw new \Exception('campaign id missing');
```

bdsl (approved):
Added some small comments inline, but overall LGTM.

src/Application/Commands/PatchFunds8Oct.php
NoelLH: I think because Redis is the source of truth there isn't a very easy way to do that. Ideally yes, but as it's small fractions of a second I think the risk is low compared to the increased risk of fund over allocation by delaying.

PR #1914, opened 2026-04-02
MAT-471 Split up & speed up retrospective matching after a campaign

NoelLH:


src/Application/Commands/RetrospectivelyMatch.php
```diff
@@ -72,81 +56,35 @@ protected function configure(): void
     protected function doExecute(InputInterface $input, OutputInterface $output): int
     {
         $recentlyClosedMode = !is_numeric($input->getArgument('days-back'));
-        $oneHourBeforeExecStarted = (new DateTime('now'))->sub(new \DateInterval('PT1H'));
+        // Default is 1 hour before exec started.
+        $sinceDate = (new \DateTimeImmutable('now')->sub(new \DateInterval('PT1H')));
```
bdsl, line 60: not a big issue since this is near the top level of the app but might be worth injecting a clock or timestamp for testability.

bdsl: also I see the only change on this line is adding the immutability so my comment isn't strictly about the change.

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
```diff
@@ -0,0 +1,163 @@
...
+        $numWithMatchingAllocated = 0;
+        /** @var int[] Fine to include duplicates; use of Redis set in `addToStats()` will efficiently discard them. */
+        $campaignIds = [];
+        $totalNewMatching = '0.00';
+
+        $donationRepository = $this->container->get(DonationRepository::class);
+        $donations = $donationRepository->findByUuids($message->donationUuids);
+        foreach ($donations as $donation) {
+            $amountAllocated = $this->allocator->allocateMatchFunds($donation);
+
```
bdsl, line 68: There are only a few uses of bc functions here so definitely not a blocker but either here or next time we're using bc functions might be worth considering using the new [BcMath\Number](https://www.php.net/manual/en/class.bcmath-number.php) class instead of strings for better DX.

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
```diff
@@ -0,0 +1,163 @@
...
+            if (bccomp($amountAllocated, '0.00', 2) === 1) {
+                $this->entityManager->flush();
+                $this->bus->dispatch(DonationUpserted::fromDonationEnveloped($donation));
+                $numWithMatchingAllocated++;
+                $totalNewMatching = bcadd($totalNewMatching, $amountAllocated, 2);
+                $campaignIds[] = $donation->getCampaign()->getId();
+            }
+        }
+
+        // todo decide on keeping bcmath for this or just using int pence.
```
bdsl: or as suggested above either continue using bcmath but switch to its new Number class instead of string based functions, or fourthly switch to using our custom Money class.

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
```diff
@@ -0,0 +1,163 @@
...
+            $this->redis->sAdd($this->keyForStat('campaignIdsWithChanges', $message), $campaignId);
+        }
+
+        // todo possibly set a 1 week expire on all stats as belt and braces
+    }
+
+    private function reportStats(DonationMatchingShouldBeChecked $message): void
+    {
+        /** @var int $numChecked */
+        $numChecked = $this->redis->get($this->keyForStat('numChecked', $message));
```
bdsl: probably worth replacing these `@var` comments with `assert` or `Assertion`

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
```diff
@@ -0,0 +1,163 @@
...
+        /** @var int $numWithMatchingAllocated */
+        $numWithMatchingAllocated = $this->redis->get($this->keyForStat('numAllocated', $message));
+        /** @var numeric-string $totalNewMatching */
+        $totalNewMatching = bcdiv($this->redis->get($this->keyForStat('penceAllocated', $message)), '100', 2); // todo see above
+        /** @var int $numDistinctCampaigns */
+        $numDistinctCampaigns = $this->redis->scard($this->keyForStat('campaignIdsWithChanges', $message));
+
+        $summary = "Retrospectively matched $numWithMatchingAllocated of $numChecked donations. " .
+            "£$totalNewMatching total new matching, across $numDistinctCampaigns campaigns.";
```
bdsl, line 129: Should we use the currency from the campaign instead of hard-coding '£' here if we still want to keep the ability to potentially run campaigns in other currencies? But then maybe that would mean we'd need to either return a bag of multiple monies instead of just one or check that all donations we've dealt with have the same currency.

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
```diff
@@ -0,0 +1,163 @@
...
+        $this->logger->info($summary);
+
+        // If we did any new matching allocation, whether because of campaigns just closed or because
+        // the command was run manually, send the results to Slack.
+        if ($numDistinctCampaigns > 0 && $this->environment !== Environment::Test) {
```
bdsl, line 134: not a blocker but as mentioned it doesn't seem like this function should have to care whether we're in a test or not - IMHO better to have something dependencies.php or phpunit's bootstrap to set a no-op or in-memory Chatter transport in the container for integration tests.

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
```diff
@@ -0,0 +1,163 @@
...
+        [$numberChecked, $donationsAmended] = $this->matchFundsRedistributor->redistributeMatchFunds();
+        $this->logger->info("Checked $numberChecked donations and redistributed matching for $donationsAmended");
+    }
+
+    private function syncFundUsage(DonationMatchingShouldBeChecked $message): void
+    {
+        $fundRepository = $this->container->get(FundRepository::class);
+        // We want to include funds related to all campaigns in the scope of the retro match.
+        $funds = $fundRepository->findForCampaignsClosedSince(
+            closedBeforeDate: new \DateTimeImmutable('now'),
```
bdsl, line 159: as above re injecting timestamp / clock - not a blocker near the top of the application but I think would be a bit clearer if we inject the now timestamp.

src/Application/Messenger/DonationMatchingShouldBeChecked.php
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Message to tell workers to look for & allocate further match funds if possible.
+ * $retroMatchJobUuid is generated by the command that queued all donations, typically
+ * one just after the end of each meta-campaign.
+ */
+class DonationMatchingShouldBeChecked implements MessageDeduplicationAwareInterface, MessageGroupAwareInterface
+{
+    /**
+     * @param array<string> $donationUuids
```
bdsl: probably can make this a bit more specific

```suggestion
     * @param list<string> $donationUuids
```

src/Domain/DonationRepository.php
```diff
@@ -41,14 +41,20 @@ public function findWithMatchingWhichCouldBeReplacedWithHigherPriorityAllocation
     public function findReadyToClaimGiftAid(bool $withResends): array;
 
     /**
+     * @param string[] $uuids
      * @return Donation[]
      */
-    public function findNotFullyMatchedToCampaignsWhichClosedSince(DateTime $closedSinceDate): array;
+    public function findByUuids(array $uuids): array;
```
bdsl, line 47: maybe worth making this take a `list<UuidInterface>` instead of an array of string? Then the one call would look like:

```
        $donations = $donationRepository->findByUuids(\array_map(Uuid::fromString(...), $message->donationUuids));
```

And we might not need the assertion in the message constructor that the Uuid strings are valid Uuids since we'd get the exception in the call to `fromString`. I think the seralisizable message still might want to have the UUIDs as strings, or at least not as the interface but as a specific class that we know is intended to be serialized.

bdsl:
Added some comments inline, other than that and where @NoelLH you've left todo comments already all LGTM.

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
NoelLH: I'll add a todo, but since this wouldn't critically break anything important even if we did forget it (the adapter is currency agnostic) I don't think it's worth spending time on now.

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
NoelLH: Right, I think a refactor that does this cleanly might possibly go well alongside what I mentioned back when we added Slack this way, re. hooking into Monolog. I think a conditionally invoked handler might also let us do less reinventing the wheel and avoid repetition between the logger lines and things like this.

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
NoelLH: Yep, also not new code and not stopping any testing as yet so will leave for now.

src/Domain/DonationRepository.php
NoelLH: Started trying to do this but I'm not sure.

If a UUID is invalid I think it's better for something to throw sooner (message instantiation before publishing) rather than only after a message that can never work is being consumed.

And if we want to keep the messages small and compatible by passing the UUIDs as strings, it seems like quite a big cost to both convert to UUID types for instantiation and then convert back, when we already have an assertion. I think there's quite a bit of value in a message model being simple and having constructor params that need minimal processing to get to the payload for transport.

bdsl (dismissed):

bdsl:
LGTM

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
NoelLH: Partly done in 2135081ad4a628c6a8a2559c49d770a3c3d9b21e but it turns out int is too strict because Redis GET does numeric strings, so I had to use `is_numeric()` on everything but the set count.

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
NoelLH: Moved things over in 2135081ad4a628c6a8a2559c49d770a3c3d9b21e while keeping pence as the Redis store in order to make that simple and safe for lots of threads to work on.

I think apart from the [one place](https://github.com/thebiggive/matchbot/pull/1914/changes#diff-a0bfea61eb918fdd5e339a07d376a221e8a7de9569b1980cf5702e5739b814cfR82) where this leads to `(int) (string) ...` it pretty consistently makes DX a bit better as you say. Not ideal but I think the best compromise for now.

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
```diff
@@ -0,0 +1,185 @@
...
+        $campaignIds = [];
+        $totalNewMatching = new Number('0.00');
+
+        $donationRepository = $this->container->get(DonationRepository::class);
+        $donations = $donationRepository->findByUuids($message->donationUuids);
+        foreach ($donations as $donation) {
+            $amountAllocatedNumericString = $this->allocator->allocateMatchFunds($donation);
+            $amountAllocated = new Number($amountAllocatedNumericString);
+
+            if ($amountAllocated > new Number('0.00')) {
```
bdsl, line 69: 👍 +1:  I think this is quite a bit easier to read than the string version. I'm pretty sure it would also work as:


```suggestion
            if ($amountAllocated > new Number(0)) {
```

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
```diff
@@ -56,26 +56,31 @@ public function __invoke(DonationMatchingShouldBeChecked $message): void
...
+            $amountAllocatedNumericString = $this->allocator->allocateMatchFunds($donation);
+            $amountAllocated = new Number($amountAllocatedNumericString);
 
-            if (bccomp($amountAllocated, '0.00', 2) === 1) {
+            if ($amountAllocated > new Number('0.00')) {
                 $this->entityManager->flush();
                 $this->bus->dispatch(DonationUpserted::fromDonationEnveloped($donation));
                 $numWithMatchingAllocated++;
-                $totalNewMatching = bcadd($totalNewMatching, $amountAllocated, 2);
+                $totalNewMatching = $totalNewMatching->add(num: $amountAllocated, scale: 2);
```
bdsl, line 73: IMHO we probably don't need to specify the scale for using functions like these, it automatically copies the scale from whichever operand has the greater scale. Different to `bcadd` which uses a global default value.

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
```diff
@@ -56,26 +56,31 @@ public function __invoke(DonationMatchingShouldBeChecked $message): void
...
             }
         }
 
-        // todo decide on keeping bcmath for this or just using int pence.
-        $this->addToStats($message, $campaignIds, $numWithMatchingAllocated, (int) bcmul($totalNewMatching, '100', 0));
+        $this->addToStats(
+            message: $message,
+            campaignIdsWithAllocations: $campaignIds,
+            numberAllocated: $numWithMatchingAllocated,
+            penceAllocated: (int) (string) $totalNewMatching->mul(num: new Number(100), scale: 2),
```
bdsl: will take a look on my machine, not sure if we should need to go via string? Maybe once we've multiplied by 100 we want the scale to be 0 instead of 2?

bdsl: hmm I see phpstan doesn't allow directly casting from `Number` to `int`, even with scale 0 and `Number` doesn't have any method or property to get the int value directly. So I see why going via string. Maybe this is marginally nicer but doesn't make much difference:

```suggestion
            penceAllocated: (int) $totalNewMatching->mul(num: new Number(100), scale: 0)->value,
```

bdsl (dismissed):
LGTM

src/Application/Messenger/Handler/DonationMatchCheckHandler.php
NoelLH: Yep that's how I got there. Might be because the classes are new to me but I think cast plus `->value` would take me longer to grok, will keep it as is for the moment (except scale)

PR #1915, opened 2026-04-02
BG2-3106: Withhold platoform fee (tip + fee) from charities when taki…

bdsl:
…ng payment via Ryft


bdsl:
superseded by https://github.com/thebiggive/matchbot/pull/1920

PR #1919, opened 2026-04-08
Tidy up unused message handler & old @todo

NoelLH:


src/Domain/DonationService.php
```diff
@@ -786,10 +786,6 @@ public function confirmDonationWithSavedPaymentMethod(
         $this->logger->info("PaymentIntent: {$paymentIntent->toJSON()}");
 
         if ($paymentIntent->status !== PaymentIntent::STATUS_SUCCEEDED) {
-            // @todo-regular-giving-mat-407: create a new db field on Donation - e.g. payment_attempt_count and update here
```
NoelLH, line 851: I think we covered this adequately, and thus closed the referenced ticket, using a date-based limit on retries after which mandates are already auto cancelled.

bdsl (approved):
LGTM

PR #1920, opened 2026-04-08
Don 1192 ryft

bdsl:


src/Application/Actions/Donations/Confirm.php
```diff
@@ -106,6 +123,49 @@ protected function action(Request $request, Response $response, array $args): Re
             throw new HttpBadRequestException($request, "stripeConfirmationTokenId required");
         }
 
+        if ($psp === PaymentServiceProvider::Ryft) {
+            Assertion::integer($paymentAmount);
+
+            // @todo - work out why assertion below is failling in manual tests - getting e.g.
+            // { type: "SERVER_ERROR", description: 'Value "466" is not the same as expected value "575".' }
+            // Assertion::same($paymentAmount, $donation->getAmountFractionalIncTip());
+
```
NoelLH: Is this now superseded by the amount from `$capture` below? Can we replace the assertion with one on the captured amount, maybe in `collectFromRyftPaymentSession()`?

src/Client/RyftClient.php
```diff
@@ -79,4 +74,86 @@ public function createPaymentSession(RyftAccountId $ryftAccountId, Money $amount
...
+     * @param array{id: string, ...} $paymentSession
+     * @return array{id: string, amount: int, platformFee: int, currency: string}
+     */
+    public function capturePayment(RyftAccountId $ryftAccountId, array $paymentSession, Money $platformFee): array
+    {
+        // https://api.ryftpay.com/v1/payment-sessions/{paymentSessionId}/captures
+
+        $request = new Request(
+            method: 'POST',
+            uri: $this->apiPrefix . 'payment-sessions/' . $paymentSession['id'] . '/captures/',
```
NoelLH:
```suggestion
            uri: $this->apiPrefix . 'payment-sessions/' . $paymentSession['id'] . '/captures',
```

src/Client/RyftClient.php
```diff
@@ -79,4 +74,86 @@ public function createPaymentSession(RyftAccountId $ryftAccountId, Money $amount
...
+
+        $response = $this->client->send($request);
+        $responseContents = $response->getBody()->getContents();
+        $responseData = json_decode($responseContents, true, \JSON_THROW_ON_ERROR);
+
+        $this->log->info(\var_export(\compact('responseData'), true));
+
+        $status = $responseData['status']; // @phpstan-ignore-line
+        if ($status !== 'Succeeded') {
+            throw new \Exception('Couldn ot capture Ryft payment');
```
NoelLH, line 141:
```suggestion
            throw new \Exception('Could not capture Ryft payment');
```

NoelLH (dismissed):
A question and tiny changes inline, no blockers, LGTM!

src/Application/Actions/Donations/Confirm.php
bdsl: I think it's worth asserting that the amount is right before we attempt to capture, but we can assert it based on what we've seen it to be (the part the charity will receive) instead of leaving the commented out assertion.

Done in 9d03a0946

src/Client/RyftClient.php
bdsl: Thanks, that fits the API docs a bit better, done in 73b63f6ef . Ideally the docs would give an environment-agnostic version of the URI, but they don't quite.

PR #1921, opened 2026-04-10
DON-1192: Apply same donation check / lock logic to ryft as stripe do…

bdsl:
Moving ryft-specific logic out of Confirm handler and putting it in Donation service, closer to where we do the equivalent with Stripe so that we can more easily do the same thing for both types of donation.

This also naturally means the same things like locking and unlocking the donation before and after capturing the funds happen with both services.


NoelLH (approved):
Skipped detail where it looked like a bulk move. Principle and what I did read LGTM

PR #1924, opened 2026-04-15
Revert "BG2-3129 Temporarily call `updateFromSf()` again for any Campaigns eligible for up-to-date fund info"

NoelLH:
Reverts thebiggive/matchbot#1917

Might need a little more QA discussion but this ought to be safe enough to do soon.


NoelLH:
Confirmed no specific QA blockers, but probably now leave until 29/4.

PR #1925, opened 2026-04-17
MAT-487 Continue pushing fund totals to Salesforce for up to 8 days after a campaign's close

NoelLH:


bdsl (approved):
LGTM to me as a short term fix. Wondering there might be a longer term change we could make to push any fund totals that have recently changed or where the latest total isn't already known to SF, rather than relying on the fact  that we don't normally issue refunds more than a week after a campaign closes.

PR #1928, opened 2026-04-21
Switch Stripe charge event lock errors HTTP code to 409

NoelLH:
Trying to keep the behaviour we want in other respects – halt processing and let Stripe know there was an error so their systems retry the event delivery – while avoiding noise in our alarms which treat any 5xx as unexpected.


bdsl (approved):
LGTM

PR #1929, opened 2026-04-21
MAT-406: Charge new fees for donations made from 1 July 2026

bdsl:


src/Application/Fees/Calculator.php
```diff
@@ -69,7 +81,8 @@ public static function calculate(
             currencyCode: $currencyCode,
             cardBrand: $cardBrand,
             cardCountry: $cardCountry,
-            hasGiftAid: $hasGiftAid
+            hasGiftAid: $hasGiftAid,
+            effectiveDate: $mandateActivationDate ?? $donationCreationDate,
```
NoelLH, line 85: Nice, definitely beats hard-coding campaigns. The only thing is if we were to launch a further pilot before July we might need to reconsider, but that doesn't sound that likely.

src/Domain/Donation.php
```diff
@@ -2150,4 +2152,13 @@ public function paymentServiceProvider(): ?PaymentServiceProvider
     {
         return PaymentServiceProvider::tryFrom($this->psp);
     }
+
+    #[ORM\PrePersist]
+    final public function createdNow(): void
+    {
+        // this duplicates function from TimestampsTrait but it is easier for PHPStan to follow to know
+        // that the properties are intialised at construction.
```
NoelLH: Have we checked if this might add overhead by running the lifecycle hook twice?

tests/Application/Fees/WebsiteExamplesCalculatorTest.php
```diff
@@ -25,7 +25,7 @@ class WebsiteExamplesCalculatorTest extends TestCase
      * are in GBP.
      */
     public const array WORKED_EXAMPLES = [
-        'Donation made with UK Visa Card, without gift aid' => [
+        'Donation made with UK Visa Card, without gift aid, with new fees' => [
```
bdsl:
```suggestion
        'Donation made with UK Visa Card, without gift aid' => [
```

tests/Application/Fees/CalculatorTest.php
```diff
@@ -11,6 +11,13 @@
 
 class CalculatorTest extends TestCase
 {
+    private \DateTimeImmutable $preIncreaseDate;
+
+    public function setUp(): void
+    {
+        $this->preIncreaseDate = new \DateTimeImmutable('2026-06-30T23:59:59 Europe/London');
```
NoelLH, line 18: As we're unlikely to make other changes here before July I wonder if it would make sense to swap in the new rates here? I feel like this test class would be the first place I'd look to understand fees; but we could also update it at the same time we remove the date/feature flag, I don't feel that strongly.

tests/Application/Fees/WebsiteExamplesCalculatorTest.php
```diff
@@ -35,9 +35,11 @@ class WebsiteExamplesCalculatorTest extends TestCase
             //-------------------//
             'total_fee' => '0.42',
             'total_transferred_to_charity' => '9.58',
+            'creation_datetime' => '2026-06-30T23:59:59 Europe/London',
+            'mandate_activation_date' => null,
         ],
-//
-        'Donation made with UK Visa Card, with gift aid' => [
+        //
+        'Donation made with UK Visa Card, with gift aid, new fees' => [
```
NoelLH:
```suggestion
        'Donation made with UK Visa Card, with gift aid' => [
```

NoelLH (commented):
Noted a few things inline, otherwise LGTM

src/Application/Fees/Calculator.php
bdsl: Right, if we didn't want to have this rule apply to the donations from that pilot. But yes lets leave that to decide if it happens.

src/Domain/Donation.php
bdsl: I don't see how it can, since there can't be two functions with the same name available on the same object. In any case running that function twice would be a very tiny overhead compared to the database IO so I don't think I'd be concerned.

tests/Application/Fees/CalculatorTest.php
bdsl: Yep although it may actually be that everything this test does is now duplicated by the other test class based on the examples from the web page, so maybe this one could just be deleted. Would need to check if there's anything not tested in the other one. We will still need the date flag permanently for the regular giving (or until those mandates are cancelled) unless we replace it with hard coded campaign IDs which doesn't seem better.

Will change it now if you like otherwise will leave.

src/Domain/Donation.php
NoelLH: OK, I missed that the name was identical. In that case I guess I'm more confused as to why PHPStan can't find it without the repetition, and why we haven't had this before.

tests/Application/Fees/CalculatorTest.php
NoelLH: Fair enough; fine not to change it sooner

src/Domain/Donation.php
bdsl: Yeah I'm not sure, it just started complaining because of the new usage of the created date time.

tests/Application/Fees/WebsiteExamplesCalculatorTest.php
bdsl: Thanks!

src/Domain/Donation.php
NoelLH: OK, if we don't have a clear explanation then I think putting `// @phpstan-ignore property.uninitialized` on `getCreatedDateImmutable()` in the trait is quite a lot less likely to lead to future confusion and does the same job.

bdsl:
approved verbally by @NoelLH

PR #1930, opened 2026-04-22
BG2-3110: Resubmit gift aid claims for 67 donations

bdsl:


src/Migrations/Version20260422140404.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public function up(Schema $schema): void
+    {
+        // change below means the donations should be found by
+        // \MatchBot\Domain\DoctrineDonationRepository::findReadyToClaimGiftAid next time it's called.
+
+        $this->addSql(<<<EOT
+            UPDATE Donation
+            SET tbgGiftAidRequestQueuedAt = NULL
```
NoelLH:
```suggestion
            SET tbgGiftAidRequestQueuedAt = NULL,
                tbgGiftAidRequestFailedAt = NULL,
                tbgGiftAidRequestConfirmedCompleteAt = NULL,
                tbgGiftAidRequestCorrelationId = NULL,
                tbgGiftAidResponseDetail = NULL,
                salesforcePushStatus = 'pending-update'
```

NoelLH: Given claims were sent, I think this is better as it leaves the data more self-consistent.

src/Migrations/Version20260422140404.php
```diff
@@ -0,0 +1,41 @@
...
+            WHERE
+                donationStatus in ('Paid', 'Collected') AND 
+                giftAid = 1 AND 
+                tbgShouldProcessGiftAid = 1 AND
+                salesforceId in (
+                                          'a06WS00000HFZWmYAP','a06WS00000HGQUfYAP','a06WS00000HGZXwYAP','a06WS00000HGej0YAD','a06WS00000HGuvhYAD','a06WS00000HHFIhYAP','a06WS00000HHVf9YAH','a06WS00000HHWcoYAH','a06WS00000HInjdYAD','a06WS00000HKmrZYAT','a06WS00000HMMcsYAH','a06WS00000HMmvaYAD','a06WS00000HMs9hYAD','a06WS00000HMwOaYAL','a06WS00000HMxfZYAT','a06WS00000HN00qYAD','a06WS00000HOKezYAH','a06WS00000HPBeSYAX','a06WS00000HPlmlYAD','a06WS00000HPyTpYAL','a06WS00000HQDxXYAX','a06WS00000HTF0JYAX','a06WS00000HTM9pYAH','a06WS00000HZDWvYAP','a06WS00000HZXsHYAX','a06WS00000HZaejYAD','a06WS00000HbLobYAF','a06WS00000HbaXQYAZ','a06WS00000HbwPsYAJ','a06WS00000HcLg0YAF','a06WS00000HcOFhYAN','a06WS00000HcuvZYAR','a06WS00000HeAtNYAV','a06WS00000HeHtFYAV','a06WS00000HeMxiYAF','a06WS00000HeULnYAN','a06WS00000Hf0QTYAZ','a06WS00000HfNepYAF'
```
NoelLH: I think the list needs checking, will put details on Jira.

src/Migrations/Version20260422140404.php
bdsl: Thanks, done in 4997c6b0c

src/Migrations/Version20260422140404.php
bdsl: Right, fixed in 2dc83240f65 . Also made it one line per donation to make it easy to see the length.

NoelLH (approved):
Would probably also be good to have a condition like `tbgGiftAidRequestConfirmedCompleteAt IS NOT NULL` as belt and braces in general. But in this case I added it to a simulated query and confirmed it didn't affect the number of donations matched, so not essential.

I reviewed the method of extracting donation IDs and spot checked a few to confirm they were consistent between the original ticket, the methodology comment and the PR.

PR #1932, opened 2026-04-22
MAT-489 Fix our workaround for a Stripe payment method change bug

NoelLH:
We think that this can happen if a donor starts with Pay By Bank; abandons the next_action; and switches to card while saying yes to saving it for future.

Our existing logic to clear the payment_method proactively on type change was broken (probably) because we were never successfully reading the existing payment method's type in cases where it is too new to be attached to the Customer.

### DIFF
--- a/src/Client/Stripe.php
+++ b/src/Client/Stripe.php
@@ -78,11 +78,15 @@ interface Stripe
     public function createSetupIntent(StripeCustomerId $stripeCustomerId): SetupIntent;
 
     /**
+     * Gets a specific Customer's Payment Method which must be attached. Generally for saved methods.
+     * @link https://docs.stripe.com/api/payment_methods/customer
      * @throws InvalidRequestException
      */
     public function retrievePaymentMethod(StripeCustomerId $customerId, StripePaymentMethodId $methodId): PaymentMethod;
 
     /**
+     * Gets a Payment Method that is not necessarily attached to a Customer yet.
+     * @link https://docs.stripe.com/api/payment_methods/retrieve
      * @throws ApiErrorException
      */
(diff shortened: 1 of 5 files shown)


src/Client/Stripe.php
```diff
@@ -82,6 +82,11 @@ public function createSetupIntent(StripeCustomerId $stripeCustomerId): SetupInte
      */
     public function retrievePaymentMethod(StripeCustomerId $customerId, StripePaymentMethodId $methodId): PaymentMethod;
 
+    /**
+     * @throws ApiErrorException
+     */
+    public function retrievePendingPaymentMethod(StripePaymentMethodId $methodId): PaymentMethod;
```
bdsl, line 92: may be worth adding a docblock here to explain what pending means, particularly since pending is in our method name but not in the Stripe API. I guess it means it's not yet been attached to a customer which is why the `retrievePaymentMethod` method can't be used.

NoelLH: I'm not sure I am sure enough about when attaching works to make it clearer than the method name and definitely accurate. I would have called it `retrievePendingMethod()` but I think renaming the other one to make its narrower scope clearer at a glance is a bigger change than we want to do right now.

bdsl (approved):
I think worth adding docs to both `retrievePaymentMethod` and `retrievePendingPaymentMethod` to clarify the distinction, but not a blocker.

bdsl (dismissed):

NoelLH:
> I think worth adding docs to both `retrievePaymentMethod` and `retrievePendingPaymentMethod` to clarify the distinction, but not a blocker.

Thanks, done in e0f39ba8d23eaf1bc1cc835a944bc1239a4beac7

PR #1936, opened 2026-04-24
Mat 483 ccampaign status delayed merge

bdsl:


src/Application/Actions/Campaigns/Search.php
```diff
@@ -31,13 +31,10 @@ public function __construct(
         $params = $request->getQueryParams();
         $sortField = $params['sortField'] ?? 'distanceToTarget';
         $sortDirection = $params['sortDirection'] ?? 'desc';
-        /** @var 'Active'|'Expired'|'Preview'|null */
-        $status = $params['status'] ?? null;
```
bdsl, line 35: Removed 'status' here since it looks like it's not actually in use - the SearchQuery class in Frontend at https://github.com/thebiggive/donate-frontend/blob/a0afd7640da97a89d04d050bf9a38b7437bd43e0/src/app/campaign.service.ts#L269 doesn't have a 'status' field. Maybe just there for historical reasons.

src/Domain/CampaignRepository.php
```diff
@@ -383,17 +383,13 @@ public function totalMatchFundsUsed(int $campaignId): Money
...
-        // that can work.
-
         $query = $this->getEntityManager()->createQuery(
             <<<'DQL'
             SELECT campaign FROM MatchBot\Domain\Campaign campaign
             JOIN campaign.campaignStatistics statistics
             WHERE 
              campaign.charity = :charity
-             AND campaign.status IN ('Active', 'Preview', 'Expired')
+             AND campaign.isPublished = true
```
bdsl, line 394: the `= true` feels redundant but if I leave it out PhpStorm complains with "Type mismatch: should be a conditional expression"

src/Domain/CampaignRepository.php
```diff
@@ -564,28 +555,18 @@ public function updateCampaignFromSFData(Campaign $campaign, array $campaignData
...
         QueryBuilder $qb,
-        ?string $status,
         ?string $metaCampaignSlug,
         ?string $fundSlug,
         array $jsonMatchInListConditions,
         bool $filterOutTargetMet,
         ?string $term,
     ): array|null {
         $qb->andWhere($qb->expr()->eq('campaign.hidden', '0'));
+        $qb->andWhere($qb->expr()->eq('campaign.isPublished', '1'));
```
bdsl, line 567: this replaces the status check deleted below. That was just in the if non-metacamapaign branch, but I think it makes sense to always exclude any unpublished campaigns from search.

src/Domain/CampaignRepository.php
```diff
@@ -564,28 +555,18 @@ public function updateCampaignFromSFData(Campaign $campaign, array $campaignData
...
-        // @todo MAT-483 - remove marked usage of status field below and replace with use of new isPublished
-        // field. Not doing right now as we need to deploy and run the DB migration from this commit first before
-        // that can work.
-
-        if ($status !== null) {
-            $qb->andWhere($qb->expr()->eq('campaign.status', ':status'));
-            $qb->setParameter('status', $status);
-        } elseif ($metaCampaignSlug === null) {
-            $qb->andWhere('campaign.status IN (:nonExpired)');
-            // we can't rely on the status being updated when the campaign expires, so also check that the end-date is not past:
```
bdsl, line 586: " we can't rely on the status being updated when the campaign expires, so also check that the end-date is not past"

I'm not sure if there was ever a reason that checking that the end-date is not past was not sufficient. Maybe just because that would have also allowed in the null-status unpublished campaigns.

NoelLH (dismissed):
LGTM.

`endDate` is already first in an index on Campaign so seems unlikely to significantly hurt performance, ~~but I am wondering if it's possibly worth having a single field index for `isPublished` on the stats table.~~

Edit: that's already in #1937

PR #1937, opened 2026-04-24
MAT-483: Update approx status on campaign statistics to allow sor…

bdsl:
…ting without use of SF status value

### DIFF
--- a/src/Application/Commands/UpdateApproxCampaignStatus.php
+++ b/src/Application/Commands/UpdateApproxCampaignStatus.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace MatchBot\Application\Commands;
+
+use MatchBot\Domain\CampaignRepository;
+use MatchBot\Domain\CampaignStatistics;
+use MatchBot\Domain\CampaignStatisticsRepository;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Finds campaigns that with approx status of 'Preview' that are open or will be open soon and updates
+ * to Active, and finds campaigns that with approx status of 'Active' that are closed and updates to 'Expired'
+ */
+class UpdateApproxCampaignStatus extends LockingCommand
+{
+    public function __construct(
+        private CampaignStatisticsRepository $campaignStatisticsRepository,
+    ) {
+        parent::__construct();
+    }
+    #[\Override]
+    protected function doExecute(InputInterface $input, OutputInterface $output): int
+    {
+        $this->campaignStatisticsRepository->updateApproxStatuses();
+
+        return 0;
+    }
+}
--- a/src/Application/Commands/UpdateApproxCampaignStatus.php
+++ b/src/Application/Commands/UpdateApproxCampaignStatus.php
@@ -15,7 +15,7 @@ use Symfony\Component\Console\Output\OutputInterface;
  * Called from {@see CallFrequentTasks}
  */
 #[AsCommand(
-    name: 'matchbot:update-approx-campaign-stats',
+    name: 'matchbot:update-approx-campaign-status',
     description: "Updates approximate campaign statuses to support search results ordering.",
 )]
 class UpdateApproxCampaignStatus extends LockingCommand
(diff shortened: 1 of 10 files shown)


src/Domain/CampaignRepository.php
```diff
@@ -108,8 +108,10 @@ public function pullNewFromSf(Salesforce18Id $salesforceId): Campaign
         $campaign = Campaign::fromSfCampaignData($campaignData, $salesforceId, $charity);
 
         $campaign->setSalesforceLastPull(new \DateTime());
+        $campaignStatistics = $campaign->getStatistics();
 
         $this->getEntityManager()->persist($campaign);
+        $this->getEntityManager()->persist($campaignStatistics);
```
bdsl, line 114: this should ensure every campaign has statistics when it's inserted, probably no need to run any migration to make sure all the existing campaigns have associated statistics, since I ran the query below in prod and found only 5 results, the most recent from 2022, so effectively every campaign already has statistics in the DB.

```
SELECT id, Campaign.startDate, name from Campaign left JOIN CampaignStatistics on Campaign.id = CampaignStatistics.campaign_id WHERE CampaignStatistics.campaign_id is null limit 50; 
```

src/Application/Commands/UpdateApproxCampaignStatus.php
```diff
@@ -0,0 +1,34 @@
...
+use MatchBot\Domain\CampaignStatisticsRepository;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Updates approximate campaign statuses to support search results ordering.
+ *
+ * Will update any campaigns that are due to start within the next 24 hours to active, so doesn't need to be run
+ * very frequently, suggested cadence is once every six hours.
```
NoelLH: I'm wondering if for simplicity we could do this on every 'tick' or even in the same task as other campaign stats which go to the same table. Aren't the queries likely to be reasonably light on the DB?

bdsl: Yes I think you're right, done in 144926116

src/Application/Commands/UpdateApproxCampaignStatus.php
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Updates approximate campaign statuses to support search results ordering.
+ *
+ * Will update any campaigns that are due to start within the next 24 hours to active.
+ *
+ * Called from {@see CallFrequentTasks}
+ */
+#[AsCommand(
+    name: 'matchbot:update-approx-campaign-stats',
```
NoelLH:
```suggestion
    name: 'matchbot:update-approx-campaign-status',
```
and logs of this. I think because there are more general stats too this could be confusing.

src/Domain/CampaignStatisticsRepository.php
```diff
@@ -32,4 +32,65 @@ public function getStatistics(Campaign $campaign): CampaignStatistics
...
+                SELECT IDENTITY(cs)
+                FROM MatchBot\Domain\CampaignStatistics cs
+                JOIN cs.campaign c
+                WHERE c.isPublished = true
+                AND c.startDate < :tommorrow
+                AND c.endDate > CURRENT_TIMESTAMP()
+                AND cs.approxStatus = '$preview'
+            DQL
+        );
+        $query->setParameter('tommorrow', new \DateTimeImmutable('+1 day'));
```
NoelLH:
```suggestion
                AND c.startDate < :tomorrow
                AND c.endDate > CURRENT_TIMESTAMP()
                AND cs.approxStatus = '$preview'
            DQL
        );
        $query->setParameter('tomorrow', new \DateTimeImmutable('+1 day'));
```

NoelLH (commented):
Minor things inline, LGTM otherwise

src/Application/Commands/UpdateApproxCampaignStatus.php
bdsl: Thanks, that name was a mistake, should be 812909a9b

src/Domain/CampaignStatisticsRepository.php
bdsl: Thanks, fixed in fed9d83e

bdsl:
Probably best to hold this back unmerged until after Earth Raise, then we can merge and get it into prod tomorrow.

PR #1939, opened 2026-04-26
Fix incorrect enum/string comparison in donation Confirm

NoelLH:
Without this fix, the incorrect comparison lets 'stripe' mode donations with no confirmation token get further than intended, causing a noisy error log.


bdsl (commented):
LGTM, phpstorm actually shows that the original is broken although only with a weak warning:

Image

Not sure why we don't have an issue from PHPStan about it - I might see about bringing Psalm back in and that would have told us about this.

PR #1947, opened 2026-04-29
MAT-491: Prevent injection of DateTimeImmutable

bdsl:


NoelLH (approved):
LGTM.

I had a quick review of the codebase generally – the only other place which jumps out that might also be slightly safer this way is `TakeRegularGivingDonations`, where I'm wondering if we could make the caveats around the mock clock easier to follow by using a Clock consistently. But it doesn't look likely to have the same bug as it doesn't enqueue its main work, so can definitely be for a later PR.

I primarily focused on direct calls in Message handlers for any which are a high risk of having a bug, and couldn't find any. They all seem to either already use Clock or directly instantiate a new DateTimeImmutable on every use, like `DonationMatchCheckHandler::syncFundUsage()`, which should be fine too.

PR #1948, opened 2026-04-29
MAT-483: Remove remaining usages of CampaignStatus field from SF

bdsl:
Replaced with the campaignStatistics.pproxStatus which is not derived from any 'status' field in SF

### DIFF
--- a/integrationTests/CampaignRepositoryTest.php
+++ b/integrationTests/CampaignRepositoryTest.php
@@ -9,6 +9,7 @@ use MatchBot\Domain\ApplicationStatus;
 use MatchBot\Domain\Campaign;
 use MatchBot\Domain\CampaignRepository;
 use MatchBot\Domain\CampaignStatistics;
+use MatchBot\Domain\CampaignStatus;
 use MatchBot\Domain\Charity;
 use MatchBot\Domain\CharityResponseToOffer;
 use MatchBot\Domain\Money;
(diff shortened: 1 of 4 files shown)


integrationTests/CampaignRepositoryTest.php
```diff
@@ -151,7 +155,7 @@ public function testSearchSortsByStatus(): void
             sortDirection: 'desc',
             offset: 0,
             limit: 6,
-            metaCampaignSlug: null,
+            metaCampaignSlug: 'some-slug',
```
bdsl, line 158: We have to supply a metacampaign slug for this test to work now since its more realistic and the non-slug explore page excludes all expired campaigns.

src/Domain/CampaignRepository.php
```diff
@@ -389,7 +389,7 @@ public function findCampaignsForCharityPage(Charity $charity, \DateTimeImmutable
              campaign.charity = :charity
              AND campaign.isPublished = true
              AND (statistics.donationSum.amountInPence > 0 OR campaign.endDate > :at OR campaign.endDate IS NULL)
-             ORDER BY campaign.status ASC, campaign.endDate ASC 
+             ORDER BY statistics.approxStatus ASC, campaign.endDate ASC
```
shaundbs: This looks like a good cleanup . Main thing I’d like to verify is that ordering by statistics.approxStatus ASC is guaranteed to produce Active → Expired → Preview? If that order is critical, should we make it explicit rather than relying on enum/string ordering?

bdsl: Hi, yes I think it is guaranteed to be alphabetical order but I agree this is a good time to refactor to make it explicit instead of relying on the spellings. Will do that now.

bdsl: @shaundbs commit f9a20e90c adds case statements to explicitly specify the  Active → Expired → Preview order and not rely on the alphabetical order.

shaundbs: Looks good! Thanks @bdsl approved

shaundbs (approved):

PR #1952, opened 2026-04-30
MAT-483: Stop using Metacampaign status and mastercampaign status fields

bdsl:
Better to calculate status entirely in matchbot & frontend based on other fields

### DIFF
--- a/src/Domain/MetaCampaign.php
+++ b/src/Domain/MetaCampaign.php
@@ -26,8 +26,6 @@ class MetaCampaign extends SalesforceReadProxy
 {
     use TimestampsTrait;
 
-    public const string STATUS_VIEW_CAMPAIGN = 'View campaign';
-
     /** Metacampaigns from before this date may not have accurate data so should not be indexed */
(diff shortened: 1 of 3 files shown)


src/Domain/MetaCampaign.php
```diff
@@ -273,10 +262,6 @@ public function getCurrency(): Currency
 
     public function getStatusAt(\DateTimeImmutable $date): ?string
     {
```
bdsl, line 264: this getter is only used to pass the result directly to Frontend, which has it here https://github.com/thebiggive/donate-frontend/blob/7d5076c66d773f9abbcd23cb9a8389547cde8c86/src/app/metaCampaign.model.ts#L12 , but all references to frontend don't use the status value in isolation, they also look at the browsers clock to compare to the campaign start and end times. 

And especially since almost everything is already null in the DB it does seem very much like we're not losing anything by no longer referencing either `$this->masterCampaignStatus` or `$this->status`.

NoelLH (approved):

PR #1954, opened 2026-04-30
MAT-153: Parse ryftAccountId into DB

shaundbs:
- Parse the ryftAccountId so that it is stored in the matchbot database


bdsl (approved):
LGTM

PR #1957, opened 2026-05-05
MAT-483: Stop expecting status from SF

bdsl:
Don't merge until after tomorrow's SF deploy.


src/Domain/Campaign.php
```diff
@@ -568,7 +545,7 @@ final public function updateFromSfPull(
     }
 
     /**
-     * This *does not* use the campaign status in the DB as received from Salesforce, as that may be outated,
+     * This *does not* use the campaign status as received from Salesforce, as that may be outated,
```
NoelLH: Could probably drop this whole block now?

NoelLH (dismissed):

src/Domain/Campaign.php
bdsl: done in 207f023d2

PR #1958, opened 2026-05-06
Don 1192 ryft null stripe

bdsl:


.psalm/baseline.xml
```diff
@@ -90,7 +90,6 @@
       <code><![CDATA[$ryftPaymentSessionId]]></code>
     </MixedAssignment>
     <PossiblyNullArgument>
-      <code><![CDATA[$confirmationTokenId]]></code>
```
bdsl, line 93: showing that the old version of the code was detectable by Psalm as wrong, but we weren't running Psalm at the time it was written.

bdsl: Potentially it would be worth going through this whole file since it's not hugely long and trying to fix each issue.

NoelLH (dismissed):

PR #1964, opened 2026-05-07
Fix some psalm errors

bdsl:


src/Application/CustomMySQLSchemaManager.php
```diff
@@ -19,8 +19,9 @@ class CustomMySQLSchemaManager extends MySQLSchemaManager
         ['Charity', 'FULLTEXT_NORMALISED_NAME'],
     ];
 
-    private const array GENERATED_COLUMNS = [
+    private const array IGNORED_COLUMNS = [
         ['Campaign', 'normalisedName'],
+        ['MetaCampaign', 'masterCampaignStatus'],
```
bdsl, line 24: avoiding making a destructive DB change by dropping the column at the same time as dropping the property from the class.

shaundbs (approved):

PR #1974, opened 2026-05-19
DON-1192 Fix Ryft deriving amount donor paid & related assertions breaking Confirm

NoelLH:


src/Domain/Donation.php
```diff
@@ -1616,19 +1616,24 @@ public function collectFromStripeCharge(
...
         array $paymentSession,
-        Money $totalPaidByDonor,
+        Money $netAmount,
         Money $originalFeeFractional,
         \DateTimeImmutable $at,
     ): void {
         $this->donationStatus = DonationStatus::Collected;
         $this->collectedAt = $at;
-        $this->totalPaidByDonor = $totalPaidByDonor->toNumericString();
+        // We have to add `amount` which is net amount reported by Ryft after their equivalent of application fee,
```
bdsl:
```suggestion
        // We have to add `netAmount` which is net amount reported by Ryft after their equivalent of application fee,
```

bdsl (dismissed):
LGTM

PR #1975, opened 2026-05-20
DON-1192: Add notification email for Ryft donation

bdsl:


src/Domain/DonationService.php
```diff
@@ -1252,4 +1245,19 @@ public function throwIfUnsuccessful(PaymentIntent $intent): void
             );
         }
     }
+
+    public function notifyDonationSuccesIfRequired(Donation $donation, bool $donationWasPreviouslyCollected): void
```
shaundbs, line 1249: Does this need to be public? If it’s only used internally by `DonationService`, I’d keep it private to avoid expanding the service API surface.

PR #1976, opened 2026-05-20
DON-1192: Store ryft payment session ID on donation for reference

bdsl:
For now just for manual reference via DB, but may be used more in future.


shaundbs (approved):
Looks sensible overall, storing the Ryft session ID at creation time gives us a useful internal reference and the client secret handling looks cleaner now.

PR #1981, opened 2026-05-26
MAT-490 Update campaign statistics on fund amount changes

NoelLH:
And various related refactoring to make those steps happen in a consistent way, with cleaner dependency chains.

### DIFF
--- a/src/Domain/CampaignService.php
+++ b/src/Domain/CampaignService.php
@@ -2,9 +2,12 @@
 
 namespace MatchBot\Domain;
 
+use Assert\AssertionFailedException;
 use Assert\InvalidArgumentException;
 use Assert\LazyAssertionException;
+use Doctrine\ORM\EntityManagerInterface;
 use MatchBot\Application\Assertion;
+use MatchBot\Application\Commands\UpdateCampaignDonationStats;
 use MatchBot\Application\HttpModels\Campaign as CampaignHttpModel;
 use MatchBot\Application\HttpModels\MetaCampaign as MetaCampaignHttpModel;
 use MatchBot\Client\Campaign as CampaignClient;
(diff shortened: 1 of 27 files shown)


tests/Application/Actions/Hooks/StripePaymentsUpdateTest.php
```diff
@@ -138,33 +132,33 @@ public function testMissingSignature(): void
         $this->assertSame(400, $response->getStatusCode());
     }
 
+    /**
+     * @psalm-suppress UndefinedPropertyAssignment
+     * supressing deprecation notices for now on setting properties dynamically w.r.t Stripe
+     * balance txns. Risk is low doing this in test code, and may get mutation tests working again.
```
NoelLH, line 138: Comment moved from the inline phpstan counterpart that the above replaces

src/Domain/CampaignService.php
```diff
@@ -398,4 +416,48 @@ public function getAmountRaisedForMetaCampaign(MetaCampaign $metaCampaign): Mone
...
+        $campaignId = $campaign->getId();
+        \assert($campaignId !== null);
+        $matchFundsUsed = $this->campaignRepository->totalMatchFundsUsed($campaignId);
+        $donationSum = $this->campaignRepository->totalCoreDonations($campaign);
+
+        $statistics = $campaign->getStatistics(); // New & zeroes if not done before.
+
+        try {
+            $changed = $statistics->setTotals(
+                at: new \DateTimeImmutable('now'),
```
bdsl: maybe worth injecting?

src/Domain/FundRepository.php
```diff
@@ -38,12 +38,14 @@ public function setMatchingAdapter(Matching\Adapter $matchingAdapter): void
     /**
      * @param Campaign  $campaign
      * @param DateTimeImmutable $at
+     * @return bool Whether any funding changed. Use to decide about e.g. stats regeneration.
      * @throws Client\NotFoundException if Campaign not found on Salesforce
      */
-    public function pullForCampaign(Campaign $campaign, DateTimeImmutable $at): void
+    public function pullForCampaign(Campaign $campaign, DateTimeImmutable $at): bool
```
bdsl, line 44:
```suggestion
    #[\NoDiscard("as stats will need to be regenerated iff it returns true")]
    public function pullForCampaign(Campaign $campaign, DateTimeImmutable $at): bool
```

PR #1985, opened 2026-05-28
BG2-3308 Patch unreliably-set isPublished value for campaigns created before late April

NoelLH:
Should correct some where application fields were not factored in.

### DIFF
--- a/src/Migrations/Version20260528112225.php
+++ b/src/Migrations/Version20260528112225.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MatchBot\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20260528112225 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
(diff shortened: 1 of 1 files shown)


src/Migrations/Version20260528112225.php
```diff
@@ -0,0 +1,31 @@
...
+    public function getDescription(): string
+    {
+        return 'Based isPublished on ready for campaigns not created more recently than the new logic';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // Patch all campaigns before the migration that set this unreliably went to Production (and a few more by
+        // setting it at a slightly earlier time than the merge of https://github.com/thebiggive/matchbot/pull/1942)
+        $this->addSql('UPDATE Campaign SET isPublished = ready WHERE createdAt > "2026-04-29 12:39:00"');
```
bdsl: should this be `createdAt <` ?

NoelLH: Yes, thanks! aa4c605eae42ed71ee7cb16ab3044518e002f34d

Have now tried simulating the updated query and it is giving ~25% of the rows as changed which sounds about right.

bdsl (approved):
LGTM - I'm not sure it was wrong at the time it was created but isn't consistent with either how we're now using it or what's being generated from SF for new data, so good to change the old to match.

PR #1988, opened 2026-05-28
Correct some errors found by new PHPStan version

bdsl:


src/Domain/FundRepository.php
```diff
@@ -131,7 +131,7 @@ public function pullForCampaign(Campaign $campaign, DateTimeImmutable $at): bool
     }
 
     /**
-     * @param array{currencyCode: ?string, name: ?string, slug: ?string, type: string, id:string, ...} $fundData
+     * @param array{currencyCode: ?string, name: ?string, slug: ?string, type: string, id:string, ...<mixed>} $fundData
```
bdsl: Doesn't work with Psalm

```suggestion
     * @param array{currencyCode: ?string, name: ?string, slug: ?string, type: string, id:string, ...} $fundData
```

PR #1989, opened 2026-05-28
SO-78 Store structured campaign locations with UK region support

NoelLH:
Currently requires https://github.com/thebiggive/salesforce/pull/2924 in production before merging.

Doesn't yet include populating the new data structure for legacy data, that can be in a later PR probably.

### DIFF
--- a/integrationTests/AcceptCampaignPushFromSFTest.php
+++ b/integrationTests/AcceptCampaignPushFromSFTest.php
@@ -2,16 +2,13 @@
 
 namespace MatchBot\IntegrationTests;
 
-use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\EntityManagerInterface;
 use GuzzleHttp\Psr7\ServerRequest;
 use MatchBot\Application\Auth\SalesforceAuthMiddleware;
 use MatchBot\Domain\CampaignRepository;
-use MatchBot\Domain\CharityRepository;
 use MatchBot\Domain\MetaCampaignRepository;
 use MatchBot\Domain\MetaCampaignSlug;
 use MatchBot\Domain\Salesforce18Id;
-use MatchBot\Tests\Application\Actions\Hooks\StripeTest;
 use MatchBot\Tests\TestCase;
 use MatchBot\Client;
 
--- a/schema/CampaignLocation.sql
+++ b/schema/CampaignLocation.sql
@@ -6,7 +6,7 @@
 -- information only.
 
 CREATE TABLE `CampaignLocation` (
-  `countryName` varchar(50) DEFAULT NULL,
+  `countryName` varchar(100) DEFAULT NULL,
   `regionCode` varchar(10) DEFAULT NULL,
   `id` int unsigned NOT NULL AUTO_INCREMENT,
   `campaign_id` int unsigned NOT NULL,
--- a/schema/CampaignLocation.sql
+++ b/schema/CampaignLocation.sql
@@ -1,0 +1,16 @@
+-- Auto generated file, do not edit.
+-- run ./matchbot matchbot:write-schema-files to update
+
+-- Note that CHARSET and COLLATE details have been removed to allow sucessful comparisons between schemas
+-- on dev machines and CircleCI. Do not use these files to create a database. They are provided for
+-- information only.
+
+CREATE TABLE `CampaignLocation` (
+  `countryName` varchar(50) DEFAULT NULL,
+  `regionCode` varchar(10) DEFAULT NULL,
+  `id` int unsigned NOT NULL AUTO_INCREMENT,
+  `campaign_id` int unsigned NOT NULL,
+  PRIMARY KEY (`id`),
+  KEY `IDX_6C25EDB1F639F774` (`campaign_id`),
+  CONSTRAINT `FK_6C25EDB1F639F774` FOREIGN KEY (`campaign_id`) REFERENCES `Campaign` (`id`) ON DELETE CASCADE
+)
--- a/src/Application/HttpModels/Campaign.php
+++ b/src/Application/HttpModels/Campaign.php
@@ -31,7 +31,7 @@ readonly class Campaign
      * @param list<array{amount: float, description: string}> $budgetDetails
      * @param list<string> $categories
      * @param list<string> $countries
-     * @param list<array{countryName: string|null, regionCode: string|null}> $locations
+     * @param array<int, array{countryName: ?string, regionCode: ?string}> $locations array is simplest way to match type; ArrayCollection is not covariant.
      * @param list<array{person: string, quote: string}> $quotes
      * @param list<array{content: string, modifiedDate: string}> $updates
      * @param ?array{provider: string, key: string} $video
--- a/src/Application/HttpModels/Campaign.php
+++ b/src/Application/HttpModels/Campaign.php
@@ -31,6 +31,7 @@ readonly class Campaign
      * @param list<array{amount: float, description: string}> $budgetDetails
      * @param list<string> $categories
      * @param list<string> $countries
+     * @param list<array{countryName: string|null, regionCode: string|null}> $locations
      * @param list<array{person: string, quote: string}> $quotes
      * @param list<array{content: string, modifiedDate: string}> $updates
      * @param ?array{provider: string, key: string} $video
@@ -201,6 +202,19 @@ readonly class Campaign
             example: true
         )]
         public bool $isMatched,
+        #[OA\Property(
+            property: "locations",
+            description: "List of countries or regions where the campaign operates",
+            type: "array",
+            items: new OA\Items(
+                type: "object",
+                properties: [
+                    new OA\Property(property: "countryName", type: "string", example: "Ghana", nullable: true),
+                    new OA\Property(property: "regionCode", type: "string", example: "E12000004", nullable: true)
+                ]
+            )
+        )]
+        public array $locations,
         #[OA\Property(
             property: "logoUri",
             description: "URI for the campaign's logo",
--- a/src/Client/Campaign.php
+++ b/src/Client/Campaign.php
@@ -59,6 +59,7 @@ use MatchBot\Domain\MetaCampaignSlug;
  *     amountRaised: float,
  *     summary: string,
  *     countries: list<string>,
+ *     locations: list<array{countryName: ?string, regionCode: ?string}>,
  *     categories: list<string>,
  *     budgetDetails: list<array{amount: float, description: string}>,
  *     beneficiaries: list<string>,
--- a/src/Domain/Campaign.php
+++ b/src/Domain/Campaign.php
@@ -5,13 +5,11 @@ declare(strict_types=1);
 namespace MatchBot\Domain;
 
 use DateTimeImmutable;
-use DateTimeInterface;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Event\PrePersistEventArgs;
 use Doctrine\ORM\Mapping as ORM;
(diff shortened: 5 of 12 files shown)


src/Domain/Campaign.php
```diff
@@ -669,4 +678,29 @@ public function getStatistics(): CampaignStatistics
...
+    }
+
+    /**
+     * @return array<int, array{countryName: ?string, regionCode: ?string}>
+     */
+    public function getLocationsForApi(): array
+    {
+        // Ensure sequence gaps from clear()ing are removed.
+        return array_values(
+            array_map(
```
NoelLH, line 706: `Collection->map()` made psalm confused about whether generics' types matched / whether closure's pure.

src/Client/Campaign.php
```diff
@@ -59,6 +59,7 @@
  *     amountRaised: float,
  *     summary: string,
  *     countries: list<string>,
+ *     locations: list<array{countryName: ?string, regionCode: ?string}>,
```
shaundbs, line 61: Should this also be a Set and then converted back to a sorted list before returning? That would avoid duplicate countries when multiple locations share the same country?

src/Domain/CampaignLocation.php
```diff
@@ -0,0 +1,51 @@
...
+    #[ORM\Column(length: 50, nullable: true)]
+    private ?string $countryName = null;
+
+    /**
+     * @var string|null ONS code, set when campaigns in the UK have specified regional focus.
+     */
+    #[ORM\Column(length: 10, nullable: true)]
+    private ?string $regionCode = null;
+
+    public function __construct(Campaign $campaign, ?string $countryName, ?string $regionCode)
```
shaundbs, line 34: Do we want to allow a CampaignLocation where both countryName and regionCode are null? If not, should the constructor or replaceLocations() reject empty rows?

src/Client/Campaign.php
NoelLH: There is no such type as a Set in PHP.

We already prevent making duplicates generally in the new interface, although it's true that legacy campaigns might show up with multiple copies of a country. I think if so that would be better tidied up on the Salesforce side too though. If for some reason that's hard we could use `array_unique()` in PHP in a later iteration.

src/Domain/CampaignLocation.php
NoelLH: I'm not sure how that would work at schema level. It's also a reflection of data from Salesforce, so it's too late to 'reject' anything really, we have to reflect what we saved for the charity.

src/Client/Campaign.php
shaundbs: Even if SF should ideally enforce uniqueness, I’d consider handling it defensively here too, I agree something like array_unique() + sort() before returning so the API response stays stable/alphabetical.

Not a blocker for me though if we want to keep this iteration smaller and tidy up legacy data separately.

shaundbs (dismissed):

bdsl:
> Currently requires https://github.com/thebiggive/salesforce/pull/2924 in production before merging.

Since we don't expect that I guess until Wednesday afternoon I think it might be worth adjusting to make this mergeble first - I haven't reviewed the code yet but I'm guess would need marking some properties as optional in the expected JSON from SF.

integrationTests/AcceptCampaignPushFromSFTest.php
bdsl, line 1: nice cleanup, I might look for a tool at some point to automate removing unused `use`s, although it's trickier if they're only used in docblocks.

schema/CampaignLocation.sql
```diff
@@ -0,0 +1,16 @@
...
+
+CREATE TABLE `CampaignLocation` (
+  `countryName` varchar(50) DEFAULT NULL,
+  `regionCode` varchar(10) DEFAULT NULL,
+  `id` int unsigned NOT NULL AUTO_INCREMENT,
+  `campaign_id` int unsigned NOT NULL,
+  PRIMARY KEY (`id`),
+  KEY `IDX_6C25EDB1F639F774` (`campaign_id`),
+  CONSTRAINT `FK_6C25EDB1F639F774` FOREIGN KEY (`campaign_id`) REFERENCES `Campaign` (`id`) ON DELETE CASCADE
+)
```
bdsl, line 16: wondering whether it might be better to have a many-to-many instead of this many locations to one campaign. Have you already considered pros and cons? Might make it slightly easier to check that we're not accidentally referring to the same country by two different names or anything like that.

src/Application/HttpModels/Campaign.php
```diff
@@ -31,6 +31,7 @@
      * @param list<array{amount: float, description: string}> $budgetDetails
      * @param list<string> $categories
      * @param list<string> $countries
+     * @param array<int, array{countryName: ?string, regionCode: ?string}> $locations array is simplest way to match type; ArrayCollection is not covariant.
```
bdsl, line 34: maybe works as slightly simpler:


```suggestion
     * @param list<array{countryName: ?string, regionCode: ?string}> $locations array is simplest way to match type; ArrayCollection is not covariant.
```

?

src/Domain/Campaign.php
```diff
@@ -303,6 +308,10 @@ public static function fromSfCampaignData(array $campaignData, Salesforce18Id $s
             rawData: $campaignData,
             hidden: $campaignData['hidden'],
         );
+
+        $campaign->replaceLocations($campaignData['locations']);
```
bdsl: what's the reason for calling replaceLocations here and not passing the locations into the constructor? It seems a shame to have to modify the object after construction if we can easily avoid it.

src/Domain/Campaign.php
```diff
@@ -669,4 +678,29 @@ public function getStatistics(): CampaignStatistics
     {
         return $this->campaignStatistics ?? CampaignStatistics::zeroPlaceholder($this, new \DateTimeImmutable('now'));
     }
+
+    /**
+     * @param list<array{countryName: ?string, regionCode: ?string}> $locationsData
+     */
+    public function replaceLocations(array $locationsData): void
+    {
+        $this->locations->clear();
```
bdsl, line 689: Good to know we have to use this and not just replace the collection with a new one - I see there's a proposal to enforce it in phpstan-doctrine https://github.com/phpstan/phpstan-doctrine/issues/505

src/Domain/Campaign.php
```diff
@@ -669,4 +678,29 @@ public function getStatistics(): CampaignStatistics
...
+        foreach ($locationsData as $locData) {
+            $this->locations->add(new CampaignLocation($this, $locData['countryName'] ?? null, $locData['regionCode'] ?? null));
```
bdsl: long line, worth splitting I think also worth adding the param names for safety since we have two of the same type:

```suggestion
            $this->locations->add(new CampaignLocation(
                campaign: $this,
                countryName: $locData['countryName'] ?? null,
                regionCode: $locData['regionCode'] ?? null
            ));
```

src/Domain/CampaignLocation.php
bdsl: Or one where neither are null. Don't we normally expect exactly one to be null? I agree it's later than ideal to reject here since we can't throw to any UI, but we could still reject if it would be data that the FE won't be able to do anything sensible with, and possibly alarm. We do replace any invalid URL on charities with `null` here:

https://github.com/thebiggive/matchbot/blob/54d8b97b1a9146b2904ffdf1205c114b87cf4019/src/Domain/Charity.php#L347

src/Domain/CampaignLocation.php
```diff
@@ -0,0 +1,51 @@
...
+     */
+    #[ORM\ManyToOne(targetEntity: Campaign::class, inversedBy: 'locations')]
+    #[ORM\JoinColumn(nullable: false, onDelete: 'CASCADE')]
+    private Campaign $campaign;
+
+    /**
+     * @var string|null Set when the country overall is on the campaign's list.
+     */
+    #[ORM\Column(length: 50, nullable: true)]
+    private ?string $countryName = null;
```
bdsl, line 26: probably not needed for now but are we likely to want to expand this to also have some form of code for the country? Maybe an instance of `\PrinsFrank\Standards\Country\CountryAlpha2` or equivalent?

src/Client/Campaign.php
bdsl: You can hack the PHP array into a set by using the keys to store strings and ignoring the values. But anyway this isn't just a PHP type here, it's effectively a type for JSON data, and JSON definitely doesn't have sets.

bdsl (commented):
Review done, added some comments otherwise all LGTM. It does still look like it should be simple enough to make this mergeable in advance of the related salesforce deployment by making the new property optional and defaulting the value to empty.

src/Domain/CampaignLocation.php
NoelLH: Yep as discussed the other day, if this was better standardised at the point of collection it would be nice but as it stands I don't think there is enough benefit to reworking it now.

src/Domain/CampaignLocation.php
NoelLH: We've skipped persisting data where we know that looser built-in or legacy code validation in Salesforce is liable to mean a mismatch. But here the data is either a country string or a join to the brand new table of regions, so it seems far less likely that this would be an issue in practice.

NoelLH:
> Since we don't expect that I guess until Wednesday afternoon I think it might be worth adjusting to make this mergeble first - I haven't reviewed the code yet but I'm guess would need marking some properties as optional in the expected JSON from SF.

As I suspected last week there's too much else happening for this to be likely to be useful before then anyway, so I don't think the extra work of adding a step gains us anything.

src/Application/HttpModels/Campaign.php
NoelLH: The explanation on this line is I guess slightly out of date but I did try this before. Tried again and confirmed it is still a type error according to psalm:

```
ERROR: ArgumentTypeCoercion - src/Domain/CampaignService.php:260:24 - Argument 25 of MatchBot\Application\HttpModels\Campaign::__construct expects list<array{countryName: null|string, regionCode: null|string}>, but parent type array<int, array{countryName: null|string, regionCode: null|string}> provided (see https://psalm.dev/193)
            locations: $campaign->getLocationsForApi(),
```

src/Domain/CampaignLocation.php
NoelLH: As the rationale is high confidence in the data I've added a `throw` on double nulls: 6cd629fdd615a33a9e8f21b965830c8ddee23943

schema/CampaignLocation.sql
NoelLH: Had a bit more of a think about this.

I think it's borderline but for now the added query complexity and the realities of countries being simple strings in Salesforce makes it not quite worth it to my mind. And for UK regions, the only thing we store is the ONS code so there's not really anything 'meta' to possibly go out of sync.

It's also pretty easy to make the structure more complex and decouple later if we find a concrete benefit to that, and likely a bit messier to do the inverse.

src/Domain/Campaign.php
NoelLH: I think no reason, moved to the update fn that create & update already share in 55f5a589bc3a9c1dc76ea74ce7481fa4136d8f85

bdsl (commented):
Looks good so far

src/Client/Campaign.php
NoelLH: Salesforce now cannot return duplicates, including with legacy data, following the latest fix in https://github.com/thebiggive/salesforce/commit/0e3ffc9b06da72a67f238777bc74afe2a57f5df0 so IMO this doesn't need more

PR #1992, opened 2026-06-03
BG2-3145 Stop sending Stripe per-donation descriptors

NoelLH:
These can only be combined with account defaults which isn't what we want to give donors the clearest refs we can.


bdsl (approved):
LGTM

PR #1994, opened 2026-06-04
BG2-3065: Start running linting, static analysis and guarding with Mago

bdsl:
Installation here is via the composer wrapper. I tried installing with the first reccomended method at https://mago.carthage.software/latest/en/guide/installation/ which using curl to run carthage's shell script, but that kept erroring on the verification stage.

Config settings are mostly defaults from mago init and/or things it read out of our composer.json

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,6 +66,8 @@ jobs:
 
       - run: composer run sa:check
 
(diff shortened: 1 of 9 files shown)


NoelLH (approved):
LGTM so far, can iterate on rules and baseline size as discussed

PR #1995, opened 2026-06-04
BG2-3308 Be consistent about checking application status when listing campaigns

NoelLH:


src/Domain/CampaignRepository.php
```diff
@@ -543,18 +550,9 @@ private function filterForSearch(
         if ($metaCampaignSlug === null) {
             $qb->andWhere($qb->expr()->gt('campaign.endDate', ':now'));
             $qb->setParameter('now', $this->clock->now());
-        } else {
-            $qb->andWhere('campaign.isPublished = true');
```
NoelLH, line 547: This was duplicated, line 540 does the same.

bdsl:
Probably worth merging `develop` into this branch to get the new mago checks.

src/Domain/CampaignRepository.php
```diff
@@ -26,6 +26,13 @@
 class CampaignRepository extends SalesforceReadProxyRepository
 {
     private ClockInterface $clock;  // @phpstan-ignore property.uninitialized
+    private string $appStatusWhereClause = <<<DQL
+        campaign.metaCampaignSlug IS NULL OR
+        (
+            campaign.relatedApplicationStatus = 'Approved' AND
+            campaign.relatedApplicationCharityResponseToOffer = 'Accepted'
+        )
+    DQL;
```
bdsl, line 35: can we make this a constant?

bdsl (approved):
Added a couple of comments, otherwise all LGTM

src/Domain/CampaignRepository.php
NoelLH: I think that would then mean it needs reassignment to a var each time we want heredoc interpolation, so not sure that's worth it?

NoelLH:
> Probably worth merging `develop` into this branch to get the new mago checks.

Done in 4b279dcb1aee29cb0507663115d641544453a1a4 and it's still green.

Not related to the PR but out of interest, do you know why it's unhappy with fast route @bdsl?
>  `WARN Encountered 15 parsing errors in file 'vendor/nikic/fast-route/****/HackTypechecker/fixtures/empty_options.php'. Codebase analysis may be incomplete.`
> `WARN Encountered 9 parsing errors in file 'vendor/nikic/fast-route/****/HackTypechecker/fixtures/no_options.php'. Codebase analysis may be incomplete.`
> `WARN Encountered 83 parsing errors in file 'vendor/nikic/fast-route/****/HackTypechecker/fixtures/all_options.php'. Codebase analysis may be incomplete.`

src/Domain/CampaignRepository.php
bdsl: Fair enough, let's leave it. And I see we can't make it a read only property either since those aren't allowed to have values set at declaration. I found the open PHP issue to allow class constants in strings but since it's not an RFC I don't think it's going anywhere. https://github.com/php/php-src/issues/11018

bdsl:
> > Probably worth merging `develop` into this branch to get the new mago checks.
> 
> Done in [4b279dc](https://github.com/thebiggive/matchbot/commit/4b279dcb1aee29cb0507663115d641544453a1a4) and it's still green.
> 
> Not related to the PR but out of interest, do you know why it's unhappy with fast route @bdsl?

I just looked now, the issue with those files is that they have `.php` extensions but they're not actually in php, they're in hack language. We can probably exclude them for mago.

PR #1996, opened 2026-06-04
Bg2 3065 mago lint fix

bdsl:


bdsl:
Closing for now as mass fixing lint issues causes analysis issues. Most of them go away if [inline-variable-return](https://mago.carthage.software/1.29.0/en/tools/linter/rules/#redundancy) lint option is  turned off since that option breaks variable typing in repositories, but not all go away. I think we could start using mago without fixing all this initially.

PR #1999, opened 2026-06-05
SO-78 Migrate existing Campaign countries data to the new table

NoelLH:
Diff is largely Copilot/Gemini's work except the comment about collations.

My initial prompt:
> Can you propose a clean way to populate CampaignLocations with the countries from JSON `$salesforceData`'s countries property? If we can do it in SQL in a migration that will be easiest to run. Ideally we'd check for existing records and make sure it can't insert a country which already has a record connected to the same Campaign. (regionCode will be null for all rows added this way)

### DIFF
--- a/src/Migrations/Version20260605135515.php
+++ b/src/Migrations/Version20260605135515.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MatchBot\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * SO-78 Populate CampaignLocations with countries from Campaign salesforceData
+ */
+final class Version20260605135515 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Populate CampaignLocations with countries from Campaign salesforceData';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // Looks like MySQL 8+ default collation to utf8mb4_0900_ai_ci which means CampaignLocation
+        // has that in all environments, since it is new and we didn't specify another.
+        $this->addSql(<<<EOT
+            INSERT INTO CampaignLocation (campaign_id, countryName, regionCode)
+            SELECT c.id, jt.countryName, NULL
+            FROM Campaign c
+            JOIN JSON_TABLE(c.salesforceData, '$.countries[*]' COLUMNS (countryName VARCHAR(100) PATH '$')) jt
+            WHERE jt.countryName IS NOT NULL
+              AND NOT EXISTS (
+                  SELECT 1 
+                  FROM CampaignLocation cl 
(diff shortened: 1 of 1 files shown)


src/Migrations/Version20260605135515.php
```diff
@@ -0,0 +1,45 @@
...
+        return 'Populate CampaignLocations with countries from Campaign salesforceData';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // Looks like MySQL 8+ default collation to utf8mb4_0900_ai_ci which means CampaignLocation
+        // has that in all environments, since it is new and we didn't specify another.
+        $this->addSql(<<<EOT
+            INSERT INTO CampaignLocation (campaign_id, countryName, regionCode)
+            SELECT c.id, jt.countryName, NULL
```
NoelLH, line 26: Tried running the `SELECT` bit of this in Staging and it looks right & performant to me; 22,200 rows in 0.014 seconds.

src/Migrations/Version20260605135515.php
```diff
@@ -0,0 +1,45 @@
...
+                    AND cl.countryName = jt.countryName COLLATE utf8mb4_0900_ai_ci
+                    AND (cl.regionCode IS NULL OR cl.regionCode = '')
+              );
+        EOT);
+    }
+
+    public function down(Schema $schema): void
+    {
+        // No safe down() migration because we cannot distinguish between CampaignLocations
+        // added by this migration and those that existed previously with NULL regionCode.
```
bdsl, line 43: not sure why there wouldn't be a safe down migration - the `CampaignLocation` table on staging is currently empty, wouldn't a safe down migration be just deleting everything from the table again?

NoelLH: I think no, because newer campaigns in Production may have rows that weren't made this way. We could possibly do it if we limit to records with country not region, but then we'd introduce a new restriction on how soon we can delete deprecated `countries` from Salesforce.

bdsl: Right. Prod campaigns don't have that yet because https://github.com/thebiggive/matchbot/pull/1998/changes/4a30f2c64c0c6e04f6d7ef6564b49700defde15f isn't in prod yet, but then don't we actually need to have that deployed to first and separately before we run this? Or at least if we don't then we'd need to run a second copy of this migration to make sure that it includes everything, or make sure that there are no relevant campaign updates in between?

shaundbs (approved):

bdsl (commented):
Question about sequencing as above:  I think we might want to get https://github.com/thebiggive/matchbot/pull/1998 into prod first before merging this, otherwise all LGTM.

src/Migrations/Version20260605135515.php
NoelLH: Yep I think you're right, saw this after the top level comment / Slack reply.

It will be out in a few minutes and I'll merge this to `develop` momentarily.

PR #2009, opened 2026-06-16
SO-78: Stop attempting to send countries from SF through to FE

bdsl:
This is causing type errors in staging as SF is no longer supplying a countries key - countries are now part of 'locations'. FE develop is already not expecting to see countries - FE main does still need countries so we need to update FE main before merging this.


NoelLH (approved):
LGTM. Strong branch name.

PR #2012, opened 2026-06-18
DON-1205: Enable sorting by location in Campaign Repository class

bdsl:
The idea is to allow a donor to see campaigns that are in the same UK region that they are in, or in the same region as any point of interest to them.

This PR adds the required sort option to the Campaign Repository. It's not yet exposed to the Frontend as that will require calling FindThatPostcode's API so will be a follow-up PR.


bdsl:
Actually I think I may have relied on a wrong assumption here - I assumed that any campaign with impact located in e.g. Haringey would also have the locations of London and England marked, so that would give it a match score of 3. But actually I think the more likely scenario is that they only select Haringey, so we need to add more logic to make that score higher for a search from Haringey than another campaign that is e.g. in in London as a whole.

But then I'm not sure if we have any way using the data we have to rank something e.g. Camden above something in e.g. Birmingham, since neither would match. Wouldn't we have to have a copy of the ONS data to be able to find that these campaigns are implicitly in London by virtue of being explicitly in Haringey or Camden?

shaundbs (commented):
Directionally I think this looks good, but I think it could be worthwhile adding tests for:

Exact local region match ranks highest
Nested region match ranks lower but still included
Campaign with no matching location ranks below matching campaigns

shaundbs (commented):
The location ranking logic looks good and the `SearchByLocationTest` is a nice addition - thanks!.

Are `regions` passed into `CampaignRepository::search()` from the public search action? In `Search.php` I can see `sortField`, `sortDirection`, `offset`, and `limit` being passed but not `regions`. If `sortField=location` can be requested without a populated regions array, do we fail safely or return a useful response?

bdsl:
> The location ranking logic looks good and the `SearchByLocationTest` is a nice addition - thanks!.
> 
> Are `regions` passed into `CampaignRepository::search()` from the public search action? In `Search.php` I can see `sortField`, `sortDirection`, `offset`, and `limit` being passed but not `regions`. If `sortField=location` can be requested without a populated regions array, do we fail safely or return a useful response?

Thanks @shaundbs .

The regions are not currently passed in from anywhere other than the test case - so I think this is good to merge but it's just a building block, its not usable by itself. I don't think they can be passed in directly as the donor / frontend won't know them, they will have to be found by querying Find That Postcode to get the list of regions that relate to the donor's location or point of interest. Then I'm not sure if they'd be passed in from the action class or if we'd decide to move the glue code between Find That Postcode and CampaignRepository into some service class.

If the FE sent a request to sort by location without specifying any coordinates or postcode I think we'd have to return an error response, but I haven't worked on the API between matchbot and frontend yet.

Do you think this needs any changes for now or does it look good to merge?

shaundbs:
If sortField=location were called without coordinates/postcode/regions, my expectation would be to either return a validation error or fall back to the default sort? I understand we don't have the API contract for location search yet but may be worth adding a TODO in any case?

bdsl:
> sortField

I'm not sure - I mean the entire support for location search in the HTTP API is todo, so I don't see that validation as more todo than anything else. If sort by location is selected and no list of locations is given then the code in this PR will already throw an exception here:

https://github.com/thebiggive/matchbot/blob/10956eccb73d4876a5dbcc4fc27e3e1154221bb8/src/Domain/CampaignRepository.php#L657

PR #2013, opened 2026-06-18
Fix `matchbot:pull-meta-campaign-from-sf` never changing campaigns to isPublished

NoelLH:


shaundbs (approved):
This looks fine for the internal update use case, but can we confirm `forInternalUpdate` is only used from trusted/internal code paths? I’d be cautious about this being exposed through any public search endpoint.

bdsl:
@shaundbs yes, I see `forInternalUpdate` is a new param, and tracing the code references the only time the function is called with `forInternalUpdate: true` is via `\MatchBot\Application\Commands\PullMetaCampaignFromSF::doExecute` , so not from any HTTP request handler.

PR #2016, opened 2026-06-22
Don 1205 find that postcode

bdsl:


src/Client/LiveFindThatPostcode.php
```diff
@@ -0,0 +1,135 @@
...
+        'N08', // Northern Ireland Ward or Electoral Division
+        'E06', // England Unitary Authority
+        'E09', // London Borough
+        'E06', // England Unitary Authority
+        'E12', // English Region
+        'E92', // England Country
+        'W92', // Wales Country
+        'S92', // Scotland Country
+        'N92', // Northern Ireland Country
+    ];
```
NoelLH, line 38: Not sure this matches how we've prompted charities / set up the possible locations, looking at data in Full.

E05 doesn't appear in the data while E07, E08, E10 do but are missing in this list. N09 is saved in Salesforce rather than N08. S12 is there rather than S13. And W06 rather than W05.

E06 is also on here twice.

src/Client/LiveFindThatPostcode.php
```diff
@@ -0,0 +1,135 @@
...
+        // remove duplicates
+        $areasToReturn = self::uniqueMultidimArray($areasToReturn, 'code'); // not sure why there are duplicate values in the data I pulled from FTP.
+
+        return $areasToReturn;
+    }
+
+
+    /**
+     * Based on comment at https://www.php.net/manual/en/function.array-unique.php by
+     * Ghanshyam Katriya
```
NoelLH:
```suggestion
     * Based on `array_unique()` {@link https://www.php.net/manual/en/function.array-unique.php#116302} comment by Ghanshyam Katriya}
```

NoelLH (commented):
Couple of things inline, general approach looks good so far.

src/Client/LiveFindThatPostcode.php
bdsl: Thanks. If we improve the list do you think this is a good approach? I will edit. I don't know if there's a better source of a list of list of prefixes in specificity order to use. (Although I suppose UK region types may not naturally have a total order of specificity, but there's probably a good enough total order we can find)

NoelLH: Yes, I think the principle should work. https://github.com/thebiggive/donate-frontend/blob/develop/src/assets/map/Sources.md doesn't list the prefixes itself, but I suppose we know those files have the levels we are interested in. And the 4 bullet points summarising the sources are in order starting with the largest area.

I think that because we only have one level below nation for each of the non-England UK nations, it would maybe also work to take the numeric part of the first 3 characters and order by that. Lowest numbers are probably always the most specific, as we are comparing overlapping zoom levels and not parts of one nation to another nation.

src/Client/LiveFindThatPostcode.php
bdsl: Not sure I quite get this - the @link tag is potentially useful, although I can click the link as-is with ctrl-click in phpstorm. But it's not based on array_unique, it's based on a function that Ghanshyam Katriya calls `unique_multidim_array`

NoelLH: Maybe add the word 'comment' back? The main point was to use a permalink to the comment.

src/Client/LiveFindThatPostcode.php
bdsl: Thanks, adjusted the list in f1710de6437, which means in the one unit test scenario we lose St James's and the locations to search for _SW1A 1AA_ are now just given as Westminster, London, England.

NoelLH: I think putting the England codes in numeric order by be easier to read and not change anything in practice, but the updated list seems like it should work.

NoelLH (commented):
I think we'll want `getDataOnPoint()` in place for pretty much the first end to end testing, as the first journey to build is probably based on browser geolocation. But also fine to merge this and iterate.

bdsl:
> I think we'll want `getDataOnPoint()` in place for pretty much the first end to end testing, as the first journey to build is probably based on browser geolocation. But also fine to merge this and iterate.

Thanks. I had thought we wanted to initially offer both the postcode and device location options, but if we're doing device location first will obviously need to build that first. I think the data returned from FTP for coordinates will be in the same format.

PR #2018, opened 2026-06-24
MAT-494 Fix guzzle string `timeout` deprecation notice

NoelLH:


src/Application/Settings.php
```diff
@@ -240,6 +240,20 @@ private function getStringEnv(array $env, string $varName, bool $throwIfMissing
         return $value;
     }
 
+    /**
+     * @param array<string, string> $env
+     */
+    private function getNumericStringEnv(array $env, string $varName): string
```
NoelLH, line 247: Made this always-required for simplicity, for now.

This did mean I had to set `SALESFORCE_CLIENT_TIMEOUT` in CI's vars so Doctrine migrations command passes. Not ideal as it's not actually relevant for that job, but setting a fallback in `migrations.php` or similar felt more hacky. And live envs already have the var set so this makes CI closer to reality.

src/Application/Settings.php
```diff
@@ -240,6 +240,20 @@ private function getStringEnv(array $env, string $varName, bool $throwIfMissing
         return $value;
     }
 
+    /**
+     * @param array<string, string> $env
```
bdsl, line 244:
```suggestion
     * @param array<string, string> $env
     * @return numeric-string
```

src/Application/Settings.php
```diff
@@ -240,6 +240,20 @@ private function getStringEnv(array $env, string $varName, bool $throwIfMissing
...
+     */
+    private function getNumericStringEnv(array $env, string $varName): string
+    {
+        $value = $this->getStringEnv(env: $env, varName: $varName, throwIfMissing: true);
+
+        if (! is_numeric($value)) {
+            throw new \Exception("Required environment variable $varName is missing.");
```
bdsl:
```suggestion
            throw new \Exception("Required environment variable $varName is missing or not numeric.");
```

bdsl (dismissed):
LGTM, I also added a couple of suggestions.

src/Application/Settings.php
NoelLH: Thanks, did the equivalent in an earlier version where it was optional then forgot to do it again

PR #2020, opened 2026-06-25
DON-1205: Allow getting UK area data on a point with GPS coordinates,…

bdsl:
… as well as postcode

### DIFF
--- a/src/Application/Actions/Campaigns/Search.php
+++ b/src/Application/Actions/Campaigns/Search.php
@@ -2,9 +2,11 @@
 
 namespace MatchBot\Application\Actions\Campaigns;
 
+use BcMath\Number;
 use Laminas\Diactoros\Response\JsonResponse;
 use MatchBot\Application\Actions\Action;
 use MatchBot\Application\Assertion;
+use MatchBot\Client\FindThatPostcode;
 use MatchBot\Domain\Campaign;
 use MatchBot\Domain\CampaignRepository;
 use MatchBot\Domain\CampaignService;
@@ -22,6 +24,7 @@ class Search extends Action
         LoggerInterface $logger,
         private CampaignRepository $campaignRepository,
         private CampaignService $campaignService,
+        private FindThatPostcode $findThatPostcode,
     ) {
         parent::__construct($logger);
     }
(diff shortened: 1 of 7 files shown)


src/Application/Actions/Campaigns/Search.php
```diff
@@ -35,6 +39,31 @@ public function __construct(
         $term = $params['term'] ?? null;
         /** @var ?string $country */
         $country = $params['country'] ?? null;
+
+        $lattitude = $params['lattitude'] ?? null;
```
NoelLH:
```suggestion
        $latitude = $params['latitude'] ?? null;
```
and throughout. Probably worth fixing early to reduce likelihood of future bugs from spelling

src/Application/Actions/Campaigns/Search.php
```diff
@@ -35,6 +39,31 @@ public function __construct(
...
+            Assertion::numeric($longitude, 'Numeric lattitude and longitude must be supplied for location-based search');
+            Assertion::string($longitude);
+            Assertion::string($lattitude);
+
+            try {
+                /** @var list<string> $regions
+                 * @mago-expect analysis:no-value,no-value,mixed-return-statement - mago seems to wrongly think the intersection of numeric and string is never.
+                 * @psalm-suppress MixedReturnStatement
+                 * */
+                $regions = $this->findThatPostcode->getDataOnPoint(new Number($lattitude), new Number($longitude)) |> (static fn(array $regions): array => \array_map(static fn(array $region): string => $region['code'], $regions));
```
NoelLH: This seems like quite a lot to take in on one line. Maybe a comment briefly saying the goal? Not a blocker.

NoelLH (commented):
Couple of things inline, otherwise LGTM

src/Application/Actions/Campaigns/Search.php
bdsl: Thanks, yes definitely worth fixing now.

src/Application/Actions/Campaigns/Search.php
bdsl: Thanks, will split out and/or comment. Really made more complicated than it strictly needs to be because my FTP client returns codes and names - and the names are very useful for debugging and testing - but the search function just wants a list of codes.

src/Application/Actions/Campaigns/Search.php
bdsl: Fixed in [276f938](https://github.com/thebiggive/matchbot/pull/2020/commits/276f938c8449febfe0358fc06d5c5dd22123ad84)

src/Application/Handlers/HttpErrorHandler.php
```diff
@@ -82,7 +82,12 @@ protected function respond(): Response
                 /** @psalm-suppress PossiblyUndefinedArrayOffset
                  * Offsets defined at https://www.php.net/manual/en/exception.gettrace.php
                  */
-                fn(array $frame) => "{$frame['class']}::{$frame['function']} {$frame['file']}:{$frame['line']}",
+                function (array $frame) {
+                    $class = $frame['class'] ?? '';
+                    $function = $frame['function'] ?? '';
+
+                    return "{$class}::{$function} {$frame['file']}:{$frame['line']}";
+                },
```
bdsl, line 90: this causes an extra error in the error display if the error doesn't have a class, e.g. is from a built in function.

NoelLH (approved):

PR #2021, opened 2026-06-26
DON-1196 Remove old campaign SF compatibility checker

NoelLH:


src/Application/Commands/CreateFictionalData.php
```diff
@@ -226,17 +224,6 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
 
             $this->em->flush();
-
-            $renderedCampaign = $this->campaignService->renderCampaign($campaign, metaCampaign: null);
```
bdsl, line 230: not sure if it may be still worth calling `renderCampaign` here even if we void the result to just to check it doesn't throw? This runs in CI as well as on dev machines so could be a wortwhile and simple extra check.

bdsl (approved):
LGTM - the checker is checking compatibility with a historical thing, they only really needed to be compatible during the replacement process. Added one suggestion, not a blocker.

src/Application/Commands/CreateFictionalData.php
NoelLH: Not totally sure but I think that as `CampaignServiceTest` covers the same method directly, we probably don't cover any very likely types of error by also doing it here.

PR #2022, opened 2026-06-26
Fix common methods of Campaigns being updated from Salesforce failing to update `salesforceLastPull`

NoelLH:


src/Domain/Campaign.php
```diff
@@ -554,6 +553,8 @@ final public function updateFromSfPull(
         $this->replaceLocations($locations);
 
         unset($sfData['charity']); // charity stores its own data, we don't need to keep a copy here.
+
+        $this->setSalesforceLastPull(new \DateTimeImmutable('now'));
```
NoelLH, line 557: Would be a fair bit of faff to pass a Clock through all callers, not really worth it here IMO

src/Domain/CampaignRepository.php
```diff
@@ -830,9 +830,8 @@ public function updateFromSf(Campaign $campaign, bool $withCache = true): void
         $campaignData = $this->getClient()->getById($campaign->getSalesforceId(), $withCache);
 
         $this->updateCampaignFromSFData($campaign, $campaignData);
-
-        $campaign->setSalesforceLastPull(new DateTime('now'));
```
NoelLH, line 834: `updateCampaignFromSFData()` calls `Campaign::updateFromSfPull()` which now handles this. More consistent across callers and more similar to what we do for e.g. charities.

shaundbs (approved):

bdsl (commented):
LGTM

PR #2026, opened 2026-07-01
MAT-488: Add field to donation to track how long its match reservatio…

bdsl:
…n should be initially held for

### DIFF
--- a/src/Domain/Donation.php
+++ b/src/Domain/Donation.php
@@ -139,6 +139,18 @@ class Donation extends SalesforceWriteProxy
     private Money $expectedMatchAmount;
 
     /**
+     * Time until which match any reserved match funds should be kept, *even*
+     * if the donation client goes offline. If they are online they can be expected to ask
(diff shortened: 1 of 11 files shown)


src/Domain/Donation.php
```diff
@@ -138,6 +138,18 @@ class Donation extends SalesforceWriteProxy
     #[ORM\Embedded()]
     private Money $expectedMatchAmount;
 
+    /**
+     * Time until which match any reserved match funds should be kept, *even*
+     * if the donation client goes offline. If they are online they can be expected to ask
+     * to extend this periodically, and matchbot will comply up to a limit of about half an hour.
```
NoelLH:
```suggestion
     * if the donation client goes offline. If they are online and campaign has time left they can be expected to ask
     * to extend this periodically, and matchbot will comply up to a limit of about 25 minutes.
```

NoelLH (dismissed):
Doc block minor suggestion inline, LGTM

PR #2029, opened 2026-07-03
MAT-488:: Add action to extend reseration on donation

bdsl:


shaundbs:
More generally I think it could be worth adding a few tests to improve coverage e.g.:

- testDoesNotExtendReservationPastThirtyTwoMinutesFromCreation()
- testRepeatedExtendsDoNotMoveReservationPastLimit()

shaundbs (dismissed):

bdsl:
> More generally I think it could be worth adding a few tests to improve coverage e.g.:
> 
>     * testDoesNotExtendReservationPastThirtyTwoMinutesFromCreation()
> 
>     * testRepeatedExtendsDoNotMoveReservationPastLimit()

Doesn't `testExtendsReservationOnlyUpToALimit` already show that it does not extend past 32 minute from creation? It's a scenario of a donation created at noon, extension request sent at 12:37, and the extension only goes to 12:32. Actually I noticed there's a misnamed variable in that test so have fixed.

Not sure about the repeated extends thing - as the implementation of `extendReservationFrom` doesn't read the `fundsReservedUntil` value at all I don't think that would actually cover any additional logic. But feel free to add any tests you think are worth adding or we can pair to add some tests.

PR #2034, opened 2026-07-07
Mark donations for re-push to Salesforce on `ConnectException`

NoelLH:
Same as we already do for `BadResponseException`.


bdsl (approved):
LGTM

PR #2037, opened 2026-07-08
Fix expectation that `additionalImageUris` is always set

NoelLH:
It has been deprecated and effectively unused for some time but was left required on the model

Also drop unused `logoUri` on Campaign


bdsl (approved):
LGTM

PR #2039, opened 2026-07-08
MAT-493 Support requesting command runs through CI, with consumers reading from the queue and running them

NoelLH:
Mostly from collab with Copilot + Gemini 3.1 Pro.

Not quite ready to run live. `circleci` AWS user (or role) will need permission before we can actually publish to SQS directly like this, and our internal docs are still to update. But probably good to get a next review at this stage.

As it stands, observability would be through checking CloudWatch Logs somewhat proactively, e.g.
1. tail logs for the running relevant MB consumer service's tasks, or
2. search in CloudWatch Logs Insights for the specific command you invoked, to more quickly find the particular task and timestamp

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,18 @@ orbs:
   jira: circleci/jira@2.2.0
   slack: circleci/slack@5.1.1
 
+parameters:
+  run_command:
+    type: boolean
+    default: false
+  command_environment:
+    type: enum
+    enum: ["regression", "staging", "production"]
+    default: "regression"
+  command_string:
+    type: string
+    default: "matchbot:tick"
+
 jobs:
   test: # Also lints first
     # Working directory must match docker-compose mount / Docker `COPY` for real Doctrine proxy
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -156,7 +168,52 @@ jobs:
...
+            # High-priority queue has no suffix after `matchbot` apart from the `.fifo` which
+            # all first-in-first-out queues have.
+            # Name is defined in infrastructure repo but a real example is: tbg-staging-eu-west-1-matchbot.fifo
+            QUEUE_URL="https://sqs.${AWS_REGION}.amazonaws.com/${AWS_ACCOUNT_ID}/tbg-${ENV}-${AWS_REGION}-matchbot.fifo"
+
+            DEDUP_ID=$(date +%s)-$RANDOM
+
+            # Symfony Messenger PhpSerializer wraps the serialized PHP object in a JSON payload for SQS.
+            BODY="O:46:\"MatchBot\Application\Messenger\CommandRequest\":1:{s:7:\"command\";s:${COMMAND_LEN}:\"${COMMAND}\";}"
+            MESSAGE="{\"body\": \"$BODY\", \"headers\": {\"type\": \"MatchBot\\\\Application\\\\Messenger\\\\CommandRequest\"}}"
```
NoelLH, line 199: Taking this bit on trust from an agent session for now, but I think the simplest next step is to get the user permissions in place and try it from CircleCI.

bdsl:
> As it stands, observability would be through checking CloudWatch Logs somewhat proactively, e.g.

IMHO since we don't expect to be running these too frequently and when we do we'll probably always want to know whether it worked, what it did, it might be worth making the output go to the #matchbot channel in Slack. Although depending what command we run there might be a concern that too many people have access to that in case it outputs donor PII.

.circleci/config.yml
```diff
@@ -8,6 +8,18 @@ orbs:
...
+  run_command:
+    type: boolean
+    default: false
+  command_environment:
+    type: enum
+    enum: ["regression", "staging", "production"]
+    default: "regression"
+  command_string:
+    type: string
+    default: "matchbot:tick"
```
bdsl, line 21: any reason for the default? This doesn't seem that useful to run manually since it runs every minute either. I think ideally we'd have no default and it would require an argument, or if that's not possible default to something that's maybe a purpose made command to just smoke test the command running system.

Also wondering if it would be worth taking a full command line for bash, e.g. `./matchbot matchbot:tick` and then we have flexibility to run composer commands, matchbot commands, or even standard linux CLI utilities.

bdsl: Taking a full command line would probably mean using the Symfony Process component instead of the Console component in the `CommandRequestHandler` class defined later in this PR.

src/Application/Messenger/Handler/CommandRequestHandler.php
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public function __invoke(CommandRequest $message): void
+    {
+        $this->consoleApplication->setAutoExit(false);
+        $this->consoleApplication->setCatchExceptions(false);
+
+        $exitCode = $this->consoleApplication->run(
+            new StringInput($message->command),
+            new ConsoleOutput()
```
bdsl, line 51: may be worth using a [`BufferedOutput`](https://github.com/symfony/symfony/blob/8.1/src/Symfony/Component/Console/Output/BufferedOutput.php) instead of `ConsoleOutput` so we can explicitly log the output as output of a command request run.

bdsl (commented):
Added comments inline, otherwise all LGTM. Still feel like we might want a more private way to run commands, that would allow things like entering a donor's email address to reset their password, but we could certainly start with this and then if neccassary find a more private way to generate command requests.

src/Application/Messenger/Handler/CommandRequestHandler.php
NoelLH: As discussed, agreed on making this clearer but don't think the Output class affects it.

.circleci/config.yml
NoelLH: Discussed in pairing. `matchbot:` feels like a useful hint as to the command structure. Arbitrary process running is less testable + not sure it feels worth it.

PR #2061, opened 2026-07-23
Stop requiring charity properties that are sometimes missing and not required by FE

shaundbs:
…boardingStatus (triggering Prod alarms)

### DIFF
--- a/src/Domain/CampaignService.php
+++ b/src/Domain/CampaignService.php
@@ -194,7 +194,7 @@ class CampaignService
         $charityHttpModel = new \MatchBot\Application\HttpModels\Charity(
             id: $charity->getSalesforceId(),
             name: $charity->getName(),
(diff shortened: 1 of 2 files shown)


NoelLH (dismissed):

src/Domain/CampaignService.php
```diff
@@ -194,9 +194,8 @@ public function renderCampaign(CampaignDomainModel $campaign, ?MetaCampaign $met
         $charityHttpModel = new \MatchBot\Application\HttpModels\Charity(
             id: $charity->getSalesforceId(),
             name: $charity->getName(),
-            optInStatement: $sfCharityData['optInStatement'],
+            optInStatement: $sfCharityData['optInStatement'] ?? null, // typed as `string` in ts, but also tested for truthiness.
```
bdsl:
```suggestion
            optInStatement: $sfCharityData['optInStatement'] ?? '', // typed as `string` in ts, but also tested for truthiness.
```

PR #2063, opened 2026-07-24
MAT-488: Detect if match funds expired when client requests to extend

bdsl:


src/Application/Actions/Donations/ExtendFundsReservationTime.php
```diff
@@ -87,7 +86,15 @@ protected function action(Request $request, Response $response, array $args): Re
...
-            $donation->extendReservationFrom($this->clock->now());
+            try {
+                $donation->extendReservationFrom($this->clock->now());
+            } catch (ExpectedMatchFundsNotFound $e) {
+                return $this->validationError(
+                    $response,
+                    logMessage: $e->getMessage(),
+                    publicMessage: "This donation does not have the expected match funds, so reservation cannot be extended",
+                );
+            }
```
bdsl, line 97: I suppose FE will check for the HTTP 400 code and assume this was the issue if so. If it's any other HTTP error we clearly don't want to assume the match funds are gone, since it could just be a connection issue which we're trying to be resilient too.

Or it might be worth starting system of machine readable domain-specific error codes between matchbot and frontend, so FE could check for a code that means its this specific exception. LMK if you think it'd be a good idea for me to implement that now?

NoelLH (approved):

PR #2065, opened 2026-07-27
BG2-3342: Do not allow a standalone charity campaign to be open witho…

bdsl:
…ut funding

### DIFF
--- a/integrationTests/IntegrationTest.php
+++ b/integrationTests/IntegrationTest.php
@@ -299,7 +299,8 @@ abstract class IntegrationTest extends TestCase
         bool $campaignOpen = true,
         ?string $charitySfId = null,
         string $charityName = 'Some Charity',
-        bool $isRegularGiving = false
+        bool $isRegularGiving = false,
+        bool $isStandalone = false
     ): array {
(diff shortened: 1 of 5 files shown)


NoelLH (commented):
Two questions about the broad design of the solution:
1. Are we sure being standalone is enough to exclude campaigns without matching? Unmatched campaigns are still allowed, just not promoted in Explore. Could `Campaign::$isMatched` help here?
2. It feels slightly confusing to have new logic here for standalone, expecting-matching campaigns but `isPublished` and the related application fields determining the same thing for application campaigns, i.e. "is it (probably) proceeding?" I briefly thought we may have missed the equivalent check for application campaigns, making them harder to find on e.g. BG2-3308 but not blocking donations; but I now _think_ `isPublished` likely acts as a proxy for those fields. Could we standardise by also un-publishing "Matched Funded" type CCampaigns in Salesforce that get near their start date without funding, setting `isPublished` false? If not easily, can we cross reference the two places managing this logic in MatchBot to make it easier to remember that they both exist?

NoelLH (approved):

PR #2072, opened 2026-08-04
BG2-3342: Exclude non-application charity campaigns missing funds from lists

bdsl:
I think this covers everywhere that matchbot would present the campaign in a list for discovery: Search, explore, sitemap, and the charity page.

The charity themselves will still have access to the page via a link from their portal, and may or may not choose to share it.

### DIFF
--- a/src/Domain/CampaignRepository.php
+++ b/src/Domain/CampaignRepository.php
@@ -26,7 +26,7 @@ use function trim;
 class CampaignRepository extends SalesforceReadProxyRepository
 {
     private ClockInterface $clock;  // @phpstan-ignore property.uninitialized
-    private string $appStatusWhereClause = <<<DQL
+    private string $statusAndFundingWhereClause = <<<DQL
         (
             (campaign.metaCampaignSlug IS NULL AND (campaign.isMatched = 0 OR campaignStatistics.matchFundsTotal.amountInPence > 0))
             OR
@@ -382,7 +382,7 @@ class CampaignRepository extends SalesforceReadProxyRepository
              campaign.charity = :charity
              AND campaign.isPublished = true
              AND (campaignStatistics.donationSum.amountInPence > 0 OR campaign.endDate > :at OR campaign.endDate IS NULL)
-             AND {$this->appStatusWhereClause}
+             AND {$this->statusAndFundingWhereClause}
              ORDER BY approxStatusRank ASC, campaign.endDate ASC
             DQL
         );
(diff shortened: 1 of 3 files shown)


src/Domain/CampaignRepository.php
```diff
@@ -370,16 +371,16 @@ public function findCampaignsForCharityPage(Charity $charity, \DateTimeImmutable
             <<<DQL
             SELECT campaign,
             CASE
-                WHEN statistics.approxStatus = 'Active' THEN 0
```
bdsl, line 373: renamed to align with another query so they can share the `appStatusWhereClause` edited above.

src/Domain/CampaignRepository.php
```diff
@@ -28,8 +28,10 @@ class CampaignRepository extends SalesforceReadProxyRepository
     private ClockInterface $clock;  // @phpstan-ignore property.uninitialized
     private string $appStatusWhereClause = <<<DQL
         (
-            campaign.metaCampaignSlug IS NULL OR
+            (campaign.metaCampaignSlug IS NULL AND (campaign.isMatched = 0 OR campaignStatistics.matchFundsTotal.amountInPence > 0))
+            OR
             (
+                campaign.metaCampaignSlug IS NOT NULL AND
```
NoelLH: I think this var name becomes a little misleading now there's non-trivial conditions for non-application campaigns. Maybe change to `$statusAndFundingWhereClause`?

src/Domain/CampaignStatistics.php
```diff
@@ -55,20 +55,25 @@ class CampaignStatistics
...
     #[ORM\Embedded(columnPrefix: 'match_funds_total_')]
     private Money $matchFundsTotal;
 
-    /** Set on construct and updated when donations change. */
+    /** Total of match funds that have been withdrawn for donations to this campaign. */
     #[ORM\Embedded(columnPrefix: 'match_funds_used_')]
     private Money $matchFundsUsed;
 
-    /** Set on construct and updated when donations change. */
+    /** Total of match funds still available ot use for donations to this campaign. Should be equal to
```
NoelLH:
```suggestion
    /** Total of match funds still available to use for donations to this campaign. Should be equal to
```

NoelLH (commented):
Apart from the var naming thing noted inline, LGTM

src/Domain/CampaignRepository.php
bdsl: roger wilco

bdsl: Done in b693b5159 @NoelLH

src/Domain/CampaignStatistics.php
bdsl: Thanks!

NoelLH (approved):

PR #2073, opened 2026-08-04
Upgrade to Symfony 8 packages et al

NoelLH:


integrationTests/DonationPersistenceTest.php
```diff
@@ -42,6 +42,7 @@ private function assertRowsSimilar(array $expected, array $actual, string $messa
     {
         $ignoredColumns = ['id' => 0, 'uuid' => 0, 'updatedAt' => 0, 'createdAt' => 0, 'campaign_id' => 0];
 
+        // @phpstan-ignore function.alreadyNarrowedType
```
NoelLH: Doesn't work either, leaves it with

```
  46     No error with identifier function.alreadyNarrowedType is reported on line 46.                                                                                       
         🪪  ignore.unmatchedIdentifier (non-ignorable)         
```

bdsl (commented):
LGTM other than the test failures / errors

integrationTests/CallFrequentCommandsTest.php
```diff
@@ -29,6 +29,20 @@ public function testTick(): void
...
+
+        $this->getContainer()->set(
+            SendStatistics::class,
+            new SendStatistics(
+                new NativeClock(),
+                $mockCloudWatchClient,
+                $this->getService(DonationRepository::class),
+                $this->getService(Environment::class),
+            ),
+        );
```
bdsl: actually this may be unnecessary let me try without it.

NoelLH (commented):
New commit LGTM

PR #2081, opened 2026-08-07
MAT-488: Always use short reservation times for new donations

bdsl:
This was optional previously just to account for donations from old versions of donate frontend, but as those versions of donate-frontend have been out of prod for hours those donations will have had their matching expire by now anyway.

### DIFF
--- a/src/Domain/DoctrineDonationRepository.php
+++ b/src/Domain/DoctrineDonationRepository.php
@@ -31,24 +31,13 @@ class DoctrineDonationRepository extends SalesforceProxyRepository implements Do
     #[\Override]
     public function findWithExpiredMatching(\DateTimeImmutable $now): array
     {
-        // and we only need to expire donations that were create AFTER this point, because if they were created at
-        // before it we would have already expired them in a previous run.
-        $expireAfter = $now->sub(new \DateInterval('PT1H'));
-
-        if ((int) $now->format('i') % 30 === 0) {
-            // once every 30 minutes, try expiring any older donations just in case they were missed, but it shouldn't
-            // really be necassary unless this stopped running for some reason.
-            $expireAfter = new \DateTimeImmutable('2025-08-05T00:00:00z');
-        }
-
         $query = $this->getEntityManager()->createQuery(<<<'DQL'
             SELECT d.uuid FROM MatchBot\Domain\Donation d
 
             -- Only select donations with 1+ FWs. We don't need any further info about the FWs.
             INNER JOIN d.fundingWithdrawals fw
             WHERE d.donationStatus IN (:expireWithStatuses)
-            AND (d.fundsReservedUntil is null OR d.fundsReservedUntil < :now)
-            AND d.createdAt > :expireAfter
+            AND d.fundsReservedUntil < :now
 
             -- First of a regular giving series is Pending during 3DS. If we ever make the timeout for
             -- that longer than the timeout for matching, we still want to ensure matching can't be
@@ -59,8 +48,7 @@ class DoctrineDonationRepository extends SalesforceProxyRepository implements Do
             DQL
         )
             ->setParameter('expireWithStatuses', [DonationStatus::Pending->value, DonationStatus::Cancelled->value])
-            ->setParameter('now', $now)
-            ->setParameter('expireAfter', $expireAfter);
+            ->setParameter('now', $now);
 
         // As this is used by the only regular task working with donations,
         // `ExpireMatchFunds`, it makes more sense to opt it out of result caching
--- a/src/Domain/DoctrineDonationRepository.php
+++ b/src/Domain/DoctrineDonationRepository.php
@@ -31,9 +31,6 @@ class DoctrineDonationRepository extends SalesforceProxyRepository implements Do
     #[\Override]
     public function findWithExpiredMatching(\DateTimeImmutable $now): array
     {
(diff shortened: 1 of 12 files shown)


src/Domain/DoctrineDonationRepository.php
```diff
@@ -50,7 +50,7 @@ public function findWithExpiredMatching(\DateTimeImmutable $now): array
             -- Only select donations with 1+ FWs. We don't need any further info about the FWs.
             INNER JOIN d.fundingWithdrawals fw
             WHERE d.donationStatus IN (:expireWithStatuses)
-            AND (d.createdAt < :expireBefore OR (d.fundsReservedUntil is not null AND d.fundsReservedUntil < :now))
+            AND (d.createdAt < :expireBefore OR d.fundsReservedUntil is null OR d.fundsReservedUntil < :now)
```
bdsl: actually we can probably simplify more by removing the d.createdAt part.

bdsl: done in ec81ac7

src/Domain/DoctrineDonationRepository.php
```diff
@@ -50,7 +47,7 @@ public function findWithExpiredMatching(\DateTimeImmutable $now): array
             -- Only select donations with 1+ FWs. We don't need any further info about the FWs.
             INNER JOIN d.fundingWithdrawals fw
             WHERE d.donationStatus IN (:expireWithStatuses)
-            AND (d.createdAt < :expireBefore OR (d.fundsReservedUntil is not null AND d.fundsReservedUntil < :now))
+            AND (d.fundsReservedUntil is null OR d.fundsReservedUntil < :now)
```
NoelLH: Nice to see simplification in the same query we saw running a bit slow. (I suspect the `is not null` was always superfluous in the "before" query too.)

I'm wondering if we can't go further though. Isn't it unnecessary to check that `d.fundsReservedUntil is null` now that the code has been in Production for a while and handled expiry on all donations that predate the new reservation fields? That should hopefully make it more viable to use indexes effectively on this part of the query. (If I'm right it would also be good to document the conditions for it being `null` – i.e. old data – on `Donation::$fundsReservedUntil`.

Finally, I think we should probably add an index on `fundsReservedUntil`. I ran this bit past Gemini which said:
> Adding an index on `fundsReservedUntil` will allow the database to use an Index Merge optimization (combining the results of two indexes for an OR condition) or speed up filtering.

src/Domain/DoctrineDonationRepository.php
```diff
@@ -50,7 +47,7 @@ public function findWithExpiredMatching(\DateTimeImmutable $now): array
...
             AND d.createdAt > :expireAfter
```
NoelLH: I'm also wondering if we might not be better off removing this constraint to keep it simple, now we have the new dynamic expiry. Perhaps that should be in a separate 2nd PR so we can more easily assess the performance impact. I would hope that with good indexing on the remaining parts it would not slow the query down and might speed it up.

Unless I'm missing something, limiting how far we look back isn't a hard logical requirement – hence varying it occasionally to pick up any missed older donations.

NoelLH (commented):
A couple of query performance things inline which relate to BG2-3357.

Otherwise LGTM.

src/Domain/DoctrineDonationRepository.php
bdsl: yes I think it was superfluous before. And yes we don't really need to release funds for donations with fundsReservedUntil null because the last of those was created on Wednesday 5th - so we can assume any funds they had temporarily reserved have already been released.

I guess it just seemed in principle that the funds should only be reserved IF fundsReservedUntil is set. But we could simplify by using > instead of < so that null automatically gets treated like 1970, or we could make the field non-nullable and have it default to 1970 or possibly to the donation creation time. I think I'd do 1970 as a bit simpler.

Yes adding the index definitely makes sense. What do you think about making it non-nullable?

NoelLH: I'm not sure I follow re. the comparison direction. Nulls won't appear in a `<` comparison either, they're not treated as equivalent to 1970 in that way.

I think that means it is quite simple as-is and there isn't really a strong reason to make it non-nullable or adjust the old Donations.

src/Domain/DoctrineDonationRepository.php
bdsl: Right, the expireAfter was just a performance thing to stop it repeating the same work every minute, but that's probably not necessary since we are already limiting by joining to the FundingWithdrawals and only selecting donations with unreleased withdrawals.

src/Domain/DoctrineDonationRepository.php
bdsl: I might be wrong, I was thinking that

```
AND (d.fundsReservedUntil < :now)
```

that would treat null like the year 3000 - it's not less than now.

But if you have 

```
AND NOT (d.fundsReservedUntil > :now)
```

that might make null be like the year 1970 - not more than now. But maybe it doesn't work that way.

NoelLH: I was thinking we'd just be left with `AND d.fundsReservedUntil < :now` rather than negating the whole expression. That would always be false if the field is null based on my memory and cursory testing, which seems fine?

NoelLH (dismissed):

PR #2088, opened 2026-08-21
BG2-3354: WIP - return counts of campaigns per location with search results for UK

bdsl:
This should allow adding further properties to the object for statistics about the result as a whole - like the count of how many matching campaigns there are in in total and in each geographical region.

### DIFF
--- a/src/Domain/CampaignRepository.php
+++ b/src/Domain/CampaignRepository.php
@@ -789,6 +789,46 @@ class CampaignRepository extends SalesforceReadProxyRepository
             forInternalUpdate: $forInternalUpdate,
         );
 
+        if ($country === 'United Kingdom' && ! Environment::current()->isProduction()) {
+            // also fetch a count of how many relevent campaigns have
+            // locations each major part of the UK.
+//
+//            $locationCountQueryBuilder = $this->getEntityManager()->createQuery(
+//                <<<DQL
+//                    SELECT COUNT(cl), cl.regionCode FROM MatchBot\Domain\CampaignLocation cl
+//                    WHERE cl.campaign IN (
+//                        SELECT -- here we would need to repliacte the search logic of filterForSearch but keeping all that twice seems very risky.
+//                        -- I don't think we can just call it as-is even if we start using a querybuilder here since we're trying to get a list of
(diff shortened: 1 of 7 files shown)


shaundbs (changes requested):
Could we add a test around locationCounts itself? Atm the existing tests look to have mostly been updated for the new CampaignSearchResult return type but I can’t see anything asserting the new behaviour. Would be good to cover at least a couple of campaigns in different regions and confirm the count are returned as expected.

bdsl:
@shaundbs yes not a bad idea then we can see any changes in behaviour later with changes in the test examples. Will do.

src/Domain/CampaignRepository.php
```diff
@@ -790,6 +788,63 @@ public function search(
...
+                forInternalUpdate: $forInternalUpdate,
+            );
+
+            /** @var list<array{numCampaigns: int, regionCode: string}> $locationCounts */
+            $locationCounts = $lq2->getQuery()->getResult();
+
+            foreach ($summaryRegionCodes as $code) {
+                if (! \array_any($locationCounts, fn(array $item): bool => $item['regionCode'] === $code)) {
+                    // it's not there, we can show a zero on the map (or FE could choose to hide zeroes)
+                    $locationCounts[] = ['regionCode' => $code, 'numCampaigns' => 0];
```
NoelLH: If we do want the zeroes, I think it would be fine and perhaps easier to follow to left join CampaignLocations instead. I'm not sure we need them at all though.

bdsl: I'm not sure I follow - campaignLocation is in the `from` call at the start of the query build, not a later join. Would we have to start by selecting from the hard-coded list of 12 locations?

But yes may be fine to just delete this part and not include the regions with zero in the result, no reason for anyone to zoom in to those. Think I will leave it for now and can remove in another PR if it's not needed.

bdsl:
@shaundbs I've added an assertion on the returned count of campaigns per UK location as suggested in fb4fbe297d0

shaundbs (approved):
Yep, this covers it nicely - thanks! The test exercises the new locationCounts behaviour LGTM.

